PR #20, opened 2022-03-01
WIP: pgcopydb copy blobs

### LINKED ISSUE #14: Support for large objects
Hi @dimitri ,

Just tested latest master and i see that large objects are not transferred. Could you include the *-b* pg_dump option?

dimitri:
New command to support copying large objects from the source to the target
database instances, in a streaming fashion: we avoid spilling the large
object data to disk altogether. Processing of large objects is automatically done
from an auxiliary process, ran as part as the table data processing.

Fixes #14.


df7cb:
I don't know if you want (or should) go down that route, but a frequent problem when upgrading/migrating databases is that LOs are created by individual create/populate/grant commands, which scales horribly if there's 100k+ of them. That includes pg_upgrade which transfers the LO data as bulk, but still runs the individual create/grant statements.

In experiments, I was able to speed that up by moving pg_largeobject_metadata to a regular table, pg_upgrade it, and move the data back to pg_largeobject_metadata. pgcopydb could try something similar. (IMHO that should also be fixed in PG by being smarter about how to dump LOs, but maybe pgcopydb wants to support older versions as well.)

dimitri:
Hi @df7cb ; thanks for the input! I think I will do it the naive way at first. Thanks to your message I realise I didn't look at the ACL side of things, it might be that pg_dump/pg_restore does the right thing already, I will have to check. The way to improve the situation that I though of was to introduce support for parallelism in the blob copying, share the work among several workers. 

I suppose your trick with bulk loading pg_largeobject_metadata requires superuser access, and then even maybe allowing catalog updates. That wouldn't be possible in a Postgres Cloud managed instance, unfortunately, and that's an important use case for pgcopydb...

df7cb:
Yes, I'm aware that this is icky. :)

Speeding up bulk copying of large LOs makes a lot of sense, and the case of too many too small LOs should really be fixed by the customer switching to bytea instead.

dimitri:
It turns out that `pg_dump --section=pre-data` takes care of the large object metadata, and then pgcopydb calls into pg_restore on this output, so that's how the large objet metadata are migrated. Then pgcopydb handles the copy of the large object data parts itself, and at the moment with a single auxiliary process that runs concurrently to the rest of the migration, and avoids spilling to disk.

That should be good enough for us to get started, and we could distribute the large object data parts work among a set of workers if necessary later, in a follow-up PR. Then we might want to implement some tricks to avoid the metadata scaling issue you're mentioning, but I feel like that's also later optimisation material. My plan is to now document what I've done in this PR and then merge it as it is, in order to get started with LO support and hopefully get more feedback from the field.

df7cb:
Full ack, I just wanted to make you aware. It doesn't have to be fixed now, and possibly not even at all in pgcopydb. It's a pg_dump problem.

PR #23, opened 2022-03-04
Fix exclude constraint index creation

### LINKED ISSUE #22: Error on ALTER TABLE ... USING INDEX with EXCLUDE CONSTRAINT
There's an issue with method of concurrently index creation that fails to create proper `EXCLUDE CONSTRAINS` with `ALTER TABLE ... ADD CONSTRAIND ... USING INDEX`. As stated in the PostgreSQL docs:

```
ADD table_constraint_using_index
This form adds a new **PRIMARY KEY or UNIQUE** constraint to a table based on an existing unique index. 
All the columns of the index will be included in the constraint.

The index cannot have expression columns nor be a partial index. 
Also, **it must be a b-tree index** with default sort ordering. 
These restrictions ensure that the index is equivalent to one that would be built 
by a regular ADD PRIMARY KEY or ADD UNIQUE command.

CREATE UNIQUE INDEX CONCURRENTLY dist_id_temp_idx ON distributors (dist_id);
ALTER TABLE distributors DROP CONSTRAINT distributors_pkey,
    ADD CONSTRAINT distributors_pkey PRIMARY KEY USING INDEX dist_id_temp_idx;
```

`EXCLUDE CONSTRAINT` indexes are not created as `UNIQUE` and are `GIST` indexes instead of `b-tree` (not supported by this syntax) resulting `ALTER TABLE` query to end with syntax error (missing `UNIQUE` or `PRIMARY KEY` word):

```
13:49:28 48483 INFO  ALTER TABLE "public"."recreate" ADD CONSTRAINT "constraint_fail_idx"  USING INDEX "constraint_fail_idx";
13:49:28 48483 ERROR [TARGET] ERROR:  syntax error at or near "USING"
13:49:28 48483 ERROR [TARGET] LINE 1: ...."recreate" ADD CONSTRAINT "constraint_fail_idx"  USING INDE...
13:49:28 48483 ERROR [TARGET]                                                              ^
13:49:28 48483 ERROR SQL query: ALTER TABLE "public"."recreate" ADD CONSTRAINT "constraint_fail_idx"  USING INDEX "constraint_fail_idx";
```

Here is the recreate script:

```
psql

CREATE DATABASE test;

CREATE DATABASE test_restore;

\c test

CREATE TABLE recreate (a integer, b integer);

insert into recreate values (1,2), (3,4);

ALTER TABLE recreate ADD CONSTRAINT constraint_fail_idx   EXCLUDE USING btree (a WITH=);

\q

export PGCOPYDB_SOURCE_PGURI="postgres://:5432/test"
export PGCOPYDB_TARGET_PGURI="postgres://:5432/test_restore"

pgcopydb copy-db --verbose
```

realyota:
It fixes #22 although it is the simplest way to just skip those indexes.

### DIFF
--- a/src/bin/pgcopydb/defaults.h
+++ b/src/bin/pgcopydb/defaults.h
@@ -7,7 +7,7 @@
 #define DEFAULTS_H
 
 /* additional version information for printing version on CLI */
-#define PGCOPYDB_VERSION "0.3"
+#define PGCOPYDB_VERSION "0.4"
 
 /* environment variable to use to make DEBUG facilities available */
 #define PGCOPYDB_DEBUG "PGCOPYDB_DEBUG"
--- a/src/bin/pgcopydb/schema.c
+++ b/src/bin/pgcopydb/schema.c
@@ -335,6 +335,7 @@ schema_list_table_indexes(PGSQL *pgsql,
 		"                and d.deptype = 'i'"
 		"          left join pg_constraint c ON c.oid = d.refobjid"
 		"    where r.relkind = 'r' and r.relpersistence = 'p' "
+		"          and not x.indisexclusion "
 		"      and n.nspname !~ '^pg_' and n.nspname <> 'information_schema'"
 		"      and rn.nspname = $1 and r.relname = $2"
 		" order by n.nspname, r.relname";


dimitri:
Hi @realyota ; thanks for having a look into fixing that problem! I will close this PR though, because of the two following issues (one minor, one more important):

  1. we don't edit the version number during branch development, only at release time
  2. your approach to filtering out the exclusion indexes makes it so that those are going to be built by pg_restore in the last step, and it means we miss all the concurrency approach that pgcopydb implements.

I think we instead want to recognise that the index is special (`indisexclusion`) and then be smart(er) about that fact. It could mean creating the index by means of the ALTER TABLE command entirely, of it that's possible, find a way to make our two-stages approach work even for exclusion constraints, depending on what Postgres is capable of.

Are you interested in diving into that area? Let's move the discussion to issue #23 now that this PR is closed, and if you want to, I'll be happy to review your next attempt!

PR #40, opened 2022-05-12
filters: Use pg_roles instead of pg_authid.

shubhamdhama:
The user in connection must have a superuser role to access the pg_authid
table, which may not be possible for managed services. The fix here is to
use publicly readable view pg_roles
https://www.postgresql.org/docs/current/view-pg-roles.html.


dimitri (approved):
Thanks!

PR #137, opened 2022-11-01
Debian: B-D on libzstd-dev. (Closes: #1022290)

df7cb:


df7cb:
@dimitri, can we make a new release with this to fix https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1022290 ?

dimitri:
Sure, will see about that tomorrow. I will need to find out again how to close a bug from the changelog, but that should be easy enough... I believe the documentation is pretty good, it's all about finding it ;-)

dimitri:
See #144

PR #152, opened 2022-11-21
Fix migration failure of an empty database with --drop-if-exists.

shubhamdhama:
It fixes a corner case when --drop-if-exists is used to migrate a database that contains no tables. Before this change, `copydb_target_drop_tables` generates an invalid SQL query `DROP TABLE IF EXISTSCASCADE`.


dimitri (approved):

dimitri:
Thanks @shubhamdhama !

PR #164, opened 2022-12-09
Fix `lsn` for KEEPALIVE action.

shubhamdhama:
It fixes an issue under the following conditions,
1. Source doesn't have any activity.
2. `endpos` is set to `current` LSN.
3. The action that crosses the `endpos` is `KEEPALIVE`.

in this case, the written `lsn` that corresponds to `KEEPALIVE` isn't correct, i.e. less than `endpos`. This ultimately leads to Apply process waiting for a statement `>= endpos` which never happens, hence it never terminates.

### DIFF
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -3454,6 +3454,7 @@ pgsql_stream_logical(LogicalStreamClient *client, LogicalStreamContext *context)
 			}
 
 			/* call the keepaliveFunction callback now, ignore errors */
+			context->cur_record_lsn = walEnd;
 			context->now = client->now;
 
 			(void) (*client->keepaliveFunction)(context);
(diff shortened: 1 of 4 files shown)


shubhamdhama:
Draft as I will write a unit test when no activity on source happens (sort of like edge case).

dimitri (approved):

dimitri:
My reading of the Postgres source code makes me think that this PR is okay, we actually want to advance to WalEnd when receiving a keepalive message. See function WalSndUpdateProgress in src/backend/replication/walsender.c:

```
	 * When skipping empty transactions in synchronous replication, we send a
	 * keepalive message to avoid delaying such transactions.
```

About the test cases and the Keepalive (and other messages) position, we just skip comparing them because each run may provide a different one...

shubhamdhama:
Thanks for confirming that this fix is correct.

I've published the PR since this change is important and I'm couldn't finish the unit test I mentioned. I will share them in a separate PR once I am done.

dimitri:
It seems you have added an extra KeepAlive message in some of the tests output, and the expected file has not been updated by your PR so tests are failing. Please see about updating the expected tests results.

PR #163, opened 2022-12-08
tests: Add scenario when endpos is set between BEGIN and COMMIT.

shubhamdhama:
This test reproduces the issue mentioned in #154.

PR #166, opened 2022-12-12
Fix transform process to handle endpos in between transactions.

dimitri:
We might have a transaction that spans over multiple JSON files, and in that
case we always have a SWITCH WAL message at the end of the current file,
within the current transaction.

When we reach the end of file and we're in the middle of a transaction but
we don't have a SWITCH WAL message, it means that we have reached endpos. In
that case just ignore the currently opened transaction and replace its
contents by a synthetic KEEPALIVE message to advance the target system.

Closes #163


src/bin/pgcopydb/ld_transform.c
```diff
@@ -385,10 +388,42 @@ stream_transform_file(char *jsonfilename, char *sqlfilename)
...
+	 * It might happen that --endpos has been set to an LSN found in the middle
+	 * of a transaction, in that case we ignore the transaction and insert a
+	 * KEEPALIVE message with the LSN we have reached.
 	 */
-	if (currentTx->count > 0)
+	if (currentTx->count > 0 && metadata->action == STREAM_ACTION_SWITCH)
+	{
+		++txns.count;
+	}
+	else if (currentTx->count > 0 && metadata->action != STREAM_ACTION_COMMIT)
```
shubhamdhama, line 402: _Minor comment_: `&& metadata->action != STREAM_ACTION_KEEPALIVE` as well, it's not a big deal as we will anyway regenerate `K` action if we enter into this condition.

shubhamdhama (approved):
LGTM!

src/bin/pgcopydb/ld_transform.c
dimitri: Good point, I'm leaning towards keeping the code that way for the moment though. As you're saying, adding a synthetic K message doesn't change much of anything, and the code is easier to read. Might revisit later though.

PR #175, opened 2023-01-16
Fix huge memory allocations in copydb_prepare_table_specs.

shubhamdhama:
For a migration with a huge number of tables, I'm observing large heap memory allocation. In our case, it was 9K tables occupying 11.7 GB heap. This was happening because of,
https://github.com/dimitri/pgcopydb/blob/c2760a02322fc3e73c44a622d6cdb6a3338c1e10/src/bin/pgcopydb/copydb_schema.c#L283-L285

On investigation, I found that it was due to,
```sh
print sizeof(CopyTableDataSpec)
$1 = 1215480
# which has
print sizeof(TransactionSnapshot)
$7 = 1210040
# which has
print sizeof(LogicalStreamClient)
$8 = 1206768
```

I have changed a few members of `CopyTableDataSpec` to pointers which drastically reduced these allocations.

After this fix,
```sh
print sizeof(CopyTableDataSpec)
$1 = 5440
```

### DIFF
--- a/src/bin/pgcopydb/copydb.c
+++ b/src/bin/pgcopydb/copydb.c
@@ -709,14 +709,8 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
 		.cfPaths = &(specs->cfPaths),
 		.pgPaths = &(specs->pgPaths),
 
-		.source_pguri = { 0 },
-		.target_pguri = { 0 },
-		.sourceSnapshot = {
-			.pgsql = { 0 },
-			.pguri = { 0 },
-			.connectionType = specs->sourceSnapshot.connectionType,
-			.snapshot = { 0 }
-		},
+		.source_pguri = &(specs->source_pguri),
+		.target_pguri = &(specs->target_pguri),
 
 		.section = specs->section,
 		.resume = specs->resume,
@@ -730,17 +724,6 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
 
 		.indexSemaphore = &(specs->indexSemaphore)
 	};
-
-	/* initialize the connection strings */
-	strlcpy(tmpTableSpecs.source_pguri, specs->source_pguri, MAXCONNINFO);
-	strlcpy(tmpTableSpecs.target_pguri, specs->target_pguri, MAXCONNINFO);
-
-	/* initialize the sourceSnapshot buffers */
-	if (!copydb_copy_snapshot(specs, &(tmpTableSpecs.sourceSnapshot)))
-	{
-		/* errors have already been logged */
-		return false;
-	}
 
 	/* copy the structure as a whole memory area to the target place */
 	*tableSpecs = tmpTableSpecs;
@@ -910,8 +893,8 @@ copydb_fatal_exit()
 bool
 copydb_wait_for_subprocesses()
 {
-	bool allReturnCodeAreZero = true;
 
+	bool allReturnCodeAreZero = true;
 	log_debug("Waiting for sub-processes to finish");
 
 	for (;;)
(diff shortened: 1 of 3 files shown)


src/bin/pgcopydb/table-data.c
```diff
@@ -382,7 +382,7 @@ copydb_process_table_data_worker(CopyDataSpec *specs)
 		CopyTableDataSpec *tableSpecs = &(tableSpecsArray->array[tableIndex]);
 
 		/* reuse the same connection to the source database */
-		tableSpecs->sourceSnapshot = specs->sourceSnapshot;
+		tableSpecs->sourceSnapshot = &(specs->sourceSnapshot);
```
shubhamdhama, line 385: @dimitri if converting the `CopyTableDataSpec->sourceSnapshot` to a pointer seems correct, then we might want to remove this line.

src/bin/pgcopydb/copydb.c
```diff
@@ -709,14 +709,9 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
...
 		.pgPaths = &(specs->pgPaths),
 
-		.source_pguri = { 0 },
-		.target_pguri = { 0 },
-		.sourceSnapshot = {
-			.pgsql = { 0 },
-			.pguri = { 0 },
-			.connectionType = specs->sourceSnapshot.connectionType,
-			.snapshot = { 0 }
-		},
```
dimitri, line 719: Is is safe to share the same snapshot as in the main specs? if that's safe, is there a good reason to keep a snapshot instance in the CopyTableDataSpec at all?

src/bin/pgcopydb/copydb.c
```diff
@@ -709,14 +709,9 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
...
+		.source_pguri = specs->source_pguri,
+		.target_pguri = specs->target_pguri,
```
dimitri, line 713: Please use the `&(...)` notation for the pointer assignment, like we do in the rest of the code.

src/bin/pgcopydb/copydb.c
```diff
@@ -731,17 +726,6 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
...
-	strlcpy(tmpTableSpecs.source_pguri, specs->source_pguri, MAXCONNINFO);
-	strlcpy(tmpTableSpecs.target_pguri, specs->target_pguri, MAXCONNINFO);
-
-	/* initialize the sourceSnapshot buffers */
-	if (!copydb_copy_snapshot(specs, &(tmpTableSpecs.sourceSnapshot)))
-	{
-		/* errors have already been logged */
-		return false;
-	}
-
```
dimitri, line 744: This worries me. From discussion we might be able to remove the CopyTableDataSpec snapshot entirely, that would be good.

dimitri (changes requested):

dimitri:
I'm curious, since you shared the memory use before your changes, can you also share them after the changes?

src/bin/pgcopydb/copydb.c
shubhamdhama: `sourceSnapshot` isn't providing any value so I removed it.

src/bin/pgcopydb/copydb.c
shubhamdhama: Done

shubhamdhama:
> I'm curious, since you shared the memory use before your changes, can you also share them after the changes?

Updated in the description. After this change, it's occupying only `5440` bytes.

dimitri (approved):

PR #178, opened 2023-01-26
Fix working-directory issues in stream cleanup and other commands.

shubhamdhama:
Address the failure when `stream cleanup` is used after `clone follow` caused by `dir` is passed as `NULL` to `copydb_init_workdir`, which leads `copydb_prepare_filepaths` to set `topdir` to an undesired location.

Includes cleanup of commands that do not require working directory, by removing `copydb_init_workdir`. And, sets `dir` to appropriate value that do requires working directory.


dimitri (approved):

dimitri:
Thanks for this cleanup PR @shubhamdhama ; much appreciated!

PR #184, opened 2023-02-02
Fix same-table-copy query to use cached pgcopydb.table_size.

shubhamdhama:
Query to split the table for same table concurrency should use `pgcopydb.table_size` instead of using `pg_table_size`.


src/bin/pgcopydb/schema.c
```diff
@@ -2628,22 +2628,27 @@ schema_list_partitions(PGSQL *pgsql, SourceTable *table, uint64_t partSize)
...
-		"  ) "
+		"   select min(\"%s\"), max(\"%s\") "
+		"     from \"%s\".\"%s\" "
+		" ), "
+		" t (parts) as "
+		" ( "
+		"   select greatest(ceil(bytes::float / $1), 1) "
+		"     from pgcopydb.table_size "
+		"     where oid = %d "
+		" ), "
```
shubhamdhama, line 2641: I have reduced one space ident here. Let me know if this needs to be reverted.

dimitri: Does the subquery returns a value when the WHERE clause fails to match any row in the pgcopydb.table_size table? Maybe this warrants an additional test-case in the `tests/unit` test suite...

shubhamdhama:
@dimitri I wasn't sure how to maintain the essence of `pg_regress` and validate my changes using `sql/`/`expected/`. Hence I refactored `tests/unit` to allow shell scripts as test cases. Please let me know if there are other options. If the refactor looks good I may need to update `tests/unit/README.md`.

src/bin/pgcopydb/schema.c
shubhamdhama: I have revised this, please take a look.

dimitri: It's better to use UNION ALL in that case, because it skips the deduplication step that Postgres would have to implement with using UNION: we know we are not introducing duplicate data here. Of course in terms of performance, it is not going to be dramatic given we expect only one row in both branches of the UNION. It's more to do with semantics here.

tests/unit/tests/4-list-table-split.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+
+set -x
+set -e
+
+pgcopydb list table-parts --source "${PGCOPYDB_SOURCE_PGURI}" \
```
dimitri, line 6: You can omit `--source` in that case, pgcopydb will just use the environment variable directly.

tests/unit/copydb.sh
```diff
@@ -40,12 +38,21 @@ find .
 
 pgopts="--single-transaction --no-psqlrc --expanded"
 
-for f in ./sql/*.sql
+for f in ./tests/*.{sql,sh}
```
dimitri, line 41: Could we keep the `sql` directory, like they do in Postgres, and add a new directory for the bash scripts, maybe `script` or something? Then instead of dealing with the `if [[ ... ]]` in the loop body, we would have two loops, first the SQL one I suppose, and then the bash/script specialized one. Also would make it easier to deal with DIFFOPTS which at the moment is generalized in the call to `diff` but initialized only in one case...

shubhamdhama:
@dimitri could you please provide another round of review?

src/bin/pgcopydb/string_utils.c
```diff
@@ -35,6 +35,21 @@ intToString(int64_t number)
...
+uint32ToString(uint32_t number)
+{
+	UInt32String intString;
+
+	intString.intValue = number;
+
+	sformat(intString.strValue, UINT32STRING_MAX_DIGITS, "%" PRId32, number);
+
+	return intString;
+}
```
dimitri, line 52: Can we just use the existing `intToString` function here? It takes as an argument a `int64_t` number, and an OID (`uint_32t`) should fit within that range right?

src/bin/pgcopydb/string_utils.h
```diff
@@ -16,13 +16,22 @@
 
 /* maximum decimal int64 length with minus and NUL */
 #define INTSTRING_MAX_DIGITS 21
+#define UINT32STRING_MAX_DIGITS 10
```
dimitri, line 19: Not used.

src/bin/pgcopydb/string_utils.h
```diff
@@ -16,13 +16,22 @@
...
 {
 	int64_t intValue;
 	char strValue[INTSTRING_MAX_DIGITS];
 } IntString;
 
+typedef struct UInt32String
+{
+	uint32_t intValue;
+	char strValue[INTSTRING_MAX_DIGITS];
+} UInt32String;
```
dimitri, line 31: Same comment as with the function, I believe `int64_t` can host a `int32_t` value, in a way that we don't need to add new code for that conversion...

tests/unit/copydb.sh
```diff
@@ -27,10 +27,8 @@ create collation if not exists mycol
  );
 EOF
 
-pgcopydb list collations --source "${PGCOPYDB_SOURCE_PGURI}"
-
 # pgcopydb fork uses the environment variables
-pgcopydb fork --skip-collations --debug
+pgcopydb clone --skip-collations --debug
```
dimitri, line 31: Using the `pgcopydb fork` variant of the command was done on purpose here, to have some unit testing exercise that spelling of the command...

tests/unit/copydb.sh
```diff
@@ -45,7 +43,20 @@ do
     t=`basename $f .sql`
     r=/tmp/results/${t}.out
     e=./expected/${t}.out
-    psql -d "${PGCOPYDB_TARGET_PGURI}" ${pgopts} --file ./sql/$t.sql &> $r
+    psql -d "${PGCOPYDB_TARGET_PGURI}" ${pgopts} --file $f &> $r
```
dimitri, line 46: Please revert that change, I don't think we need to change anything on that line (looks like noise now).

tests/unit/copydb.sh
```diff
@@ -45,7 +43,20 @@ do
...
     diff $e $r || exit 1
 done
+
+
+for f in ./script/*.sh
+do
+    t=`basename $f .sh`
+    r=/tmp/results/${t}.out
+    e=./expected/${t}.out
+    bash $f $r
```
dimitri, line 57: I think it would be better that the test script outputs to stdout and we capture that here. In other words:

``` diff
- bash $f $r
+ bash "$f" > "$r"

File: tests/unit/script/4-list-table-split.sh [hunk 10]
```diff
@@ -0,0 +1,16 @@
+#! /bin/bash
+
+set -x
+set -e
+
```
dimitri, line 5: Let's add some comments about what we're trying to achieve (test the split decision and number of parts, and that table_3 is not to be split because we don't have an entry for it in our pgcopydb.table_size cache). Also please mention, as in the other scripts, that we're depending on the PGCOPYDB_SOURCE_PGURI environment variable in the script.

tests/unit/setup/4-list-table-split.sql
```diff
@@ -0,0 +1,56 @@
+create table table_1 (
```
dimitri, line 1: Here again let's add some comments about what the script intends to test and setup. It's not as obvious as other tests; it's only fair now because of the context of this PR. I believe we will later thank ourselves for adding comments that might seem a tad obvious just now.

tests/unit/copydb.sh
```diff
@@ -45,7 +43,20 @@ do
...
     r=/tmp/results/${t}.out
     e=./expected/${t}.out
-    psql -d "${PGCOPYDB_TARGET_PGURI}" ${pgopts} --file ./sql/$t.sql &> $r
+    psql -d "${PGCOPYDB_TARGET_PGURI}" ${pgopts} --file $f &> $r
     test -f $e || cat $r
     diff $e $r || exit 1
 done
+
+
+for f in ./script/*.sh
```
dimitri, line 52: This also warrants a new paragraph in the local README.md file.

dimitri (changes requested):
Thanks for that new version of the PR, it's quite better overall, nice progress. I left some comments where I believe more work is needed. Mostly comments, and also some new code that I think we don't need and can remove now.

src/bin/pgcopydb/string_utils.h
shubhamdhama: Yeah I was about to push but now removing it.

tests/unit/copydb.sh
dimitri: You can still use `2>&1` at the end of the `pgcopydb` call and still get the logs in the outfile, which I think was nicer anyways, because it serves as documentation too...

dimitri (approved):

PR #211, opened 2023-03-17
Create new process group for pgcopydb initialization.

shubhamdhama:
Previously, when `pgcopydb` encountered a FATAL error and called `copydb_fatal_exit`, any wrapper processes around `pgcopydb` would also get terminated due to the `if (kill(0, SIGTERM) == -1)` statement. To fix this issue, a new process group is now created for `pgcopydb` during initialization. This ensures that only the `pgcopydb` process will be terminated in case of a FATAL error, leaving the wrapper processes intact.


src/bin/pgcopydb/main.c
```diff
@@ -43,6 +43,14 @@ static void log_semaphore_unlink_atexit(void);
 int
 main(int argc, char **argv)
 {
+	/* create a new process group and set current process as its leader.  This
+ 	 * allows the process group to be easily controlled, without affecting any
```
dimitri, line 47: Nitpicking: please conform to the comment style used in the rest of the project. It would then look like this:

``` C
    /*
     * Create a new process group...
     * 
```

shubhamdhama: Done.

dimitri (approved):

PR #218, opened 2023-03-23
Inclusive schema name filtering, and file logging changes

devinshteyn2:
Please review the proposed changes


lospejos:
I upvote this PR, please consider to test and merge/approve. Thanks!

dimitri:
Thanks for your contributions @devinshteyn2 ! The work may now begin: we have a lot to talk about in your PR, I'll try to give you an overall review first and then use the GitHUB UI to have a detailed review. First thing I notice is that we should split the PR and have several with contained changes. For instance the log file is a good idea, but requires its own review (command line switch, rotate on SIGHUP signal, etc).

devinshteyn2:
About the log file, that was a workaround on my end to facilitate faster debugging for testing with many schema objects, and grepping logs for errors. 

IMHO, file logging is also a good thing for use in automation projects and pipelines where running things manually from console isn't an option, like in production environment. If you like the idea, I can add a new parameter, please name it.  But why stop there? I would also suggest adding a separate option for log formatting. Logging in plain text format isn't always the best option. Modern log monitoring solutions especially in cloud based environments prefer JSON for easy of log parsing and aggregation. It seems that the existing log library doesn't support that, it may need to be updated to support different formatting options. Optional suppression of color coding symbols might be desired too, so they don't get in the way of integrated log monitoring, in Splunk, CloudWatch, Azure Log Monitor, and likes.

devinshteyn2:
Looking forward to your PR review and further comments

dimitri:
Please have a look at #234 that implements your ideas about the log file, including JSON support. Note that we already have support for bypassing colors when we detect that stderr is not a tty, so that should be fine for logging pipelines already.

dimitri (changes requested):
Here is a first round of review. In the meantime several other PRs have made it to the main branch that require a rebase and your attention, with changes in areas you're also working on. In particular PR #247 introduces support for SourceFilters in dump_restore.c in a way that you can benefit now.

dimitri:
ping @devinshteyn2 ; do you think you will have more time to spend on finishing this PR?

.gitignore
```diff
@@ -41,3 +41,5 @@ lib*.pc
 /env/
 /GIT-VERSION-FILE
 /version
+/tests/schema-filter
+.vscode/
```
dimitri, line 45: Please refrain from adding personal preferences to the repository, add .vscode to your local .gitignore setup!

src/bin/pgcopydb/dump_restore.c
```diff
@@ -155,12 +157,27 @@ copydb_target_prepare_schema(CopyDataSpec *specs)
...
+		make sure they exist in the target database, if not create them.
+	 */
+	 if (specs->filters.includeOnlySchemaList.count > 0)
+	 {
+		if (!copydb_target_prepare_namespaces(specs))
+		{
+			/* errors have already been logged */
+			return false;
+		}
+	 }
```
dimitri, line 170: I thought we would take care of that automatically within the pg_dump and pg_restore calls. If that's not happening, I think the comment ought to say why and give some context. I would prefer that we don't have to create the schemas here, that said, because what if the user made a typo in their filtering setup? Then we create unwanted schemas on the target databases...

Also nitpicking: Please follow the code style in use all around in the same code/file. It looks like this:

``` c
	/* 
     * If restoring specific schemas as specified in the inclusion filter
	 *	make sure they exist in the target database, if not create them.
	 */
```

Note the capital, empty lines, and stars at the beginning of all lines within the comment.

src/bin/pgcopydb/main.c
```diff
@@ -278,6 +280,27 @@ set_logger()
...
+			{
+				log_set_fp(log_fp);
+			}
+			else
+			{
+				log_error("Failed to open log file for writing %s. Error reason: %s", env_log_file, strerror(errno));
+				exit(EXIT_CODE_BAD_ARGS);
+			}
+		}
+	}
```
dimitri, line 303: Please remove that parts, the corresponding feature has been added in the meantime already. See https://pgcopydb.readthedocs.io/en/latest/ref/pgcopydb_clone.html#environment and PGCOPYDB_LOG_TIME_FORMAT, PGCOPYDB_LOG_JSON, PGCOPYDB_LOG_FILENAME, and PGCOPYDB_LOG_JSON_FILE.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -829,6 +880,9 @@ pg_restore_db(PostgresPaths *pgPaths,
 		args[argsIndex++] = (char *) listFilename;
 	}
 
+	/* verbose output */
+	//args[argsIndex++] = "--verbose";
+
```
dimitri, line 885: I suspect this is meant to be removed later, before final review and merge? You might also be able to use `log_get_level` to selectively add `--verbose` to the pg_dump call when using --debug or --trace at the pgcopydb level?

src/bin/pgcopydb/pgcmd.h
```diff
@@ -17,8 +17,13 @@
 #include "file_utils.h"
 #include "filtering.h"
 #include "pgsql.h"
+#include "filtering.h"
+
 
 #define PG_VERSION_STRING_MAX 12
+#define PG_CMD_MAX_ARG 256
+#define PG_DUMP_CMD_RESERVED_ARG 10
+#define PG_RESTORE_CMD_RESERVED_ARG 15
```
dimitri, line 26: Nice addition, still deserves some commenting.

src/bin/pgcopydb/schema.c
```diff
@@ -428,13 +429,18 @@ struct FilteringQueries listSourceTableSizeSQL[] = {
...
+		"         left join pg_temp.filter_include_only_schema fn "
+		"                on n.nspname = fn.nspname "
+
 		/* include-only-table */
-		"         join pg_temp.filter_include_only_table inc "
-		"           on n.nspname = inc.nspname "
-		"          and c.relname = inc.relname "
+		"         left join pg_temp.filter_include_only_table ft "
+		"                on n.nspname = ft.nspname "
+		"               and c.relname = ft.relname "
```
dimitri, line 439: Why using a LEFT JOIN for an “include-only” filter implementation? I expected a JOIN here.

dimitri (changes requested):
Thanks for taking the time and efforts to prepare this PR for merge, still some more to do!

src/bin/pgcopydb/schema.c
dimitri: Ah yeah, nice trick!

dimitri:
I just found some time to review this PR again, and when playing with the SQL queries and filtering I realized the behaviour was a little suprising. Thinking more, what I wanted from this feature is the exact same behavior as the already existing "exclude-schema" filter, just built the other way around by listing the schemas to filter-in rather than the schema to filter-out.

Maybe the best way to explain it is this part of the implementation for the new "include-only-schema":

``` sql
insert into pg_temp.filter_exclude_schema
     select n.nspname
       from pg_namespace n
  left join pg_temp.filter_include_only_schema inc
         on n.nspname = inc.nspname
      where inc.nspname is null;
```

devinshteyn2:
Thank you very much for merging this PR! 

If I understand correctly, you expected an inverse filter, having multiple exclusion filters inserted into the filter table, leaving just one schema (or a few) to be copied. Perhaps using that inverse method required much less code changes, can't argue with that. In our use case, we deal with hundreds sometimes thouthands of schemas in a single database and we copy just one when we need to move it to a different server. Exclusion filters feel very non intuitive in such cases. Maybe that's why to me the inverse method feels like... if I go to a grocery store to buy an apple, I bring all other apples found in the store to the cashier and tell them these are the apples I don't want to buy  :-)

dimitri:
I agree it's counter-intuitive. 

Making the SQL query actually work with the "include-only-schema" approach was also very complex, because when the `filter_include_only_schema` table is empty (0 rows) you want to keep everything rather than filter-out everything, and inverting the filter condition based on how a table being empty or not in SQL is... well I couldn't find a nicer way to do it than inverting the filter and re-using a framework and a set of queries that all work already.

PR #234, opened 2023-04-05
Implement the feature to log to file, with support for JSON.

dimitri:
This allows pgcopydb to log both to the terminal and also to a file, where the format is not adjusted for terminal width (long time format, source file and line always included).

Also implements logging in JSON, using a JSON object per line with a static set of attributes that includes all the metadata we have about the log message.

The control of this feature set is entirely based on using environment variables, no new command line switch has been introduced. It is possible to log both in plain text with colors on the terminal and to a JSON file at the same time.


devinshteyn2 (commented):
Couple of ideas:
In case pgcopydb is used in some sort of automation pipeline handled by a group of machines or kind of serverless, and logs are monitored and aggregated, it would be nice to know machine name. Maybe give users an option to add machine name?

Aligning log tag names with PG tag log names? ts -> timestamp, log -> message, level -> severity?
https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-LOGGING-JSONLOG

devinshteyn2:
If you would like me to make these changes, I'd be happy to do it if you can wait for a couple of weeks, I'm literally going on vacation tomorrow, and got my plate full for today :-)  Yet, they are fairly small, probably makes more sense if somebody else adds them.

dimitri:
Done. Have a good vacation time ;-)

PR #278, opened 2023-05-05
Update pgcopydb_stream.rst

thecatontheflat:


dimitri:
The command line help string also needs to be updated, see https://github.com/dimitri/pgcopydb/blob/19f72f6a81564164ee523dede6f66489b0757616/src/bin/pgcopydb/cli_stream.c#L116 etc. Also, did you check the other sub-commands of `pgcopydb stream` in passing?

PR #287, opened 2023-05-12
Fix follow_prepare_mode_switch not finishing when previous mode is REPLAY.

shubhamdhama:
When the previous streaming mode was REPLAY and we attempt to empty the transform queue using `stream_transform_from_queue`, the function waits indefinitely for `QMSG_TYPE_STOP` (or any message) because the queue has already been emptied.

### DIFF
--- a/src/bin/pgcopydb/follow.c
+++ b/src/bin/pgcopydb/follow.c
@@ -274,6 +274,7 @@ follow_main_loop(CopyDataSpec *copySpecs, StreamSpecs *streamSpecs)
 
 	uint64_t loop = 0;
 	LogicalStreamMode currentMode = modeArray[0];
+	LogicalStreamMode previousMode = STREAM_MODE_UNKNOW;
 
 	while (true)
 	{
@@ -297,6 +298,7 @@ follow_main_loop(CopyDataSpec *copySpecs, StreamSpecs *streamSpecs)
 		}
 
 		/* switch to the next mode, increment loop counter */
+		previousMode = currentMode;
 		currentMode = modeArray[++loop % count];
 
 		/* and re-init our streamSpecs for the new mode */
@@ -311,7 +313,7 @@ follow_main_loop(CopyDataSpec *copySpecs, StreamSpecs *streamSpecs)
 		 * ensure to catch-up with files on-disk before switching
 		 * to another mode of operations.
 		 */
-		if (!follow_prepare_mode_switch(streamSpecs))
+		if (!follow_prepare_mode_switch(streamSpecs, previousMode))
 		{
 			/* errors have already been logged */
 			return false;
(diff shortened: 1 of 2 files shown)


src/bin/pgcopydb/follow.c
```diff
@@ -388,11 +388,18 @@ follow_prepare_mode_switch(StreamSpecs *streamSpecs)
 		}
 	}
 
-	/* first empty the transform queue, where a STOP message has been sent */
-	if (!stream_transform_from_queue(streamSpecs))
+	/*
+	 * If the current mode is replay, it means the previous mode was catch-up.
+	 * Therefore, we need to empty the transform queue before proceeding, where
+	 * a STOP message has been sent.
```
dimitri, line 394: That's a little confusing. I think it would be better if we would pass previousMode as well as currentMode (or maybe nextMode) to the function. At the moment we only have two modes of operations and we always go from the one to the next, but that could change and passing both args is good future-proofing. And will make this code much easier to reason about, too.

dimitri (changes requested):

src/bin/pgcopydb/follow.c
shubhamdhama: Make sense, I'm going with `currentMode` and `previousMode`, I'm not sure how `nextMode` could fit here.

Though I have one doubt related to my change, is it guaranteed that the transform queue would be empty if say during the replay the transform process got interrupted while still have some messages in the queue?

src/bin/pgcopydb/follow.c
```diff
@@ -388,11 +394,19 @@ follow_prepare_mode_switch(StreamSpecs *streamSpecs)
...
 
-	/* first empty the transform queue, where a STOP message has been sent */
-	if (!stream_transform_from_queue(streamSpecs))
+	/*
+	 * If the current mode is replay and the previous mode was catch-up, then
+	 * before proceeding, we need to empty the transform queue where the STOP
+	 * message was sent.
+	 */
+	if (currentMode == STREAM_MODE_REPLAY &&
+		previousMode == STREAM_MODE_CATCHUP)
```
dimitri, line 403: I believe the condition  that previousMode == CATCHUP is all we need to make that decision, whatever the current/next mode is going to be, right?

dimitri (changes requested):

src/bin/pgcopydb/follow.c
dimitri: The transform queue is used to send LSN positions that belong to new JSON files, and triggers the transformation of an entire file at a time from JSON to SQL. In replay mode, a Unix PIPE is used in which we write the JSON messages, and the transform process reads JSON messages from the other end of the same pipe. The transform queue is never used when in replay mode.

I am not sure I got the question right, but it seems to me that the answer is “won't happen”.

src/bin/pgcopydb/follow.c
shubhamdhama: Yep, I also had a similar thought, but then by "future-proofing" I assumed we only want to empty the queue if both the conditions are met (which doesn't make much sense but who knows what the future holds :). Let me change this right away.

dimitri (approved):

Issue #241, opened 2023-04-13
Errors with decoding


aborschev:
Hi! 
I've got a test stand, with 2 VMs Debian Bulleye 4 CPU 4 GB RAM, one is with source DB, another-target.
pgcopydb 0.11 is on target host.

Source: PG15 vanilla.  10 GB  of some data + pgbench.
I have emulated  workload with pgbench:
```
postgres@borschev-pg-copydb1:~$ /usr/lib/postgresql/10/bin/pgbench --builtin=tpcb-like --client=4 --jobs=4 --time=600 --rate=20 --port=5433 --host=localhost --progress=20  --user=pgbench demo
starting vacuum...end.
progress: 20.0 s, 10.1 tps, lat 10.034 ms stddev 16.880, lag 0.896 ms
progress: 40.0 s, 10.0 tps, lat 11.158 ms stddev 27.644, lag 0.611 ms
...
```

- i.e. the workload not very high, only 20 transactions per second. 

I tried pgcopydb both  with --plugin=test_decoding. 

```
postgres@borschev-pg-copydb2:~$ pgcopydb clone --table-jobs=4 --index-jobs=4 --plugin=test_decoding --source="postgresql://postgres:pgpass@192.168.21.10:5433/demo" --target="postgresql://postgres:pgpass@localhost:5432/demo" --follow
12:54:18 42260 INFO   Running pgcopydb version 0.11-1.pgdg110+1 from "/usr/bin/pgcopydb"
12:54:18 42260 INFO   Using default origin node name "pgcopydb"
12:54:18 42260 INFO   [SOURCE] Copying database from "postgres://postgres@192.168.21.10:5433/demo?password=****"
12:54:18 42260 INFO   [TARGET] Copying database into "postgres://postgres@localhost:5432/demo?password=****"
12:54:18 42260 INFO   Exported snapshot "00000007-0000095F-1" from the source database
12:54:18 42260 INFO   Created logical replication slot "pgcopydb" with plugin "test_decoding" at LSN 2/BCB3CF68
12:54:18 42260 INFO   create schema if not exists pgcopydb
12:54:18 42260 INFO   drop table if exists pgcopydb.sentinel
12:54:18 42260 WARN   NOTICE:  table "sentinel" does not exist, skipping
[...]

PR #290, opened 2023-05-15
Use column names in COPY statements.

### LINKED ISSUE #275: pgcopydb not using the column names while copying data
pg_dump is using command like this to copy data from source database table

**COPY public.part_config (parent_table, control, partition_type, partition_interval, constraint_cols, premake, optimize_trigger, optimize_constraint, epoch, inherit_fk, retention, retention_schema, retention_keep_table, retention_keep_index, infinite_time_partitions, datetime_string, use_run_maintenance, jobmon, sub_partition_set_full, undo_in_progress, trigger_exception_handling, upsert, trigger_return_null) FROM stdin;**

pgcopydb is performing like this 

**copy (SELECT * FROM "public"."part_config" ) to stdout;
copy "public"."part_config" from stdin;
COPY part_config, line 1: "public.time_taptest_table        col3   native   1 day   \N      4       4       30      none    t       \N      \N      t       t       f       YYYY_MM_DD     on       t       f       f       f               t       pu..."
missing data for column "inherit_privileges"
TARGET Context: Failed to copy data to target**

Because of this if there are any newly added columns to that table that are having default value they are not getting copied and giving error like
**missing data for column "inherit_privileges"
TARGET Context: Failed to copy data to target**

More logs can be found here [partman.txt](https://github.com/dimitri/pgcopydb/files/11381571/partman.txt)

dimitri:
This requires loading more of the Postgres catalog information in-memory. That's the cost to fix extension configuration tables that have changed schema between source and target, and to later support schema changes at pgcopydb time.

Fixes #275.
See #241.


thecatontheflat:
This is amazing! Thank you so much for adding so many useful features!

I was missing this feature for changing the columns order for our 20-TB DB. Looks like I'll be able to do it soon! :)

Issue #282, opened 2023-05-09
First few prefetched rows not applied to target even after transformation


shubhamdhama:
Pgcopydb version : Latest
Reproducing this was easy for me, just have a high rate of transactions on the source. Once, the online migration is complete, only the first few rows will be missed on the target, in my case it was 1 row the first time and 4 rows the second time.

Here is the `transform` line,
```log
2023-05-05 15:46:10 39440 INFO   follow.c:253              Current sentinel replay_lsn is 0/0
2023-05-05 15:46:10 39440 INFO   follow.c:408              Catching-up from existing on-disk files
2023-05-05 15:46:10 39440 TRACE  pgsql.c:3059              parseTimeLineHistory[0]: tli 1 [0/0 0/0]
2023-05-05 15:46:10 39440 DEBUG  ld_transform.c:464        stream_transform_from_queue: 2B/800092E8
2023-05-05 15:46:10 39440 TRACE  ld_transform.c:553        stream_compute_pathnames: 2B/800092E8: 000000010000002B00000080
2023-05-05 15:46:10 39440 NOTICE ld_transform.c:619        Transforming JSON file "/datadrive/migrations_data/abde5c8c-46d6-4132-be49-9cbee56523c9/oid-32227/cdc/000000010000002B00000080.json" into SQL file "/datadrive/migrations_data/abde5c8c-46d6-4132-be49-9cbee56523c9/oid-32227/cdc/000000010000002B00000080.sql"
2023-05-05 15:46:10 39440 DEBUG  ld_transform.c:641        stream_transform_file: read 806 lines from "/datadrive/migrations_data/abde5c8c-46d6-4132-be49-9cbee56523c9/oid-32227/cdc/000000010000002B00000080.json"
2023-05-05 15:46:10 39440 TRACE  ld_transform.c:686        stream_transform_file[ 0]: {"action":"B","xid":"205372","lsn":"2B/800092E8","timestamp":"2023-05-05 10:14:52.264545+0530","message":{"action":"B","xid":205372}}
[...]

PR #295, opened 2023-05-22
Fix issue with missing data from transactions with BEGIN LSN less than consistent_point.

shubhamdhama:
Currently we skip applying SQL statements whose source/origin LSN is less than `previousLSN`. When using `pgcopydb clone --follow`, the initial `previousLSN` matches the`consistent_point` obtained during the snapshot.

Comparing LSN of each SQL statement creates a problem when parallel transactions are running during the snapshot capture. These transactions could have statements whose LSN is less than the `previousLSN`. Consequently, the initial portion of some transactions, containing statements with LSN less than the `consistent_point`, are inadvertently omitted.

The fix here is to compare the COMMIT LSN instead of each statement's LSN as it is guaranteed that the COMMIT LSN of every streamed message will be greater than `consistent_point`.

Fixes: #282

### DIFF
--- a/tests/follow-data-only/copydb.sh
+++ b/tests/follow-data-only/copydb.sh
@@ -20,10 +20,21 @@ psql -d ${PGCOPYDB_TARGET_PGURI} -f /usr/src/pgcopydb/ddl.sql
 # insert a first batch of 10 rows (1..10)
 psql -v a=1 -v b=10 -d ${PGCOPYDB_SOURCE_PGURI} -f /usr/src/pgcopydb/dml.sql
 
+# take a snapshot with concurrent activity happening it would be hard to sync
+# concurrent activity in the inject service, so use a job instead
+bash ./run-background-traffic.sh &
+BACKGROUND_TRAFFIC_PID=$!
+
+# wait for few seconds to allow snapshot to happen in between the traffic
+sleep 2
+
 # grab a snapshot on the source database
 coproc ( pgcopydb snapshot --follow --plugin wal2json --notice )
 
-sleep 1
+sleep 2
+
+# stop the background traffic
+kill -TERM ${BACKGROUND_TRAFFIC_PID}
 
 # check the replication slot file contents
 cat /var/lib/postgres/.local/share/pgcopydb/slot
--- a/src/bin/pgcopydb/ld_apply.c
+++ b/src/bin/pgcopydb/ld_apply.c
@@ -496,26 +496,43 @@ stream_apply_sql(StreamApplyContext *context,
 
 		case STREAM_ACTION_BEGIN:
 		{
-			/* did we reach the starting LSN positions now? */
-			if (!context->reachedStartPos)
-			{
-				context->reachedStartPos =
-					context->previousLSN < metadata->lsn;
-			}
-
-			log_trace("BEGIN %lld LSN %X/%X @%s, previous LSN %X/%X %s",
-					  (long long) metadata->xid,
-					  LSN_FORMAT_ARGS(metadata->lsn),
-					  metadata->timestamp,
-					  LSN_FORMAT_ARGS(context->previousLSN),
-					  context->reachedStartPos ? "" : "[skipping]");
-
 			if (metadata->lsn == InvalidXLogRecPtr ||
+				metadata->txnCommitLSN == InvalidXLogRecPtr ||
 				IS_EMPTY_STRING_BUFFER(metadata->timestamp))
 			{
 				log_fatal("Failed to parse BEGIN message: %s", sql);
 				return false;
 			}
+
+			/* did we reach the starting LSN positions now? */
+			if (!context->reachedStartPos)
+			{
+				/*
+				 * compare previousLSN with COMMIT LSN to safely include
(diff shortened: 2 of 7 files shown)


dimitri (changes requested):
Thanks! please edit the main comment of that PR as indicated below.

shubhamdhama:
We have two important follow-ups for this PR,
1. We need to take care of `<` and `<=`.
2. If the transaction spills to the next file, then we don't have COMMIT LSN.

tests/follow-data-only/copydb.sh
```diff
@@ -20,10 +20,21 @@ psql -d ${PGCOPYDB_TARGET_PGURI} -f /usr/src/pgcopydb/ddl.sql
 # insert a first batch of 10 rows (1..10)
 psql -v a=1 -v b=10 -d ${PGCOPYDB_SOURCE_PGURI} -f /usr/src/pgcopydb/dml.sql
 
+# run concurrent transactions in background and take snapshot in between to
+# verify we copy newly inserted data when the snapshot is taken
+bash ./run-background-traffic.sh &
+BACKGROUND_TRAFFIC_PID=$!
```
dimitri, line 26: Is there a way to move the background traffic to the inject service that runs in the background already?

shubhamdhama: I want to have on-going traffic for at least some time before we take the snapshot (2 seconds here), that can be easily achieved here IMO. What do you think?

dimitri: I see. Then let's add a comment that's what is happening, and that having control over the sequence of events is not a good fit for multiple services. 

Also, can we use the bash `coproc` facility more than once? I believe we can give each coproc a name, see https://www.gnu.org/software/bash/manual/html_node/Coprocesses.html. Would be best to use the same background job mechanism twice rather than two different ways to do the same thing.

shubhamdhama: I tried `coproc` at first, but it was giving this error,
```
warning: execute_coproc: coproc still exists
```
on a quick search, it seems we can't have a parallel `coproc` without closing an existing one.

dimitri: You need to use the coproc syntax with a NAME, you can't run multiple coproc with the same name.

shubhamdhama: Yes, I gave them different unique names. Let me see if I can still reproduce that.

shubhamdhama: Oh it worked but it still gave me this warning
```
/usr/src/pgcopydb/copydb.sh: line 31: warning: execute_coproc: coproc [35:BACKGROUND_TRAFFIC] still exists
```

here are the changes
```diff
diff --git a/tests/follow-data-only/copydb.sh b/tests/follow-data-only/copydb.sh
index 594e699..fc340b7 100755
--- a/tests/follow-data-only/copydb.sh
+++ b/tests/follow-data-only/copydb.sh
@@ -22,14 +22,13 @@ psql -v a=1 -v b=10 -d ${PGCOPYDB_SOURCE_PGURI} -f /usr/src/pgcopydb/dml.sql

 # run concurrent transactions in background and take snapshot in between to
 # verify we copy newly inserted data when the snapshot is taken
-bash ./run-background-traffic.sh &
-BACKGROUND_TRAFFIC_PID=$!
+coproc BACKGROUND_TRAFFIC ( bash ./run-background-traffic.sh )

 # wait for few seconds to allow snapshot to happen in between the traffic
 sleep 2

 # grab a snapshot on the source database
-coproc ( pgcopydb snapshot --follow --plugin wal2json --notice )
+coproc SNAPSHOT ( pgcopydb snapshot --follow --plugin wal2json --notice )

 sleep 2

@@ -51,8 +50,8 @@ pgcopydb follow --notice
 # the follow command ends when reaching endpos, set in inject.sh
 pgcopydb stream cleanup

-kill -TERM ${COPROC_PID}
-wait ${COPROC_PID}
+kill -TERM ${SNAPSHOT_PID}
+wait ${SNAPSHOT_PID}

 # check how many rows we have on source and target
 sql="select count(*), sum(some_field) from table_a"
```

shubhamdhama: > Then let's add a comment that's what is happening

I have already added some comments, I think they are enough, can you please suggest what are we missing?

> and that having control over the sequence of events is not a good fit for multiple services.

Can you please elaborate?

dimitri: Okay Sorry about that, I am now reading more about coproc and bash seems to only guarantee coproc for a single sub-process at a time. Let's keep your initial implementation. About the comment, here is my suggestion:

``` diff
- # run concurrent transactions in background and take snapshot in between to
- # verify we copy newly inserted data when the snapshot is taken
+ # take a snapshot with concurrent activity happening
+ # it would be hard to sync concurrent activity in the inject service, so use a job instead
```

shubhamdhama: I have made the changes, please reivew.

dimitri (approved):

PR #298, opened 2023-05-26
Fix file rotation issue during streaming.

shubhamdhama:
Fixes a bug that occurs when multiple transactions have BEGIN and other statements in one WAL file, while their COMMITs are in the next file. The issue arises when pgcopydb receives the COMMIT of the first transaction, causing it to move to the next file. However, upon receiving the BEGIN of the next transaction, it incorrectly goes back to the previous file, resulting in an inconsistent file rotation pattern.

To address this issue, this commit introduces tracking the LSN of the last message written to the file. With this, file rotation is skipped if the incoming message has an LSN less than the lastWrittenLSN, ensuring that writes consistently occur in the current or next file.


src/bin/pgcopydb/ld_stream.c
```diff
@@ -312,7 +312,7 @@ stream_init_context(StreamContext *privateContext, StreamSpecs *specs)
 	privateContext->previous.action = STREAM_ACTION_UNKNOWN;
 
 	privateContext->lastWrite = 0;
-
+	privateContext->lastWrittenLSN = InvalidXLogRecPtr;
```
shubhamdhama, line 315: I could have avoided `if...else` later in the PR, only if I could find a better name by which I could have assigned `lastWrittenLSN` to `startpos` and that conveys correct intention.
It still seem fine to me, but I'm open to better names even with the current intentions.

src/bin/pgcopydb/ld_stream.c
```diff
@@ -767,8 +772,54 @@ streamRotateFile(LogicalStreamContext *context)
 		return true;
 	}
 
-	/* compute the WAL filename that would host the current LSN */
-	XLByteToSeg(context->cur_record_lsn, segno, context->WalSegSz);
+	/*
+	 * calculate the lsn based upon which we determine the walFileName
+	 */
+	uint64_t lsn;
```
shubhamdhama, line 778: After countless minutes of thinking I couldn't think of a better name :)

shubhamdhama (commented):
- [ ] Add tests

src/bin/pgcopydb/ld_stream.c
```diff
@@ -704,6 +704,11 @@ stream_write_json(LogicalStreamContext *context, bool previous)
 	/* time to update our lastWrite mark */
 	privateContext->lastWrite = time(NULL);
 
+	if (privateContext->lastWrittenLSN < metadata->lsn)
+	{
+		privateContext->lastWrittenLSN = metadata->lsn;
+	}
+
```
dimitri, line 711: This requires some comments to explain what's going on here...

src/bin/pgcopydb/ld_stream.c
```diff
@@ -767,8 +772,54 @@ streamRotateFile(LogicalStreamContext *context)
...
+		else if (context->cur_record_lsn < privateContext->lastWrittenLSN)
+		{
+			lsn = privateContext->lastWrittenLSN;
+		}
+		else
+		{
+			log_error("lastWrittenLSN %X/%X can't be equal to "
+					  "cur_record_lsn %X/%X",
+					  LSN_FORMAT_ARGS(privateContext->lastWrittenLSN),
+					  LSN_FORMAT_ARGS(context->cur_record_lsn));
```
dimitri, line 808: Reword message, at the moment you're saying it can't be but it is. Suggestion: 

    "Failed to compare lastWrittenLSN %X/%X with current record LSN %X/%X, expected different positions"

src/bin/pgcopydb/ld_stream.c
```diff
@@ -767,8 +772,54 @@ streamRotateFile(LogicalStreamContext *context)
...
+			return false;
+		}
+	}
+	else if (privateContext->startpos < context->cur_record_lsn)
+	{
+		lsn = context->cur_record_lsn;
+	}
```
dimitri, line 815: Requires a comment about what's happening here. I suppose that if we didn't write anything yet, we want to start at startpos, but the way the code is written something else is happening here and I'm not sure what at first reading.

src/bin/pgcopydb/ld_stream.h
```diff
@@ -101,6 +101,8 @@ typedef struct StreamContext
 	LogicalMessageMetadata metadata;
 	LogicalMessageMetadata previous;
 
+	uint64_t lastWrittenLSN;	/* LSN of last message written to the file */
+
 	uint64_t lastWrite;
```
dimitri, line 106: Add a comment that lastWrite is a timestamp, or change the name of the variable to `lastWriteTime`, or both.

src/bin/pgcopydb/ld_stream.c
dimitri: At least expand the comment to explain the logic behind not always using the cur_record_lsn for the computation of the WAL filename.

What I had in mind personally was to compute the WAL filename anyway and then ignore it when it's in the past. That can be determined by a `strcmp` or the filename + 8 (skip the timeline). That said working with LSN is probably better here (we know the timeline can't change without a full reconnect anyway).

dimitri: Variable name candidate: `jsonFileLSN` ; allows the comment to explain that the WAL LSN is monotonous but the logical decoding protocol only sends transactions at commit time, and then concurrent transaction interleaving means that we don't always receive the LSNs in order.

src/bin/pgcopydb/ld_apply.c
```diff
@@ -421,7 +422,7 @@ stream_apply_file(StreamApplyContext *context)
 
 		LogicalMessageMetadata metadata = { 0 };
 
-		if (!parseSQLAction(sql, &metadata))
+		if (!parseSQLAction(context, sql, &metadata))
```
dimitri, line 425: Same question about passing down a whole context... how could we avoid that?

src/bin/pgcopydb/ld_stream.c
```diff
@@ -767,8 +775,49 @@ streamRotateFile(LogicalStreamContext *context)
 		return true;
 	}
 
-	/* compute the WAL filename that would host the current LSN */
-	XLByteToSeg(context->cur_record_lsn, segno, context->WalSegSz);
+	/*
+     * determine the LSN to calculate walFileName in which to write the current
+     * message.
```
dimitri, line 780: single line comments start lowercase.

Multiple lines comments are actual English sentences and start with a capital letter. I know it's nitpicking, that said if I don't share that information I just make it hard for you to get to it ;-)

src/bin/pgcopydb/ld_stream.c
```diff
@@ -767,8 +775,49 @@ streamRotateFile(LogicalStreamContext *context)
...
+	 * This walFileName calculation later ensures safe transaction formation in
+	 * the transform/apply process by always appending messages here to the
+	 * latest file and preventing rotation to earlier files.
+	 *
+	 * In most cases, jsonFileLSN should be the same as cur_record_lsn. However,
+	 * occasionally, current messages may have LSNs lower than the previous
+	 * ones. This can occur due to concurrent transactions with interleaved
+	 * LSNs. Since the logical decoding protocol sends the complete transaction
+	 * at commit time, the LSNs for messages within one transaction may be lower
+	 * than those of the previous transactions. In such cases, we use maximum
```
dimitri, line 791: I think it's more helpful to say: the LSNs for messages within one transaction may be lower than the COMMIT lsn of the previously streamed transaction. At least for me it makes things more obvious/immediate. See what you feel about it.

src/bin/pgcopydb/ld_stream.h
```diff
@@ -121,6 +123,7 @@ typedef struct StreamContext
 typedef struct StreamApplyContext
 {
 	CDCPaths paths;
+	LogicalStreamMode mode;
```
dimitri, line 126: I am not quite sure yet in this review why that's needed?

src/bin/pgcopydb/ld_stream.h
```diff
@@ -494,6 +501,7 @@ bool stream_write_insert(FILE *out, LogicalMessageInsert *insert);
 bool stream_write_truncate(FILE *out, LogicalMessageTruncate *truncate);
 bool stream_write_update(FILE *out, LogicalMessageUpdate *update);
 bool stream_write_delete(FILE * out, LogicalMessageDelete *delete);
+bool stream_write_txn_metadata(FILE *out, LogicalTransaction *tx);
```
dimitri, line 504: I think that's the wrong place to take care of the xid.json metadata file.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -180,14 +180,18 @@ stream_transform_line(void *ctx, const char *line, bool *stop)
...
 		{
 			/* errors have already been logged */
 			return false;
 		}
 
 		/* now write the transaction out also to file on-disk */
-		if (!stream_write_message(privateContext->sqlFile, currentMsg))
+		if (!stream_write_message(privateContext,
+								  privateContext->sqlFile,
+								  currentMsg))
```
dimitri, line 194: I didn't expect this to change in the context of your PR. I believe that's the place where we should consider also writing the xid.json file, when the message is the STREAM_ACTION_COMMIT of a continued transaction. Otherwise there is this whole business of distinguishing between writing to a file or to stdout, and we don't want the lower-level API to know about that, all they know about is a FILE * output stream.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -597,8 +601,17 @@ stream_transform_send_stop(Queue *queue)
  * target database.
  */
 bool
-stream_transform_file(char *jsonfilename, char *sqlfilename)
+stream_transform_file(StreamSpecs *specs, char *jsonfilename, char *sqlfilename)
```
dimitri, line 604: Given previous comment, I believe we should be able to remove all the changes here.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1493,31 +1512,106 @@ stream_write_transaction(FILE *out, LogicalTransaction *txn)
 		}
 	}
 
+	/*
+	 * If we're in a continued transaction, it means that the earlier write of
+	 * this txn's BEGIN statement didn't have the COMMIT LSN. Therefore, we need
+	 * to maintain that LSN as a separate metadata file. This is necessary
+	 * because the COMMIT LSN is required later in the apply process.
+	 */
```
dimitri, line 1520: Again, not the right place to implement this bit.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1493,31 +1512,106 @@ stream_write_transaction(FILE *out, LogicalTransaction *txn)
...
+		if (!stream_write_txn_metadata(txnFile, txn))
+		{
+			return false;
+		}
+
+		if (fclose(txnFile) == EOF)
+		{
+			log_error("Failed to write file \"%s\"", txnfilename);
+			return false;
+		}
```
dimitri, line 1556: Should be a single call to `write_file`. We shouldn't be using the `stream_write_...` approach here.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1493,31 +1512,106 @@ stream_write_transaction(FILE *out, LogicalTransaction *txn)
...
+{
+	int ret =
+		fformat(out,
+				"{\"xid\":%lld,\"lsn\":\"%X/%X\",\"timestamp\":\"%s\"}\n",
+				(long long) txn->xid,
+				LSN_FORMAT_ARGS(txn->commitLSN),
+				txn->timestamp);
+
+	return ret != -1;
+}
```
dimitri, line 1576: Let's avoid that entirely and have a specialized function that writes to the xid.json file directly, without this indirection.

dimitri (changes requested):
I believe most of the lower-level API changes that you made are due to misplacing parts of the implementation, namely where to handle writing the xid.json file. See my comments about that, there is an opportunity for simplifying your PR a lot there.

src/bin/pgcopydb/cli_stream.c
```diff
@@ -920,7 +920,7 @@ cli_stream_transform(int argc, char **argv)
 			exit(EXIT_CODE_INTERNAL_ERROR);
 		}
 	}
-	else if (!stream_transform_file(jsonfilename, sqlfilename))
+	else if (!stream_transform_file(&specs, jsonfilename, sqlfilename))
```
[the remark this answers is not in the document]
shubhamdhama: Yes you make sense, maybe I just overthinked here, that earlier it is read as here is sqlfilename and jsonfilename, and let's transform. Adding Specs felt more natural and extensible than adding a random directory whose use comes much later.

But again it's just my overthinking so I'm going to change with what you proposed as at least that way we can avoid sending complete specs

src/bin/pgcopydb/ld_stream.c
shubhamdhama: I understand the importance of this. Thank you for point this out, I'll keep this in mind.

src/bin/pgcopydb/ld_stream.h
shubhamdhama: Yeah I need to remove this. Earlier I thought we only need spilled txn for catch-up mode only, but I was wrong.

src/bin/pgcopydb/ld_transform.c
shubhamdhama: Thank you, I think this way I can also eliminate similar problem in `ld_apply.c` and fix a problem that I left as a follow-up inside this PR only.

src/bin/pgcopydb/ld_apply.c
```diff
@@ -495,13 +495,30 @@ stream_apply_sql(StreamApplyContext *context,
...
 			{
 				log_fatal("Failed to parse BEGIN message: %s", sql);
 				return false;
 			}
 
+			/* if txnCommitLSN is be invalid, then fetch it from txn metadata file */
+			if (metadata->txnCommitLSN == InvalidXLogRecPtr)
+			{
+				log_trace("stream_apply_sql: BEGIN message without commit LSN, "
+						  "fetching commit LSN from transaction metadata file");
```
dimitri, line 508: This looks like it should be LOG_DEBUG, as it's not an implementation detail. I try to keep TRACE level for checking internals (algorithms, resource usage, low-level programmer things). Again, there is no hard rule about it, unfortunately. Here the user can follow what's happening. I would go as far as:

  1. write a separate function to compute the filename where to read/parse the XID from
  2. use a log_debug message that includes the name of the file, so that if something goes wrong we can use the logs to open the file and see its contents and if it exists etc
  3. change `parseTxnMetadataFile` to take a `const char *filename, LogicalMessageMetadata *metadata` argument list

src/bin/pgcopydb/ld_apply.c
```diff
@@ -991,3 +1008,85 @@ parseSQLAction(const char *query, LogicalMessageMetadata *metadata)
...
+	/*
+	 * Read the transaction metadata file created by the transform process for
+	 * transactions spanning multiple WAL files. The .json file is generated
+	 * upon encountering the COMMIT statement, but it may take some time to
+	 * become available for transformation. Therefore, we retry indefinitely.
+	 */
+
+	ConnectionRetryPolicy retryPolicy = { 0 };
+
+	int maxT = 1800;            /* 1800s */
```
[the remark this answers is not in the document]
dimitri: I am not a fan of infinite timeouts. Worst case of a failure here is that the follow supervisor is going to switch to the next apply mode (catchup/replay), right? Let's use a not-too-short timeout there, minutes, not hours. We're waiting for a single transaction to be written to disk, and even if the number of messages per transaction is unbounded, that should happen: after all the logical decoding output plugin did manage to stream the whole transaction to our client code.

src/bin/pgcopydb/ld_stream.c
```diff
@@ -313,6 +313,25 @@ stream_init_context(StreamContext *privateContext, StreamSpecs *specs)
 
 	privateContext->lastWrite = 0;
```
dimitri, line 314: I know it's not directly in your PR, but please consider renaming `privateContext->lastWrite` to `privateContext->lastWriteTime` or something, so that we avoid confusion with `maxWrittenLSN`.

src/bin/pgcopydb/ld_stream.c
```diff
@@ -767,8 +792,50 @@ streamRotateFile(LogicalStreamContext *context)
...
+	 * latest file and preventing rotation to earlier files.
+	 *
+	 * In most cases, jsonFileLSN should be the same as cur_record_lsn. However,
+	 * occasionally, current messages may have LSNs lower than the previous
+	 * ones. This can occur due to concurrent transactions with interleaved
+	 * LSNs. Since the logical decoding protocol sends the complete transaction
+	 * at commit time, the LSNs for messages within one transaction could be
+	 * lower than those of the previously streamed transactions. In such cases,
+	 * we use maximum LSN of the messages written so far to the disk in order to
+	 * write to the current file.
```
dimitri, line 810: I think we should copy/paste your ascii art drawing of 3 interleaved transactions here, it would make the whole thing so much easier to grasp when reading this part of the code.

dimitri (changes requested):
Almost there! A couple of refactoring improvements could still be made before merging. Thanks!

src/bin/pgcopydb/ld_stream.c
shubhamdhama: Done

src/bin/pgcopydb/ld_apply.c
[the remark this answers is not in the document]
shubhamdhama: After timing out here we are returning `false` so this will lead to failed process, follow supervisor also consider it as failed and hence the whole migration fails. I've kept it 15 minutes for now, but it needs more thinking here or probably in follow-up PR as an improvement, what do you think?

src/bin/pgcopydb/ld_apply.c
```diff
@@ -991,3 +1021,91 @@ parseSQLAction(const char *query, LogicalMessageMetadata *metadata)
...
+bool
+parseTxnMetadataFile(const char *filename, LogicalMessageMetadata *metadata)
+{
+	/* store xid as it will be overwritten while parsing metadata */
+	uint32_t xid = metadata->xid;
+
+	if (xid == 0)
+	{
+		log_error("BUG: parseTxnMetadataFile is called with "
+				  "0 transaction xid");
```
dimitri, line 1039: I would spell it: “called with transaction xid: %d”, xid. Just to make sure.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1880,3 +1923,69 @@ stream_write_value(FILE *out, LogicalMessageValue *value)
...
+ *  computeTxnMetadataFilename computes the file path for transaction metadata
+ *  based on its transaction id
+ */
+bool
+computeTxnMetadataFilename(uint32_t xid, const char *dir, char *filename)
+{
+	if (dir == NULL)
+	{
+		log_error("BUG: computeTxnMetadataFilename is called with "
+				  "a NULL transaction directory");
```
dimitri, line 1938: Put value at the end of the line please, it's easier to read in the logs.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1880,3 +1923,69 @@ stream_write_value(FILE *out, LogicalMessageValue *value)
...
+		return false;
+	}
+
+	if (xid == 0)
+	{
+		log_error("BUG: computeTxnMetadataFilename is called with "
+				  "0 transaction xid");
```
dimitri, line 1945: Again, value at the end of the line, and let's use %d, xid; even though we know it's zero.

dimitri (changes requested):

dimitri (approved):

PR #311, opened 2023-06-06
Support for JWT tokens as PG passwords

cpj2195:
Made necessary changes to pgcopydb clone(not follow), ping, list, dump, restore and snapshot commands.
Minor refactoring to remove unused functions.

### DIFF
--- a/src/bin/pgcopydb/cli_common.c
+++ b/src/bin/pgcopydb/cli_common.c
@@ -172,8 +172,8 @@ cli_copydb_getenv(CopyDBOptions *options)
 	if (env_exists(PGCOPYDB_SOURCE_PGURI))
 	{
 		if (!get_env_dup(PGCOPYDB_SOURCE_PGURI,
-						  &(options->source_pguri),
-						  NULL))
+						 &(options->source_pguri),
+						 NULL))
 		{
 			/* errors have already been logged */
 			++errors;
@@ -183,8 +183,8 @@ cli_copydb_getenv(CopyDBOptions *options)
 	if (env_exists(PGCOPYDB_TARGET_PGURI))
 	{
 		if (!get_env_dup(PGCOPYDB_TARGET_PGURI,
-						  &(options->target_pguri),
-						  NULL))
+						 &(options->target_pguri),
+						 NULL))
 		{
 			/* errors have already been logged */
 			++errors;
--- a/src/bin/pgcopydb/cli_common.c
+++ b/src/bin/pgcopydb/cli_common.c
@@ -171,9 +171,9 @@ cli_copydb_getenv(CopyDBOptions *options)
 	/* now some of the options can be also set from the environment */
 	if (env_exists(PGCOPYDB_SOURCE_PGURI))
 	{
-		if (!get_env_copy(PGCOPYDB_SOURCE_PGURI,
-						  options->source_pguri,
-						  sizeof(options->source_pguri)))
+		if (!get_env_copy_dynamic(PGCOPYDB_SOURCE_PGURI,
+						  &(options->source_pguri),
+						  NULL))
 		{
 			/* errors have already been logged */
 			++errors;
@@ -182,9 +182,9 @@ cli_copydb_getenv(CopyDBOptions *options)
 
 	if (env_exists(PGCOPYDB_TARGET_PGURI))
 	{
-		if (!get_env_copy(PGCOPYDB_TARGET_PGURI,
-						  options->target_pguri,
-						  sizeof(options->target_pguri)))
+		if (!get_env_copy_dynamic(PGCOPYDB_TARGET_PGURI,
+						  &(options->target_pguri),
+						  NULL))
 		{
 			/* errors have already been logged */
 			++errors;
@@ -717,7 +717,13 @@ cli_copy_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					++errors;
 				}
-				strlcpy(options.source_pguri, optarg, MAXCONNINFO);
+				size_t len = strlen(optarg) + 1;
+				options.source_pguri = (char *) calloc(len, sizeof(char));
+				if (options.source_pguri  == NULL) {
+					log_error(ALLOCATION_FAILED_ERROR);
+					break;
+				}
+				strlcpy(options.source_pguri, optarg, len);
 				log_trace("--source %s", options.source_pguri);
 				break;
 			}
@@ -730,7 +736,13 @@ cli_copy_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					++errors;
 				}
-				strlcpy(options.target_pguri, optarg, MAXCONNINFO);
+				size_t len = strlen(optarg) + 1;
+				options.target_pguri = (char *) calloc(len, sizeof(char));
+				if (options.target_pguri  == NULL) {
+					log_error(ALLOCATION_FAILED_ERROR);
+					break;
+				}
+				strlcpy(options.target_pguri, optarg, len);
 				log_trace("--target %s", options.target_pguri);
 				break;
 			}
--- a/src/bin/pgcopydb/cli_common.c
+++ b/src/bin/pgcopydb/cli_common.c
@@ -171,7 +171,7 @@ cli_copydb_getenv(CopyDBOptions *options)
 	/* now some of the options can be also set from the environment */
 	if (env_exists(PGCOPYDB_SOURCE_PGURI))
 	{
-		if (!get_env_copy_dynamic(PGCOPYDB_SOURCE_PGURI,
+		if (!get_env_dup(PGCOPYDB_SOURCE_PGURI,
 						  &(options->source_pguri),
 						  NULL))
 		{
@@ -182,7 +182,7 @@ cli_copydb_getenv(CopyDBOptions *options)
 
 	if (env_exists(PGCOPYDB_TARGET_PGURI))
 	{
-		if (!get_env_copy_dynamic(PGCOPYDB_TARGET_PGURI,
+		if (!get_env_dup(PGCOPYDB_TARGET_PGURI,
 						  &(options->target_pguri),
 						  NULL))
 		{
@@ -717,13 +717,7 @@ cli_copy_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					++errors;
 				}
-				size_t len = strlen(optarg) + 1;
-				options.source_pguri = (char *) calloc(len, sizeof(char));
-				if (options.source_pguri  == NULL) {
-					log_error(ALLOCATION_FAILED_ERROR);
-					break;
-				}
-				strlcpy(options.source_pguri, optarg, len);
+				options.source_pguri = strdup(optarg);
 				log_trace("--source %s", options.source_pguri);
 				break;
 			}
@@ -736,13 +730,7 @@ cli_copy_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					++errors;
 				}
-				size_t len = strlen(optarg) + 1;
-				options.target_pguri = (char *) calloc(len, sizeof(char));
-				if (options.target_pguri  == NULL) {
-					log_error(ALLOCATION_FAILED_ERROR);
-					break;
-				}
-				strlcpy(options.target_pguri, optarg, len);
+				options.target_pguri = strdup(optarg);
 				log_trace("--target %s", options.target_pguri);
 				break;
 			}
--- a/src/bin/pgcopydb/cli_common.c
+++ b/src/bin/pgcopydb/cli_common.c
@@ -172,8 +172,7 @@ cli_copydb_getenv(CopyDBOptions *options)
 	if (env_exists(PGCOPYDB_SOURCE_PGURI))
 	{
 		if (!get_env_dup(PGCOPYDB_SOURCE_PGURI,
-						 &(options->source_pguri),
-						 NULL))
+						 &(options->source_pguri)))
 		{
 			/* errors have already been logged */
 			++errors;
@@ -183,8 +182,7 @@ cli_copydb_getenv(CopyDBOptions *options)
 	if (env_exists(PGCOPYDB_TARGET_PGURI))
 	{
 		if (!get_env_dup(PGCOPYDB_TARGET_PGURI,
-						 &(options->target_pguri),
-						 NULL))
+						 &(options->target_pguri)))
 		{
 			/* errors have already been logged */
 			++errors;
@@ -717,7 +715,7 @@ cli_copy_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					++errors;
 				}
-				options.source_pguri = strdup(optarg);
+				options.source_pguri = pg_strdup(optarg);
 				log_trace("--source %s", options.source_pguri);
 				break;
 			}
@@ -730,7 +728,7 @@ cli_copy_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					++errors;
 				}
-				options.target_pguri = strdup(optarg);
+				options.target_pguri = pg_strdup(optarg);
 				log_trace("--target %s", options.target_pguri);
 				break;
 			}
--- a/src/bin/pgcopydb/cli_list.c
+++ b/src/bin/pgcopydb/cli_list.c
@@ -228,13 +228,7 @@ cli_list_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					exit(EXIT_CODE_BAD_ARGS);
 				}
-				size_t len = strlen(optarg) + 1;
-				options.source_pguri = (char *) calloc(len, sizeof(char));
-				if (options.source_pguri  == NULL) {
-					log_error(ALLOCATION_FAILED_ERROR);
-					break;
-				}
-				strlcpy(options.source_pguri, optarg, len);
+				options.source_pguri = strdup(optarg);
 				log_trace("--source %s", options.source_pguri);
 				break;
 			}
@@ -1538,13 +1532,7 @@ copydb_init_specs_from_listdboptions(CopyDBOptions *options,
 									 ListDBOptions *listDBoptions)
 {
 	strlcpy(options->dir, listDBoptions->dir, MAXPGPATH);
-	options->source_pguri = calloc(strlen(listDBoptions->source_pguri)+1, sizeof(char));
-	if(options->source_pguri == NULL){
-		log_error(ALLOCATION_FAILED_ERROR);
-		return false;
-	}
-	strlcpy(options->source_pguri, listDBoptions->source_pguri, strlen(listDBoptions->source_pguri)+1);
-
+	options->source_pguri = listDBoptions->source_pguri;
 	options->splitTablesLargerThan = listDBoptions->splitTablesLargerThan;
 
 	strlcpy(options->splitTablesLargerThanPretty,
--- a/src/bin/pgcopydb/cli_list.c
+++ b/src/bin/pgcopydb/cli_list.c
@@ -208,6 +208,12 @@ cli_list_db_getopts(int argc, char **argv)
 	};
 
 	optind = 0;
+	/* read values from the environment */
+	if (!cli_copydb_getenv(&options))
+	{
+		log_fatal("Failed to read default values from the environment");
+		exit(EXIT_CODE_BAD_ARGS);
+	}
 
 	while ((c = getopt_long(argc, argv, "S:T:D:j:s:t:PL:cCJVvdzqh",
 							long_options, &option_index)) != -1)
@@ -222,7 +228,13 @@ cli_list_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					exit(EXIT_CODE_BAD_ARGS);
 				}
-				strlcpy(options.source_pguri, optarg, MAXCONNINFO);
+				size_t len = strlen(optarg) + 1;
+				options.source_pguri = (char *) calloc(len, sizeof(char));
+				if (options.source_pguri  == NULL) {
+					log_error(ALLOCATION_FAILED_ERROR);
+					break;
+				}
+				strlcpy(options.source_pguri, optarg, len);
 				log_trace("--source %s", options.source_pguri);
 				break;
 			}
@@ -403,21 +415,6 @@ cli_list_db_getopts(int argc, char **argv)
 
 			default:
 			{
-				++errors;
-			}
-		}
-	}
-
-	/* list commands support the source URI environment variable */
-	if (IS_EMPTY_STRING_BUFFER(options.source_pguri))
-	{
-		if (env_exists(PGCOPYDB_SOURCE_PGURI))
-		{
-			if (!get_env_copy(PGCOPYDB_SOURCE_PGURI,
-							  options.source_pguri,
-							  sizeof(options.source_pguri)))
-			{
-				/* errors have already been logged */
 				++errors;
 			}
 		}
@@ -1326,6 +1323,12 @@ cli_list_schema(int argc, char **argv)
 		/* errors have already been logged */
 		exit(EXIT_CODE_BAD_ARGS);
 	}
+	/* allocating it to the length of listDBoptions.source_pguri as ListDBOptions does not have target_pguri */
+	options.target_pguri = calloc(strlen(listDBoptions.source_pguri)+1, sizeof(char));
+	if(options.target_pguri == NULL){
+		log_error(ALLOCATION_FAILED_ERROR);
+		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
 
 	if (!copydb_init_specs(&copySpecs, &options, DATA_SECTION_ALL))
 	{
@@ -1535,7 +1538,12 @@ copydb_init_specs_from_listdboptions(CopyDBOptions *options,
 									 ListDBOptions *listDBoptions)
 {
 	strlcpy(options->dir, listDBoptions->dir, MAXPGPATH);
-	strlcpy(options->source_pguri, listDBoptions->source_pguri, MAXCONNINFO);
+	options->source_pguri = calloc(strlen(listDBoptions->source_pguri)+1, sizeof(char));
+	if(options->source_pguri == NULL){
+		log_error(ALLOCATION_FAILED_ERROR);
+		return false;
+	}
+	strlcpy(options->source_pguri, listDBoptions->source_pguri, strlen(listDBoptions->source_pguri)+1);
 
 	options->splitTablesLargerThan = listDBoptions->splitTablesLargerThan;
 
--- a/src/bin/pgcopydb/cli_list.c
+++ b/src/bin/pgcopydb/cli_list.c
@@ -1319,7 +1319,8 @@ cli_list_schema(int argc, char **argv)
 	}
 	/* allocating it to the length of listDBoptions.source_pguri as ListDBOptions does not have target_pguri */
 	options.target_pguri = calloc(strlen(listDBoptions.source_pguri)+1, sizeof(char));
-	if(options.target_pguri == NULL){
+	if(options.target_pguri == NULL)
+	{
 		log_error(ALLOCATION_FAILED_ERROR);
 		exit(EXIT_CODE_INTERNAL_ERROR);
 	}
--- a/src/bin/pgcopydb/cli_list.c
+++ b/src/bin/pgcopydb/cli_list.c
@@ -229,7 +229,7 @@ cli_list_db_getopts(int argc, char **argv)
 							  "see above for details.");
 					exit(EXIT_CODE_BAD_ARGS);
 				}
-				options.source_pguri = strdup(optarg);
+				options.source_pguri = pg_strdup(optarg);
 				log_trace("--source %s", options.source_pguri);
 				break;
 			}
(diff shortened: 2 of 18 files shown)


src/bin/pgcopydb/cli_snapshot.c
```diff
@@ -84,7 +84,13 @@ cli_create_snapshot_getopts(int argc, char **argv)
...
 					exit(EXIT_CODE_BAD_ARGS);
 				}
-				strlcpy(options.source_pguri, optarg, MAXCONNINFO);
+				size_t len = strlen(optarg) + 1;
+				options.source_pguri = (char *) calloc(len, sizeof(char));
+				if (options.source_pguri  == NULL) {
+					log_error(ALLOCATION_FAILED_ERROR);
+					break;
+				}
+				strlcpy(options.source_pguri, optarg, len);
```
dimitri, line 93: strdup

src/bin/pgcopydb/copydb.c
```diff
@@ -633,12 +633,12 @@ copydb_init_specs(CopyDataSpec *specs,
...
 
-		.source_pguri = { 0 },
-		.target_pguri = { 0 },
+		.source_pguri = (char *) calloc(strlen(options->source_pguri) + 1, sizeof(char)),
+		.target_pguri = (char *) calloc(strlen(options->target_pguri) + 1, sizeof(char)),
 
 		.sourceSnapshot = {
 			.pgsql = { 0 },
-			.pguri = { 0 },
+			.pguri = (char *) calloc(strlen(options->source_pguri) + 1, sizeof(char)),
```
dimitri, line 641: Is there a way that we can share the existing memory area and just copy the pointer to it here?

dimitri: Otherwise again, just use strdup.

src/bin/pgcopydb/copydb.c
```diff
@@ -787,8 +791,8 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
 		.cfPaths = &(specs->cfPaths),
 		.pgPaths = &(specs->pgPaths),
 
-		.source_pguri = (char *) &(specs->source_pguri),
-		.target_pguri = (char *) &(specs->target_pguri),
+		.source_pguri = (char *) calloc(strlen(specs->source_pguri) + 1, sizeof(char)),
+		.target_pguri = (char *) calloc(strlen(specs->target_pguri) + 1, sizeof(char)),
```
dimitri, line 795: Again, either keep the same pointer the a memory area that's known to be kept around for the whole duration of the command, or use strdup.

src/bin/pgcopydb/copydb.h
```diff
@@ -76,7 +76,7 @@ typedef struct TransactionSnapshot
 	TransactionSnapshotKind kind;
 	TransactionSnapshotState state;
 
-	char pguri[MAXCONNINFO];
+	char *pguri;
```
dimitri, line 79: Add same-line comment: `/* malloc'ed area */`.

src/bin/pgcopydb/copydb.h
```diff
@@ -216,8 +216,8 @@ typedef struct CopyDataSpec
 	SourceFilterItem *hOid;     /* hash table of objects, by Oid */
 	SourceFilterItem *hName;    /* hash table of objects, by pg_restore name */
 
-	char source_pguri[MAXCONNINFO];
-	char target_pguri[MAXCONNINFO];
+	char *source_pguri;
+	char *target_pguri;
```
dimitri, line 220: `/* malloc'ed area */`

src/bin/pgcopydb/parsing_utils.h
```diff
@@ -81,7 +81,7 @@ typedef struct URIParams
 typedef struct SafeURI
 {
 	char pguri[MAXCONNINFO];
-	char password[MAXCONNINFO];
+	char *password;
```
dimitri, line 84: `/* malloc'ed area */`

src/bin/pgcopydb/parsing_utils.c
```diff
@@ -940,7 +940,7 @@ extract_connection_string_password(const char *pguri, SafeURI *safeURI)
 		{
 			if (option->val != NULL)
 			{
-				strlcpy(safeURI->password, option->val, MAXCONNINFO);
+				strlcpy(safeURI->password, option->val, strlen(pguri)+1);
```
dimitri, line 943: strdup

src/bin/pgcopydb/pgsql.c
```diff
@@ -174,11 +174,17 @@ pgsql_init(PGSQL *pgsql, char *url, ConnectionType connectionType)
...
 		/* size of url has already been validated. */
-		strlcpy(pgsql->connectionString, url, MAXCONNINFO);
+		size_t len = strlen(url) + 1;
+		pgsql->connectionString = (char *) calloc(len, sizeof(char));
+		if (pgsql->connectionString == NULL) {
+			// Log or handle the error when memory allocation fails
+			log_error(ALLOCATION_FAILED_ERROR);
+			return false;
+		}
+		strlcpy(pgsql->connectionString, url, len);
```
dimitri, line 187: Either strdup, or keep the given pointer?

src/bin/pgcopydb/cli_common.c
```diff
@@ -171,9 +171,9 @@ cli_copydb_getenv(CopyDBOptions *options)
 	/* now some of the options can be also set from the environment */
 	if (env_exists(PGCOPYDB_SOURCE_PGURI))
 	{
-		if (!get_env_copy(PGCOPYDB_SOURCE_PGURI,
-						  options->source_pguri,
-						  sizeof(options->source_pguri)))
+		if (!get_env_copy_dynamic(PGCOPYDB_SOURCE_PGURI,
+						  &(options->source_pguri),
+						  NULL))
```
dimitri, line 176: That feels strange. I suppose I would have gone with `get_env_dup` that calls `strdup` instead of `strcpy`, and I would arrange to use the same preliminary code from `get_env_copy_with_fallback`. Also we might have to introduce the variant with and without a fallback value.

dimitri (changes requested):
Thanks for the PR! A lot of work needs to happen still. Please consider following the code style (in particular with brackets on their own lines, also same-line comments for any memory area that is allocated dynamically).

On a higher level, I think in many cases we should just call `strdup` to duplicate the given C string, and in many other cases we should just keep the given pointer to the memory location that is kept around for the whole duration of the program run anyway. This PR is trying to hard...

src/bin/pgcopydb/cli_list.c
```diff
@@ -1326,6 +1317,13 @@ cli_list_schema(int argc, char **argv)
 		/* errors have already been logged */
 		exit(EXIT_CODE_BAD_ARGS);
 	}
+	/* allocating it to the length of listDBoptions.source_pguri as ListDBOptions does not have target_pguri */
+	options.target_pguri = calloc(strlen(listDBoptions.source_pguri)+1, sizeof(char));
+	if(options.target_pguri == NULL)
+	{
+		log_error(ALLOCATION_FAILED_ERROR);
+		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
```
dimitri, line 1326: In the function copydb_init_specs_from_listdboptions() I see that we just keep the pointer around, so I don't think we need to allocate memory here. If we want to duplicate the memory, I would ask to use strdup() anyway, within the function cited above.

src/bin/pgcopydb/copydb.c
```diff
@@ -633,12 +633,12 @@ copydb_init_specs(CopyDataSpec *specs,
 		.cfPaths = specs->cfPaths,
 		.pgPaths = specs->pgPaths,
 
-		.source_pguri = { 0 },
-		.target_pguri = { 0 },
+		.source_pguri = !IS_EMPTY_STRING_BUFFER(options->source_pguri) ? options->source_pguri : NULL,
+		.target_pguri = !IS_EMPTY_STRING_BUFFER(options->target_pguri) ? options->target_pguri : NULL,
```
dimitri, line 637: It seems to me that `options->source_pguri` (and target) are now dynamically allocated strings (via calls to `strdup`) and not string buffers anymore. The IS_EMPTY_STRING_BUFFER call is not necessary here, please just use:

```c
    .source_pguri = options->source_pguri,
    .target_pguri = options->target_pguri,
```

src/bin/pgcopydb/copydb.c
```diff
@@ -633,12 +633,12 @@ copydb_init_specs(CopyDataSpec *specs,
...
 
 		.sourceSnapshot = {
 			.pgsql = { 0 },
-			.pguri = { 0 },
+			.pguri = !IS_EMPTY_STRING_BUFFER(options->source_pguri) ? options->source_pguri : NULL,
```
dimitri, line 641: Same as previous remark, just assign the pointer. It's not a string buffer anymore.

src/bin/pgcopydb/copydb.c
```diff
@@ -679,23 +679,11 @@ copydb_init_specs(CopyDataSpec *specs,
...
+	if(tmpCopySpecs.source_pguri == NULL || tmpCopySpecs.target_pguri == NULL || tmpCopySpecs.sourceSnapshot.pguri == NULL)
 	{
-		strlcpy(tmpCopySpecs.source_pguri,
-				options->source_pguri,
-				MAXCONNINFO);
-
-		strlcpy(tmpCopySpecs.sourceSnapshot.pguri,
-				options->source_pguri,
-				MAXCONNINFO);
+		return false;
```
dimitri, line 684: As we never test the result of `strdup()`, which is NULL when there is not enough memory, then having a test here seems okay. That said, we should `log_error(ALLOCATION_FAILED_ERROR)` before returning false.

Nitpicking: please install the code formatting tools we use, same as Citus, and cut your lines down to 80 chars max. It's old, but the physical reality of concern is that eyes tire more in horizontal movements than in vertical movements. Well this is not a good place for line length arguments, let's just all apply the same limit: 80.

src/bin/pgcopydb/copydb.c
```diff
@@ -803,6 +791,11 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
 		.indexSemaphore = &(specs->indexSemaphore)
 	};
 
+	if(tmpTableSpecs.source_pguri == NULL || tmpTableSpecs.target_pguri == NULL)
+	{
+		return false;
+	}
+
```
dimitri, line 798: Requires a log entry, using the constant ALLOCATION_FAILED_ERROR (so that we don't have to allocate memory for reporting OOM condition).

src/bin/pgcopydb/copydb.h
```diff
@@ -76,7 +76,7 @@ typedef struct TransactionSnapshot
 	TransactionSnapshotKind kind;
 	TransactionSnapshotState state;
 
-	char pguri[MAXCONNINFO];
+	char *pguri;  	/* malloc'ed area */
```
dimitri, line 79: Is it really though? I believe now it just points to a memory area that's allocated/managed elsewhere, right?

src/bin/pgcopydb/copydb.h
```diff
@@ -216,8 +216,8 @@ typedef struct CopyDataSpec
 	SourceFilterItem *hOid;     /* hash table of objects, by Oid */
 	SourceFilterItem *hName;    /* hash table of objects, by pg_restore name */
 
-	char source_pguri[MAXCONNINFO];
-	char target_pguri[MAXCONNINFO];
+	char *source_pguri; 	    /* malloc'ed area */
+	char *target_pguri;         /* malloc'ed area */
```
dimitri, line 220: Again, just pointers now, the memory allocation does not happen in that structure anymore, right?

src/bin/pgcopydb/env_utils.c
```diff
@@ -116,6 +116,44 @@ get_env_copy_with_fallback(const char *name, char *result, int maxLength,
...
 }
 
+/*
+ * get_env_dup copies the environment variable with "name" into
+ * the result buffer using strdup. It returns false when it fails. If the environment
+ * variable is not set the fallback string will be written in the buffer.
+ * Except when fallback is NULL, in that case an error is returned.
+ */
+bool
+get_env_dup(const char *name, char **result, const char *fallback)
```
dimitri, line 126: Can we have the same API as the copy based function, meaning `get_env_dup` and `get_env_dup_with_fallback`?

src/bin/pgcopydb/env_utils.c
```diff
@@ -116,6 +116,44 @@ get_env_copy_with_fallback(const char *name, char *result, int maxLength,
...
+	{
+		envvalue = fallback;
+		if (envvalue == NULL)
+		{
+			log_error("Failed to get value for environment variable '%s', "
+					  "which is unset", name);
+			return false;
+		}
+	}
+	*result = strdup(envvalue);
```
dimitri, line 151: This requires testing the result value of the `strdup` call, which might return NULL when memory allocation failed. See https://man7.org/linux/man-pages/man3/strdup.3.html.

src/bin/pgcopydb/parsing_utils.c
```diff
@@ -940,7 +940,7 @@ extract_connection_string_password(const char *pguri, SafeURI *safeURI)
 		{
 			if (option->val != NULL)
 			{
-				strlcpy(safeURI->password, option->val, MAXCONNINFO);
+				safeURI->password = strdup(option->val);
```
dimitri, line 943: Worth checking that we didn't get a NULL back.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -350,8 +350,14 @@ pg_dump_db(PostgresPaths *pgPaths,
...
-	char PGPASSWORD[MAXCONNINFO] = { 0 };
-
+	size_t len = strlen(pguri) + 1;
+	char *PGPASSWORD = (char *) calloc(len, sizeof(char));
+	safeURI.password = (char *) calloc(len, sizeof(char));
+	if (PGPASSWORD  == NULL || safeURI.password  == NULL)
+	{
+		log_error(ALLOCATION_FAILED_ERROR);
+		return false;
+	}
```
dimitri, line 360: When do you use the memory allocated here? I think you're using `strdup` which does its own memory allocation anyway, right?

src/bin/pgcopydb/pgcmd.c
```diff
@@ -467,7 +473,14 @@ pg_dumpall_roles(PostgresPaths *pgPaths,
...
 	bool pgpassword_found_in_env = env_exists("PGPASSWORD");
-	char PGPASSWORD[MAXCONNINFO] = { 0 };
+	size_t len = strlen(pguri) + 1;
+	char *PGPASSWORD = (char *) calloc(len, sizeof(char));
+	safeURI.password = (char *) calloc(len, sizeof(char));
+	if (PGPASSWORD  == NULL || safeURI.password  == NULL)
+	{
+		log_error(ALLOCATION_FAILED_ERROR);
+		return false;
+	}
```
dimitri, line 483: Again, I don't think those allocated memory areas are ever going to be used?

src/bin/pgcopydb/pgcmd.c
```diff
@@ -757,7 +770,14 @@ pg_restore_db(PostgresPaths *pgPaths,
...
```
dimitri, line 780: Same again.

src/bin/pgcopydb/pgsql.c
```diff
@@ -2024,80 +2023,6 @@ pgsql_handle_notifications(PGSQL *pgsql)
 }
 
 
-/*
- * hostname_from_uri parses a PostgreSQL connection string URI and returns
- * whether the URL was successfully parsed.
- */
-bool
-hostname_from_uri(const char *pguri,
-				  char *hostname, int maxHostLength, int *port)
```
dimitri, line 2033: Thanks for the clean-up, I think we should consider doing that in a separate PR though.

dimitri (changes requested):
Thanks for the new version of the PR. There is still quite some confusion around the memory management used here, and also please consider following the code formatting rules per https://github.com/citusdata/citus/blob/main/CONTRIBUTING.md#following-our-coding-conventions ; which I should probably make more obvious within this repo too.

src/bin/pgcopydb/env_utils.c
cpj2195: Kept just one, because there no `maxLength` that is involved now. So even if its without fallback, one can pass in NULL to `get_env_dup`

src/bin/pgcopydb/pgsql.c
cpj2195: The function wasn't entirely in different context from the PR's(connection strings, URI, passwords etc.), so removed it in this PR itself. Can do a separate cleanup if that's better?

src/bin/pgcopydb/cli_list.c
cpj2195: But pointing `options->target_pguri` to `listDBoptions.source_pguri` or using `strdup` wont be fully correct either, right? Even it has to be done, shouldnt it be done in `cli_list_db_getopts` and not in `copydb_init_specs_from_listdboptions` because that function is only called for cli_list commands which need the source URI to be set. 
Anyway , we still will have allocate this memory , because the NULL checks I added in `copydb.c` would fail.
But, we cant even remove those null checks because `copydb_init_specs` is called at multiple other places which require  the target_uri to not be NULL.

src/bin/pgcopydb/copydb.h
cpj2195: No, actually it is allocated memory in `copydb_copy_snapshot` Line 32

src/bin/pgcopydb/env_utils.c
dimitri: I still think we should offer both APIs, same as the copy stuff. It's easier to understand, remember, and deal with.

src/bin/pgcopydb/pgsql.c
dimitri: Yes let's do a separate PR for cleanup here. Makes it easier to review commits in the git history when we have to dive into advanced issues later.

src/bin/pgcopydb/copydb.c
```diff
@@ -679,23 +679,13 @@ copydb_init_specs(CopyDataSpec *specs,
...
 	{
-		strlcpy(tmpCopySpecs.source_pguri,
-				options->source_pguri,
-				MAXCONNINFO);
-
-		strlcpy(tmpCopySpecs.sourceSnapshot.pguri,
-				options->source_pguri,
-				MAXCONNINFO);
+		log_error(ALLOCATION_FAILED_ERROR);
+		return false;
```
dimitri, line 686: I wonder if testing against NULL for allocation failure is the right thing to do in this part of the code. The allocation failure happened quite a long time ago already. Maybe we could use `pg_strdup` each place you've been calling `strdup` in that PR. The trouble with `pg_strdup` is that it does `exit(1)` when `strdup` fails, but in that context when parsing the command line options that's not a big problem.

src/bin/pgcopydb/copydb.c
```diff
@@ -803,6 +793,12 @@ copydb_init_table_specs(CopyTableDataSpec *tableSpecs,
 		.indexSemaphore = &(specs->indexSemaphore)
 	};
 
+	if (tmpTableSpecs.source_pguri == NULL || tmpTableSpecs.target_pguri == NULL)
+	{
+		log_error(ALLOCATION_FAILED_ERROR);
+		return false;
+	}
+
```
dimitri, line 801: Same, the allocation error was so far back that I would prefer to not have this check here.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -350,8 +350,9 @@ pg_dump_db(PostgresPaths *pgPaths,
 
 	SafeURI safeURI = { 0 };
 	bool pgpassword_found_in_env = env_exists("PGPASSWORD");
-	char PGPASSWORD[MAXCONNINFO] = { 0 };
-
+	size_t len = strlen(pguri) + 1;
+	char *PGPASSWORD = (char *) calloc(len, sizeof(char));
```
dimitri, line 354: The call to `pg_env_dup` is going to take care of memory allocation by calling into `strdup` for us. So that piece of memory is wasted. Just set `char *PGPASSWORD = NULL;`.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -350,8 +350,9 @@ pg_dump_db(PostgresPaths *pgPaths,
...
+	safeURI.password = (char *) calloc(len, sizeof(char));
```
dimitri, line 355: In the function `extract_connection_string_password` the safeURI.password memory allocation is dealt with via calling `strdup` again, so that's another wasted memory area, we can just remove that line entirely.

src/bin/pgcopydb/parsing_utils.h
```diff
@@ -81,7 +81,7 @@ typedef struct URIParams
 typedef struct SafeURI
 {
 	char pguri[MAXCONNINFO];
```
dimitri, line 83: How does that work now that we're not using MAXCONNINFO anywhere else? I suppose that should be another malloc'ed area? It looks to me that `buildPostgresURIfromPieces` should now be using the PQexpBuffer API to build the connection string and then `strdup` the local buffer into the `pguri` slot, which is now a `/* malloc'ed aread */`.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -350,8 +350,9 @@ pg_dump_db(PostgresPaths *pgPaths,
 
 	SafeURI safeURI = { 0 };
```
dimitri, line 351: Now that the connection strings need to be dynamic, the safeURI needs to be freed by the end of the function. We might need a specialized function that takes care of that (two pointers need to be freed in the structure) and we need to call that function before returning (even when we return false).

src/bin/pgcopydb/pgcmd.c
```diff
@@ -467,7 +468,9 @@ pg_dumpall_roles(PostgresPaths *pgPaths,
...
 	bool pgpassword_found_in_env = env_exists("PGPASSWORD");
-	char PGPASSWORD[MAXCONNINFO] = { 0 };
+	size_t len = strlen(pguri) + 1;
+	char *PGPASSWORD = (char *) calloc(len, sizeof(char));
+	safeURI.password = (char *) calloc(len, sizeof(char));
```
dimitri, line 473: Same as before.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -757,7 +760,9 @@ pg_restore_db(PostgresPaths *pgPaths,
...
```
dimitri, line 765: Same as before.

src/bin/pgcopydb/pgsql.c
```diff
@@ -174,11 +174,10 @@ pgsql_init(PGSQL *pgsql, char *url, ConnectionType connectionType)
 
 	/* set our default retry policy for interactive commands */
 	(void) pgsql_set_interactive_retry_policy(&(pgsql->retryPolicy));
-
 	if (validate_connection_string(url))
 	{
 		/* size of url has already been validated. */
```
dimitri, line 179: That comment needs to be removed now, right?

src/bin/pgcopydb/snapshot.c
```diff
@@ -29,8 +29,7 @@ copydb_copy_snapshot(CopyDataSpec *specs, TransactionSnapshot *snapshot)
 
 	/* remember if the replication slot has been created already */
 	snapshot->exportedCreateSlotSnapshot = source->exportedCreateSlotSnapshot;
-
-	strlcpy(snapshot->pguri, source->pguri, sizeof(snapshot->pguri));
+	snapshot->pguri = strdup(source->pguri);
```
dimitri, line 32: Then I suppose we need to call `free(snapshot->pguri)` in the function that closes the snapshot?

src/bin/pgcopydb/snapshot.c
```diff
@@ -230,7 +229,7 @@ copydb_close_snapshot(CopyDataSpec *copySpecs)
 			/* only COMMIT sql snapshot kinds, no need for logical rep ones */
 			if (!pgsql_commit(pgsql))
 			{
-				char pguri[MAXCONNINFO] = { 0 };
+				char *pguri = (char *) calloc(strlen(snapshot->pguri) + 1, sizeof(char));
```
dimitri, line 232: Why is that necessary? The `parse_and_scrub_connection_string` calls into `buildPostgresURIfromPieces` which should be responsible for memory allocation like discussed earlier. Just initialize to NULL here. And make sure to `free` the allocated memory.

dimitri (changes requested):
Still lots of confusion around memory allocation responsibilities, and now that we dynamically allocate memory we also need to think about when to call `free` on these new areas.

src/bin/pgcopydb/parsing_utils.h
cpj2195: It's because , it was the password that was getting the URI past the 1024 bytes limit and SafeURI.pgURI has the password masked so (or removed), so it works fine. I didnt remove this for the very same reason.

src/bin/pgcopydb/snapshot.c
cpj2195: The function docs for `buildPostgresURIfromPieces` says it's the caller's responsibility to allocate pguri to atleast 1024 bytes which is what I did here as well and so at other places where `parse_and_scrub_connection_string` is being called. This is in follow to the conversation above where sageURI.pguri or scrubbedURI are working with just MAXXCONNINFO length because the passwords are masked.

src/bin/pgcopydb/cli_list.c
```diff
@@ -1327,6 +1319,14 @@ cli_list_schema(int argc, char **argv)
...
 	}
 
+	/* allocating it to the length of listDBoptions.source_pguri as ListDBOptions does not have target_pguri */
+	options.target_pguri = calloc(strlen(listDBoptions.source_pguri) + 1, sizeof(char));
+	if (options.target_pguri == NULL)
+	{
+		log_error(ALLOCATION_FAILED_ERROR);
+		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
+
```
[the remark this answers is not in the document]
cpj2195: Then I believe we dont even need this , right? since we are removing the NULL checks in the downstream fucntion calls.

src/bin/pgcopydb/snapshot.c
dimitri: Well the comment there applies to code before your PR, when we used to use static memory allocation (string buffers) for connection strings. It's not the case anymore, so that needs to change. Also we can't approach this PR with having in mind that only the password field is going to be a problem. For instance, we build a new Postgres URI to connect in logical replication mode, and we need our internal facilities to be able to handle that.

src/bin/pgcopydb/parsing_utils.h
dimitri: It's not about just the password. Any other field or combination of fields might push beyond the current limit. The JWT password is a nice/clean/obvious way to reach into the problem, we can't believe it's going to be the only one.

cpj2195: Got it. That kind of settles it that safeURI.pg_uri should be dynamically allocated.
But does it mean that everytime `parse_and_scrub_connection_string` is called, the scrubbedPGURIs should also be dynamically allocated because right now, they arent?( This is at multiple places )
Also, does the same argument apply to:
`char escapedUsername[MAXCONNINFO] = { 0 };`
`char escapedHostname[MAXCONNINFO] = { 0 };`
`char escapedDBName[MAXCONNINFO] = { 0 };`
in the function `buildPostgresURIfromPieces`(lines 670-672).?

@dimitri FYI

dimitri: I think for now the escaped values can be kept as a static buffer, thanks to `escapeWithPercentEncoding()` checking for size overflow. That said, let's use BUFSIZE for their dimension rather than using MAXCONNINFO which we deprecate here.

dimitri: About the main connection string, `buildPostgresURIfromPieces` should be made to allocate dynamic memory rather than fill-in a pre-allocated string buffer ; which means passing `char **pguri` I suppose. Then like I said I believe it should use the PQexpBuffer API internally, and then `strdup` to `*pguri`.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -364,7 +363,7 @@ pg_dump_db(PostgresPaths *pgPaths,
 	if (!IS_EMPTY_STRING_BUFFER(safeURI.password))
 	{
 		if (pgpassword_found_in_env &&
-			!get_env_copy("PGPASSWORD", PGPASSWORD, MAXCONNINFO))
+			!get_env_dup("PGPASSWORD", &(PGPASSWORD)))
```
dimitri, line 366: Nitpicking: no need for parentheses here. Adding them doesn't change anything implementation-wide, but makes the code a little harder to read (for my eyes at least).

src/bin/pgcopydb/pgsql.c
```diff
@@ -174,11 +174,9 @@ pgsql_init(PGSQL *pgsql, char *url, ConnectionType connectionType)
 
 	/* set our default retry policy for interactive commands */
 	(void) pgsql_set_interactive_retry_policy(&(pgsql->retryPolicy));
-
 	if (validate_connection_string(url))
 	{
-		/* size of url has already been validated. */
-		strlcpy(pgsql->connectionString, url, MAXCONNINFO);
+		pgsql->connectionString = url;
```
dimitri, line 179: I think this requires editing the comment of the function to say that the given `char *url` is expected to point to a string that is kept around for at least as long as the PGSQL instance.

dimitri (changes requested):

PR #318, opened 2023-06-12
apply: Remove unnecessary reading for txnCommitLSN after reaching startpos.

shubhamdhama:
Currently, we use txnCommitLSN to determine if we have reached the startpos. If the txnCommitLSN is invalid, we read the transaction metadata file from disk or wait until it exists. However, there is a bug that we continue to check the validity of txnCommitLSN and read metadata file even after reaching startpos. This additional work is unnecessary.

Furthermore, this commit partially address an issue where a transaction is split by KEEPALIVE message during the replay phase. Previously, we would wait for transaction metadata file to be created, which would never happen and eventually lead to migration failure.

### DIFF
--- a/src/bin/pgcopydb/ld_apply.c
+++ b/src/bin/pgcopydb/ld_apply.c
@@ -501,34 +501,6 @@ stream_apply_sql(StreamApplyContext *context,
 				return false;
 			}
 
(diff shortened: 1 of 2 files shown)


dimitri (changes requested):
I sense too many indentation now... time to refactor with another specialized function?

dimitri (approved):

PR #335, opened 2023-06-21
Fix tests to use password auth as md5

cpj2195:


hanefi:
I do not see the value of this PR. Is there any harm in setting `POSTGRES_HOST_AUTH_METHOD=trust` in a test environment?

dimitri:
> I do not see the value of this PR. Is there any harm in setting `POSTGRES_HOST_AUTH_METHOD=trust` in a test environment?

I think the idea was that we used to have a problem with parsing connection strings longer than MAXCONNINFO, a static length provided by libpq, size 1024 bytes. We wanted to make sure that with some long authentication entries we would still be able to parse the connection string and connect. Meanwhile I believe we have fixed that issue, but making sure we're good on that front is a good idea.

Issue #373, opened 2023-07-07
BUG: logical message I received with !isTransaction, Failed to parse JSON message, Failed to transform JSON messages from input stream


plvnv:
pgcopydb --version

```
INFO   Running pgcopydb version 0.12.12.g30ba542.dirty from "/opt/pgpro/std-13/bin/pgcopydb"
pgcopydb version 0.12.12.g30ba542.dirty
compiled with PostgreSQL 13.10 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44), 64-bit
compatible with Postgres 10, 11, 12, 13, 14, and 15
```

Despite https://github.com/dimitri/pgcopydb/pull/372 fix 

Got the same error with

pgcopydb clone  --follow --restart --table-jobs 4 --index-jobs 4 --plugin wal2json --trace 

```
2023-07-07 16:28:15 15085 SQL    pgsql.c:510               Connecting to [source] "postgres://localhost:5432/archive"                                                                                                          
2023-07-07 16:28:15 15085 SQL    pgsql.c:1634              [SOURCE] update pgcopydb.sentinel set replay_lsn = $1 returning startpos, endpos, apply, write_lsn, flush_lsn, replay_lsn;                                          
2023-07-07 16:28:15 15085 SQL    pgsql.c:1638              '2272/70421348'                                                                                                                                                     
2023-07-07 16:28:15 15083 DEBUG  ld_stream.c:691           Received action B for XID 299147205 at LSN 2272/70431690                                                                                                            
2023-07-07 16:28:15 15083 DEBUG  ld_stream.c:691           Received action C for XID 299147205 at LSN 2272/70433570                                                                                                            
2023-07-07 16:28:15 15083 DEBUG  ld_stream.c:1541          Skipping write for action B for XID 299147149 at LSN 2272/6FE891A0: startpos 2272/6FFC5E20 not been reached                                                         
2023-07-07 16:28:15 15083 DEBUG  ld_stream.c:1541          Skipping write for action U for XID 299147149 at LSN 2272/6FE891A0: startpos 2272/6FFC5E20 not been reached                                                         
2023-07-07 16:28:15 15083 DEBUG  ld_stream.c:691           Received action I for XID 299147149 at LSN 2272/704E3698                                                                                                            
[...]

PR #394, opened 2023-07-19
Fix issue with resuming streaming of interleaved transactions.

shubhamdhama:
This resolves a problem encountered during the transition from prefetch to replay mode, where the transform process identifies some incomplete transactions. This issue also arises when we interrupt 'follow' and execute a 'follow --resume'.

The root cause is upon resumption of streaming, we may encounter a transaction where the initial messages have LSNs less than the startpos calculated from the latest json file.

The implemented fix involves comparing the LSNs with 'startpos' only when resuming streaming of partially complete transactions. Once 'startpos' is reached, comparison of LSNs for subsequent messages is discontinued.

Fixes: #373

### DIFF
--- a/src/bin/pgcopydb/ld_stream.c
+++ b/src/bin/pgcopydb/ld_stream.c
@@ -297,7 +297,6 @@ stream_init_context(StreamSpecs *specs)
 	privateContext->endpos = specs->endpos;
 	privateContext->startpos = specs->startpos;
 	privateContext->startposActionFromJSON = specs->startposActionFromJSON;
-	privateContext->startposComputedFromJSON = specs->startposComputedFromJSON;
 
 	privateContext->mode = specs->mode;
 	privateContext->stdIn = specs->stdIn;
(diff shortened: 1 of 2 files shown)


shubhamdhama:
- [ ] Add test

src/bin/pgcopydb/ld_stream.c
```diff
@@ -318,6 +317,29 @@ stream_init_context(StreamSpecs *specs)
...
+		 specs->startposActionFromJSON == STREAM_ACTION_UPDATE ||
+		 specs->startposActionFromJSON == STREAM_ACTION_DELETE ||
+		 specs->startposActionFromJSON == STREAM_ACTION_TRUNCATE))
+	{
+		privateContext->reachedStartPos = false;
+	}
+	else
+	{
+		privateContext->reachedStartPos = true;
+	}
```
shubhamdhama, line 341: It just seems to many lines in `if` statement, is that okay? Should we do direct assignment to variable rather than `if...else`. Second, should I extract a small function for `action` comparision to B/I/U/D/T?

dimitri (approved):

src/bin/pgcopydb/ld_stream.c
dimitri: That's not necessary.

shubhamdhama: After thinking for a while and making the change, I believe the code is more readable as it is now so maybe we can leave it as it is.

PR #397, opened 2023-07-20
Fix sequences dependencies tracking.

### LINKED ISSUE #390: Sequence in public schema cannot be restored creating table in non-public schema
Hello. This might not really be a bug, but I don't know how to proceed.

For our use case we want to make a copied table with a reduced set of rows, then use that table as the source table with pgcopydb. The original table is in the public schema, so a copy of the table is put into `myschema` but it has the same definition.

Sort of like this:

```sql
create schema if not exists myschema;
CREATE TABLE myschema.accounts (LIKE accounts INCLUDING ALL);
INSERT INTO myschema.accounts SELECT * FROM accounts where id = 123;
```

Now the new copied table `myschema.accounts`  primary key `id` column still has a sequence `public.accounts_id_seq` in the public schema.

We'd like to copy the `myschema.accounts` table including the public scoped sequence. It finds the right number of sequences in the initial Fetching stage, but then fails in the pg_restore stage.

Filter file:

```
[include-only-table]
myschema.accounts
```

Failures finding the public schema sequence:

```
01:39:29 16 ERROR  pg_restore: error: could not execute query: ERROR:  relation "public.accounts_id_seq" does not exist
01:39:29 16 ERROR  LINE 2:     id integer DEFAULT nextval('public.accounts_id_seq'::reg...
01:39:29 16 ERROR                                         ^
01:39:29 16 ERROR  Command was: CREATE TABLE myschema.accounts (
01:39:29 16 ERROR      id integer DEFAULT nextval('public.accounts_id_seq'::regclass) NOT NULL,
```

* Ideally we don't have to change the ownership of the sequence since it's still working for the original table. Should we copy the sequence first and configure it the same as a pre-step? That seems like a pain.
* I tried putting `public.accounts_id_sequence` with and without into the `.ini` filter file, but that was just a guess and that doesn't seem to change the behavior.

Any advice on how to copy a schema-qualified table and get a sequence for the table from another schema with pgcopydb? 

Thanks.

dimitri:
A single sequence could be used by more than one table/column thanks to calling the nextval() function in the default value definition of the column. That said, a sequence is OWNED BY a single table.

In this patch we fix the case where the table that owns the sequence is filtered out from the pgcopydb scope while another table that uses the same sequence is to be copied over... with the sequence, then.

A simple way to create the situation is using the following DDL:

    create table app.foo(id bigserial, f1 text);
    create table copy.foo(like app.foo including all);

And then filter-out schema app from pgcopydb scope.

Fixes #390.

### DIFF
--- a/src/bin/pgcopydb/cli_list.c
+++ b/src/bin/pgcopydb/cli_list.c
@@ -1278,21 +1278,26 @@ cli_list_sequences(int argc, char **argv)
 
 	log_info("Fetched information for %d sequences", sequenceArray.count);
 
-	fformat(stdout, "%8s | %20s | %30s | %10s \n",
-			"OID", "Schema Name", "Sequence Name", "attroid");
+	fformat(stdout, "%8s | %20s | %30s | %10s | %10s | %10s \n",
+			"OID", "Schema Name", "Sequence Name",
+			"Owned By", "attrelid", "attroid");
 
-	fformat(stdout, "%8s-+-%20s-+-%30s-+-%10s\n",
+	fformat(stdout, "%8s-+-%20s-+-%30s-+-%10s-+-%10s-+-%10s\n",
 			"--------",
 			"--------------------",
 			"------------------------------",
+			"----------",
+			"----------",
 			"----------");
 
 	for (int i = 0; i < sequenceArray.count; i++)
 	{
-		fformat(stdout, "%8d | %20s | %30s | %10d\n",
+		fformat(stdout, "%8d | %20s | %30s | %10d | %10d | %10d\n",
 				sequenceArray.array[i].oid,
 				sequenceArray.array[i].nspname,
 				sequenceArray.array[i].relname,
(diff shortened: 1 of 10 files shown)


andyatkinson:
Is there a way I can run the newer version via Docker? Looks like the current version is v0.12 https://github.com/dimitri/pgcopydb/pull/352 from June 2023 and there isn't yet a release branch for a new version I'm seeing, although I'm guessing it would be v0.13.

In local development I run Mac OS and in production we run on some Linux container (CentOS I think) so we'd be interested in whatever next version this update lands in for Linux as well.

Thanks!

dimitri:
At the moment we don't have an automated CI recipe to build a docker image at each commit in the main branch, so instead you can build a docker image from the current source by doing `make build`. It expands to the following:

``` 
$ make version
$ docker build -t pgcopydb .
```

So you could also use that to label the image with your own tags etc if needed.

PR #410, opened 2023-07-31
Implement same-table concurrency using Postgres CTID column.

dimitri:
The Postgres system column "ctid" is the physical location of the row version within its table. It is encoded using the tid datatype: `(pagenumber,rownumber)`.

Every single Postgres table has a "ctid" column and it is always possible to split a table's contents by using ctid based ranges. The question of this approach being good at reducing pgcopydb timings remains open: this PR is meant for allowing more experimentations.


marcocitus (commented):
High level note: Not all table access methods naturally support TID Scan (e.g. Citus Columnar does not), in which case this approach is probably much worse than a single-thread full table scans.

Might be best to limit to heap / an allowlist of table access methods.

PR #433, opened 2023-08-28
Fix memory leaks in transform code

JamesGuthrie:


dimitri:
Thanks a lot @JamesGuthrie for your contribution and taking the time to run a test case through valgrind and all. Much appreciated!

PR #457, opened 2023-09-05
Improve connection string TCP keepalive parameters handling.

dimitri:
Allow overriding at the command-line level, also set the parameters in the connection string so that they're effective even when connecting.

### DIFF
--- a/src/bin/pgcopydb/copydb.c
+++ b/src/bin/pgcopydb/copydb.c
@@ -61,6 +61,27 @@ GUC serverSetttings[] = {
 
 
 /*
+ * These parameters are added to the connection strings, unless the user has
+ * added them, allowing user-defined values to be taken into account.
+ */
+KeyVal connStringDefaults = {
+	.count = 4,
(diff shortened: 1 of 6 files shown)


arajkumar:
@dimitri I'm not sure it is related to this PR, but I see quite a bit of WARNINGS like below which started appearing with builds from main.

>07:19:12 321507 WARN   Failed to connect to "postgres://tsdbadmin@xxxxx:5432/tsdb?sslmode=require&keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60",

dimitri:
Hi @arajkumar ; thanks for the feedback. Can you play with the keepalive parameters and see if another set of values have a positive impact on the situation, or maybe assess if we need to only use the `keepalives_idle` parameter in there, or something like this?

We are internally working in a context where those patches seem to be needed, and I didn't anticipate it would be so specific to our context as to disturb other production environments. Sorry about that.

PR #460, opened 2023-09-08
Optimize INSERT statements by translating them to COPY operations

arajkumar:
This commit optimizes INSERT statements by converting them into COPY operations, significantly enhancing insert throughput. The conversion begins when an INSERT statement is encountered and continues to push data to the COPY operation until a new table or other DML operations are encountered.

Performance tests were conducted using a timeseries benchmark suite[1]. Initially, pgcopydb provided approximately 2500 inserts per second. After implementing the COPY optimization, the insertion rate increased to nearly 26000 inserts per second, resulting in nearly a 10x improvement.

**Please note that this optimization is most effective when consecutive INSERT statements target the same table. Performance may not be as significant when INSERT statements are directed to different tables.**

[1] https://github.com/timescale/tsbs

### DIFF
--- a/src/bin/pgcopydb/ld_apply.c
+++ b/src/bin/pgcopydb/ld_apply.c
@@ -729,6 +729,13 @@ stream_apply_sql(StreamApplyContext *context,
 				return true;
 			}
 
+			/* end ongoing COPY operation */
+			if (context->currentCopyHash != 0)
+			{
+				pg_copy_end(pgsql);
+				context->currentCopyHash = 0;
+			}
+
 			/*
 			 * update replication progress with metadata->lsn, that is,
 			 * transaction COMMIT LSN
@@ -845,6 +852,13 @@ stream_apply_sql(StreamApplyContext *context,
 					context->previousLSN < metadata->lsn;
 			}
 
+			/* end ongoing COPY operation */
+			if (context->currentCopyHash != 0)
+			{
+				pg_copy_end(pgsql);
+				context->currentCopyHash = 0;
+			}
+
 			/* in a transaction only the COMMIT LSN is tracked */
 			if (context->transactionInProgress)
 			{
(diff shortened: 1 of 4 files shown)


src/bin/pgcopydb/ld_apply.c
```diff
@@ -986,20 +1042,22 @@ stream_apply_sql(StreamApplyContext *context,
 			}
 
 			uint32_t hash = metadata->hash;
-			PreparedStmt *stmtHashTable = context->preparedStmt;
-			PreparedStmt *stmt = NULL;
+			bool isCopyOperation = context->currentCopyHash == hash &&
+								   context->currentCopyHash != 0;
```
dimitri, line 1046: The context->currentCopyHash currently serves two purposes: is there a copy operation in flight, and which one. I think I would prefer it if we had two separate variables, one for each purpose. Maybe introduce a boolean slot `context->copyInProgress` or something?

dimitri (changes requested):
Thanks for the contribution!

I am wondering if we are covering enough cases in your code at the moment. First, do we switch from INSERT to COPY at the first INSERT? I read a little too fast and didn't get the answer yet.

Then what happens when we have a single transaction with interleaved INSERT  / UPDATE / DELETE statements? Then we would need to finish the current COPY as soon as we see an UPDATE or DELETE statement, right? I think that case is missing at the moment.

Finally I'm surprised at the lack of changes in unit tests result files. If that's indeed the case that they didn't need to change, then I suppose that means we should probably add a new test case that makes sure we use COPY when we should?

src/bin/pgcopydb/ld_apply.c
```diff
@@ -946,6 +995,13 @@ stream_apply_sql(StreamApplyContext *context,
 				return true;
 			}
 
+			/* end ongoing COPY operation */
```
arajkumar, line 998: >Then what happens when we have a single transaction with interleaved INSERT / UPDATE / DELETE statements? Then we would need to finish the current COPY as soon as we see an UPDATE or DELETE statement, right? I think that case is missing at the moment.

This has been handled here.

src/bin/pgcopydb/ld_apply.c
```diff
@@ -938,6 +952,41 @@ stream_apply_sql(StreamApplyContext *context,
...
+			}
+			else if (context->currentCopyHash != 0)
+			{
+				/*
+				 * INSERT for a different table, end ongoing COPY and start
+				 * new one.
+				 */
+				pg_copy_end(pgsql);
+				context->currentCopyHash = 0;
+			}
```
arajkumar, line 973: This takes care of ending the current copy if the new insert targets a different table.

src/bin/pgcopydb/ld_apply.c
arajkumar: That makes sense, let me change the code.

arajkumar:
>Finally I'm surprised at the lack of changes in unit tests result files. If that's indeed the case that they didn't need to change, then I suppose that means we should probably add a new test case that makes sure we use COPY when we should?

I'm thinking of moving the COPY translator into transform code, that would make the testing simple. Also, we can even introduce a switch to let the user decide between COPY or normal INSERT. WDYT?

arajkumar:
> Finally I'm surprised at the lack of changes in unit tests result files. If that's indeed the case that they didn't need to change, then I suppose that means we should probably add a new test case that makes sure we use COPY when we should?

I haven't deep dived into existing tests, but if it already covers INSERTS, then COPY is covered already right?

dimitri:
> > Finally I'm surprised at the lack of changes in unit tests result files. If that's indeed the case that they didn't need to change, then I suppose that means we should probably add a new test case that makes sure we use COPY when we should?
> 
> I'm thinking of moving the COPY translator into transform code, that would make the testing simple. Also, we can even introduce a switch to let the user decide between COPY or normal INSERT. WDYT?

I think having the SQL files show COPY statements would be better indeed. It would “compress” the data a little compared to PREPARE INSERT and EXECUTE version. It would allow for simple reviewing and testing. When doing that, we also need to adjust the parse in the apply module, though.

I don't think we need to expose a user choice between COPY and normal INSERT, but we might want to switch to COPY only when there are more than one INSERT to cover rather than systematically.

arajkumar:
> I don't think we need to expose a user choice between COPY and normal INSERT, but we might want to switch to COPY only when there are more than one INSERT to cover rather than systematically.

Okay, I was looking into ld_transform.c, streamLogicalTransactionAppendStatement. Currently it is missing the logic to coalesce the sequence of INSERT statements targeting the same tables into single statement. I will fix this first, hopefully it would make the COPY translation simple.

arajkumar:
Superseded by https://github.com/dimitri/pgcopydb/pull/465

PR #465, opened 2023-09-13
Implement multi-value inserts

arajkumar:
This commit optimizes multi-value inserts by transferring insert values from a new statement to an existing one if they target the same table. This optimization significantly reduces network latency when multiple consecutive insert statements are executed within a transaction.


arajkumar:
@dimitri , Do you think this PR can go independent of https://github.com/dimitri/pgcopydb/pull/460?

arajkumar:
Insertion throughput between multi value vs serial insert.

image

(Note: This is a highly opinionated benchmark for time series data, which only performs INSERT operations. Results may vary significantly for workloads that involve other interleaved DML operations.)

.gitignore
```diff
@@ -26,6 +26,7 @@ win32ver.rc
 *.exe
 lib*dll.def
 lib*.pc
+.ccls-cache
```
dimitri, line 29: This looks like it's needed for your local tooling rather than the project's tooling. Please use your own local git setup files for this, I don't think this ignore rule belongs to the project's files (see $GIT_DIR/info/exclude).

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1169,6 +1169,64 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
 }
 
 
+/* append the new insert statement if targets same table as previous */
+static bool
+coalesceInsertStatements(LogicalMessageInsert *last,
+						 LogicalMessageInsert *stmt)
+{
```
dimitri, line 1176: The code style we use returns false when we failed to do the operation, rather than when we decided that we didn't have anything to do. To distinguish the two things, usually we end up adding a `bool *` parameter to the function, but in this case here I believe we could also have the function decide to either coalesce to the last statement or append to the list of statements.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1169,6 +1169,64 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+	if (strcmp(last->nspname, stmt->nspname) != 0 ||
+		strcmp(last->relname, stmt->relname) != 0)
+	{
+		return false;
+	}
```
dimitri, line 1181: I think we're missing a check that the list of columns in the INSERT INTO target list is the same for both the statements. We can't add a VALUES entry otherwise.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1169,6 +1169,64 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+	if (lastValuesArray->capacity < (lastValuesArray->count + 1))
+	{
+		/* double the capacity to avoid excessive realloc */
+		lastValuesArray->capacity *= 2;
+		lastValuesArray->array = (LogicalMessageValues *) realloc(lastValuesArray->array,
+																  sizeof(
+																	  LogicalMessageValues)
+																  *
+																  lastValuesArray->
+																  capacity);
```
dimitri, line 1194: What if realloc fails? Please check for a NULL pointer being returned. BTW that's when we return false.

Nitpicking: review the formatting, it's too far on the right here.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1169,6 +1169,64 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+	lastValuesArray->array[lastValuesArray->count++] = stmt->new.array->values.array[0];
+
+	/* avoid deallocation of moved value */
+	stmt->new.array->values.count = 0;
+
+	return true;
+}
+
+
+/* Append the new statement if the action is same and targets same table. */
```
dimitri, line 1207: Formatting: see other function level comments (all are multi-line).
Semantics: needs to also be the same target list.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1213,16 +1268,28 @@ streamLogicalTransactionAppendStatement(LogicalTransaction *txn,
...
 
-		/* the new statement now becomes the last entry of the linked-list */
-		stmt->prev = txn->last;
-		stmt->next = NULL;
-		txn->last = stmt;
+			/* the new statement now becomes the last entry of the linked-list */
+			stmt->prev = txn->last;
+			stmt->next = NULL;
+			txn->last = stmt;
+		}
```
dimitri, line 1292: I wonder if we should have `coalesceStatements` return a boolean that indicates if it could coalesce, have another function that only check if coalescing makes sense, or have the pointer management done within `coalesceStatements`.

At the moment I think that when adding the code to check that both statement (previous/next) share the same column target list we will prefer having a separate function to check that.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1169,6 +1169,64 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
 }
 
 
+/* append the new insert statement if targets same table as previous */
+static bool
+coalesceInsertStatements(LogicalMessageInsert *last,
+						 LogicalMessageInsert *stmt)
+{
+	if (strcmp(last->nspname, stmt->nspname) != 0 ||
+		strcmp(last->relname, stmt->relname) != 0)
```
dimitri, line 1178: Please use `streq()`; a macro made available everywhere in the pgcopydb code.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1169,6 +1169,64 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+	{
+		return false;
+	}
+
+	LogicalMessageValuesArray *lastValuesArray = &last->new.array->values;
+
+	if (lastValuesArray->capacity < (lastValuesArray->count + 1))
+	{
+		/* double the capacity to avoid excessive realloc */
+		lastValuesArray->capacity *= 2;
```
dimitri, line 1188: We should also check that we are not pushing the number of EXECUTE arguments above the limit. I seem to remember that you can have as many as 16384 arguments (int4).

dimitri (changes requested):
Thanks for that work, that's awesome!

As you have seen I anticipated that we would need to be smarter around the INSERT traffic and using multiple-VALUES statements, and I'm very glad that you're contributing the feature.

Now the current PR needs more work before it's ready, though it's a very solid first step. Can't wait to merge the capability!

dimitri:
> @dimitri , Do you think this PR can go independent of #460?

Yes indeed. Multiple VALUES inserts is a great step already, and then we will be able to compare using COPY to using big INSERT statements. I expect that COPY is going to win, and then we can implement COPY support in the transform writer (and the apply SQL parser); at least we will already have the in-memory representation correct and ready for reuse.

src/bin/pgcopydb/ld_transform.c
arajkumar: Looks like type of `nParams` is `int`. Do you mean the limit is imposed in the server side?
```
PGresult *PQexecPrepared(PGconn *conn,
                         const char *stmtName,
                         int nParams,
                         const char * const *paramValues,
                         const int *paramLengths,
                         const int *paramFormats,
                         int resultFormat);
```

dimitri: Have a look at the protocol definition for the Parse message at https://www.postgresql.org/docs/16/protocol-message-formats.html: 

> **Int16** The number of parameter data types specified (can be zero). Note that this is not an indication of the number of parameters that might appear in the query string, only the number that the frontend wants to prespecify types for.

The number of parameters must fit in an Int16 in the protocol, hence the limit I was sharing.

src/bin/pgcopydb/ld_transform.c
arajkumar: I'm considering another approach: Call coalesce only for the DML case here in https://github.com/dimitri/pgcopydb/blob/a15881860dc18ad268c7bd8cae11dbe0055538dd/src/bin/pgcopydb/ld_transform.c#L1162

something like below,
```c
			if (canCoalesceLogicalTrasactionStatement(txn, stmt))
			{
				if (!coalesceLogicalTrasactionStatement(txn, stmt))
				{
					return false;
				}
			}
			else
			{
				(void) streamLogicalTransactionAppendStatement(txn, stmt);
			}
```

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1159,7 +1163,18 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+				if (!coalesceLogicalTransactionStatement(txn, stmt))
+				{
+					/* errors have already been logged */
+					return false;
+				}
+			}
+			else
+			{
+				(void) streamLogicalTransactionAppendStatement(txn, stmt);
+			}
```
dimitri, line 1177: It looks to me like this code should be found within the function `streamLogicalTransactionAppendStatement`.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1170,20 +1185,137 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+{
+	/* LogicalTransaction is empty when last is NULL */
+	if (txn == NULL || txn->last == NULL || new == NULL)
+	{
+		return false;
+	}
+
+	LogicalTransactionStatement *last = txn->last;
+
+	/* We support coalesce only for INSERTS */
```
dimitri, line 1260: We could support it also for UPDATE and DELETE, it's just that we don't at the moment. 

For UPDATE statements, we should start with multiple statement sharing the same SET clause (columns *and* values) and with only the WHERE clause being different, then we can have an UPDATE FROM statement (UPDATE with a join).

For DELETE statement, same idea, we could delete more than one statement at a time from the same table.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1170,20 +1185,137 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+
+		lastValuesArray->capacity = capacity;
+		lastValuesArray->array = array;
+	}
+
+	/* Move the new value from the 'newValuesArray' to the 'lastValuesArray' of
+	 * the existing statement. Additionally, set the count of the 'newValuesArray'
+	 * to 0 to prevent it from being deallocated by FreeLogicalMessageTupleArray,
+	 * as it has been moved to the 'lastValuesArray'.
+	 */
```
dimitri, line 1228: Nitpicking: It would be nice to better explain the “it has been moved”, in other words recap what `realloc` is doing for us in the memory allocation areas here. Also please have the first line of the comment empty.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1170,20 +1185,137 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+ * array created during the last INSERT statement in a logical transaction.
  *
- *  2. it's the same statement as the previous one, we only add an entry to the
- *     already existing tuple array created on the previous statement
+ * This functionality enables the generation of multi-values INSERT or COPY
+ * commands, enhancing efficiency.
  *
- * This allows to then generate multi-values insert commands, for instance.
+ * Important: Before invoking this function, ensure that validation is performed
+ * using canCoalesceLogicalTransactionStatement.
```
dimitri, line 1195: Makes me wonder if we should inline the whole thing within `streamLogicalTransactionAppendStatement` somehow. It would need to make sense that we do that for UDPATE and DELETE statements too, then, maybe, so that might be for another PR.

dimitri (changes requested):

src/bin/pgcopydb/ld_transform.c
arajkumar: >It would need to make sense that we do that for UDPATE and DELETE statements too, then, maybe, so that might be for another PR.

yeah, it was my thought process too.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1170,20 +1185,137 @@ parseMessage(StreamContext *privateContext, char *message, JSON_Value *json)
...
+	 * Check if adding the new statement would exceed libpq's limit on the total
+	 * number of parameters allowed in a single PQsendPrepare call.
+	 * If it would exceed the limit, return false to indicate that coalescing
+	 * should not be performed.
+	 *
+	 * TODO: This parameter limit check is not applicable for COPY operations.
+	 * It should be removed once we switch to using COPY.
+	 */
+	if (((lastValuesArray->count + 1) * lastInsertColumns->cols) >
+		PQ_QUERY_PARAM_MAX_LIMIT)
```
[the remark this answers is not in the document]
dimitri: Sounds good. Maybe better still in `pg_utils.c`?

src/bin/pgcopydb/ld_transform.c
arajkumar: Sure, I can do it in a separate PR.

dimitri: Well if you prefer to do it here, that's fine too. There is value in finishing up the current PR though ;-)

dimitri (approved):
LGTM

arajkumar:
Thanks @dimitri .

dimitri:
Well thank you @arajkumar ; your PR is a very fine contribution! I hope you'll be able to continue on that path with the other items we mentioned (UPDATE/INSERT multi-rows support, COPY support).

arajkumar:
>  I hope you'll be able to continue on that path with the other items we mentioned (UPDATE/INSERT multi-rows support, COPY support).
I'm on it :) I'm planning to resume the work on supporting the pipeline next. Hope that is brings additional value.

arajkumar:
@dimitri, what is the typical release schedule for pgcopydb? We're writing a blog post about "low downtime migration from Postgres to Timescale" using pgcopydb. It would be super helpful to have a new version with this change!

dimitri:
Hi @arajkumar ; I try to target a monthly release cycle but got distracted the last two months. I should prepare a Milestone on GitHub to make it visible what I'm considering for the next release. I wish we could fix our current situation in logical decoding replay and introduce that bug fix in the next release. Might not be realistic though.

PR #475, opened 2023-09-22
fix drop cascade issue

### LINKED ISSUE #474: ERROR:  cannot drop desired object(s) because other objects depend on them
Built `pgcopydb` from `main` branch at 2023-09-21.
When starting a data migration process, error occur:

```
2023-09-22 07:59:42 3566 SQL    pgsql.c:1493              [TARGET 9899] DROP TABLE IF EXISTS  s.t1, s.t2,<here goes the long list of tables prefixed with schema>, s.tNCASCADE;
2023-09-22 07:59:42 3566 WARN   pgsql.c:847               NOTICE:  table "s.tNCASCADE" does not exist, skipping
2023-09-22 07:59:42 3566 DEBUG  pgsql.c:2033              Postgres result status is PGRES_FATAL_ERROR
2023-09-22 07:59:42 3566 ERROR  pgsql.c:1920              [TARGET 9899] ERROR:  cannot drop desired object(s) because other objects depend on them
```

Notice the double space after `EXISTS` clause and no space before `CASCADE` clause.
It seems that one space should be before `CASCADE` and one space after `EXISTS`.
Could it be the error cause?

The suspicious place is `dump_restore.c`, function `copydb_target_drop_tables`, the loop between `appendPQExpBuffer(query, "DROP TABLE IF EXISTS");` and `appendPQExpBuffer(query, "CASCADE");`

lospejos:
Possibly fixes #474

### DIFF
--- a/src/bin/pgcopydb/dump_restore.c
+++ b/src/bin/pgcopydb/dump_restore.c
@@ -219,7 +219,7 @@ copydb_target_drop_tables(CopyDataSpec *specs)
 						  source->relname);
 	}
 
-	appendPQExpBuffer(query, "CASCADE");
+	appendPQExpBuffer(query, " CASCADE");
 
 	/* memory allocation could have failed while building string */
 	if (PQExpBufferBroken(query))


lospejos:
I've tested this fix on my database and the error mentioned in issue #474 gone. Migrated data successfully.

dimitri:
Thanks a lot for your contribution @lospejos !

PR #479, opened 2023-09-26
Switch our call to PQgetCopyData to async in pg_copy().

dimitri:
This call should not be blocking in our implementation, because we want to be able to react to network problems or server-side interruptions.


shubhamdhama (approved):
@dimitri this worked fine when ran with `--fail-fast`.

PR #482, opened 2023-09-27
Fix identifier escaping issue with test_decoding plugin

arajkumar:
The test_decoding plugin escapes identifiers by using double quotes for column names. However, the pgcopydb transform process also handles escaping by using double quotes, resulting in double escaping and apply process failures.

This commit resolves the issue by removing quotes from column names if present in test_decoding output and allowing the transform process to handle escaping.


src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -715,6 +715,17 @@ parseNextColumn(TestDecodingColumns *cols,
...
+	 * The test_decoding plugin already quotes keywords. We remove the quotes
+	 * here because we will add them later during the transformation to SQL.
+	 */
+	if (cols->colnameStart[0] == '"')
+	{
+		cols->colnameStart++;
+		cols->colnameLen -= 2;
+	}
+
+
```
dimitri, line 728: This looks strange to me. I understand it's the smallest amount of code changes required to pass the tests, but we're now removing quotes only to add them again later. What about instead we only decide to add quotes in `stream_write_...` functions when it's not already quoted?

dimitri (changes requested):

src/bin/pgcopydb/ld_test_decoding.c
arajkumar: That's how I started with :) But yeah, I need to handle this in `stream_write_{insert, delete, update},` functions. I could do that.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1928,7 +1928,9 @@ stream_write_insert(FILE *out, LogicalMessageInsert *insert)
 
 		for (int c = 0; c < stmt->cols; c++)
 		{
-			appendPQExpBuffer(buf, "%s\"%s\"",
+			const char *quoteFormatStr = (stmt->columns[c][0] == '"') ? "%s%s" :
+										 "%s\"%s\"";
+			appendPQExpBuffer(buf, quoteFormatStr,
```
dimitri, line 1933: Thanks for the changes! I think it's worth adding a comment about why this is happening...

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2101,7 +2103,9 @@ stream_write_update(FILE *out, LogicalMessageUpdate *update)
 
 				if (!skip)
 				{
-					appendPQExpBuffer(buf, "%s\"%s\" = $%d",
+					const char *quoteFormatStr = (colname[0] == '"') ? "%s%s = $%d" :
+												 "%s\"%s\" = $%d";
+					appendPQExpBuffer(buf, quoteFormatStr,
```
dimitri, line 2108: Same comment.

dimitri (changes requested):
Thanks for the changes! Let's comment on those quoteFormatStr business to explain why we're doing that?

dimitri (approved):

dimitri:
Thanks a lot for your contribution @arajkumar ; much appreciated!

PR #497, opened 2023-10-15
Add check constraint, language entry mapping

VaibhaveS:
- Add archive entry mappings for "CHECK CONSTRAINT", "LANGUAGE"
- Add unit test for CHECK CONSTRAINT
- Fix false positive typo


src/bin/pgcopydb/pgcmd.c
```diff
@@ -1022,6 +1023,7 @@ struct ArchiveItemDescMapping pgRestoreDescriptionArray[] = {
 	INSERT_MAPPING(ARCHIVE_TAG_FUNCTION, "FUNCTION"),
 	INSERT_MAPPING(ARCHIVE_TAG_INDEX_ATTACH, "INDEX ATTACH"),
 	INSERT_MAPPING(ARCHIVE_TAG_INDEX, "INDEX"),
+	INSERT_MAPPING(ARCHIVE_TAG_LANGUAGE, "LANGUAGE"),
```
shubhamdhama, line 1026: Can we add a test for this too?

src/bin/pgcopydb/cli_stream.c
```diff
@@ -609,7 +609,7 @@ cli_stream_cleanup(int argc, char **argv)
 	bool service = false;
 
 	if (!copydb_init_workdir(&copySpecs,
-							 NULL,
+							 streamDBoptions.dir,
```
shubhamdhama, line 612: We can delete this commit https://github.com/dimitri/pgcopydb/pull/497/commits/ce02dc435e199d4ed7975cbcfbfc0f9312ba1678 from this PR.

src/bin/pgcopydb/pgcmd.c
VaibhaveS: We can have ACL/COMMENT statements for multiple objects like LANGUAGE/SCHEMA/INDEX/COLLATION etc. I think we should add a test case covering all of these.

VaibhaveS: Will create a issue for this, let's take care of this later if need be.

src/bin/pgcopydb/cli_stream.c
dimitri: I believe that because I'm going “squash & merge” anyway, that won't be visible once merged, in a way that it's not important to clean-up the history of the PR after review. It might still be useful at times during review, of course, this PR does not look complex enough (yet?) to warrant history cleanup in my eyes.

PR #503, opened 2023-10-18
Add byte level details to list progress command

hanefi:
This patch allows reporting the total number of bytes transferred during
a copy operation. However, it has a limitation that the reported value
is only updated after each table is copied. This is because we do not
update table summary files while the copy is in progress.

Changes include:
- 3 new fields in table summary structures:
  - network.bytes: total number of bytes transmitted
  - network.bytes-pretty: pretty printed form of network.bytes
  - network.transmit-rate: pretty printed bytes transmitted per second
- Some new fields on the `pgcopydb list progress --json --summary`
  command:
  - steps[].network is a new field that exists for COPY step and has
    the following fields:
	- bytes: total number of bytes transmitted for all tables
	- bytes-pretty: pretty printed form of bytes
  - tables[].network is a new json value with the following fields:
	- bytes: total number of bytes transmitted for table
	- bytes-pretty: pretty printed form of bytes
	- transmit-rate: pretty printed bytes transmitted per second
- A new column on top level summary that shows total number of bytes
  copied.

### DIFF
--- a/docs/ref/pgcopydb_clone.rst
+++ b/docs/ref/pgcopydb_clone.rst
@@ -891,15 +891,16 @@ Examples
    13:09:09 81991 INFO  STEP 7: restore the post-data section to the target database
    13:09:09 81991 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'postgres://@:/plop?' --single-transaction --clean --if-exists --use-list /var/folders/d7/zzxmgs9s16gdxxcm0hs0sssw0000gn/T//pgcopydb/schema/post.list /var/folders/d7/zzxmgs9s16gdxxcm0hs0sssw0000gn/T//pgcopydb/schema/post.dump
 
-                                             Step   Connection    Duration   Concurrency
-    ---------------------------------------------   ----------  ----------  ------------
-                                      Dump Schema       source       355ms             1
-                                   Prepare Schema       target       135ms             1
-    COPY, INDEX, CONSTRAINTS, VACUUM (wall clock)         both       641ms        8 + 12
-                                COPY (cumulative)         both       1s598             8
-                       Large Objects (cumulative)         both        29ms             1
-           CREATE INDEX, CONSTRAINTS (cumulative)       target       4s072            12
-                                  Finalize Schema       target       366ms             1
-    ---------------------------------------------   ----------  ----------  ------------
-                        Total Wall Clock Duration         both       1s499        8 + 12
-    ---------------------------------------------   ----------  ----------  ------------
+                                               Step   Connection    Duration    Transfer   Concurrency
+ --------------------------------------------------   ----------  ----------  ----------  ------------
+                                        Dump Schema       source       225ms                         1
+   Catalog Queries (table ordering, filtering, etc)       source       842ms                         1
+                                     Prepare Schema       target       694ms                         1
+      COPY, INDEX, CONSTRAINTS, VACUUM (wall clock)         both       1s538                    8 + 12
+                                  COPY (cumulative)         both       2s771     2955 kB             8
+                         Large Objects (cumulative)         both       242ms                         1
+             CREATE INDEX, CONSTRAINTS (cumulative)       target       3s489                        12
+                                    Finalize Schema       target       654ms                         1
+ --------------------------------------------------   ----------  ----------  ----------  ------------
+                          Total Wall Clock Duration         both       3s984                    8 + 12
+ --------------------------------------------------   ----------  ----------  ----------  ------------
--- a/docs/ref/pgcopydb_clone.rst
+++ b/docs/ref/pgcopydb_clone.rst
@@ -842,65 +842,87 @@ Examples
 --------
 
 ::
-
-   $ export PGCOPYDB_SOURCE_PGURI="port=54311 host=localhost dbname=pgloader"
-   $ export PGCOPYDB_TARGET_PGURI="port=54311 dbname=plop"
+   $ export PGCOPYDB_SOURCE_PGURI=postgres://pagila:0wn3d@source/pagila
+   $ export PGCOPYDB_TARGET_PGURI=postgres://pagila:0wn3d@target/pagila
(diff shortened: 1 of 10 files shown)


src/bin/pgcopydb/summary.c
```diff
@@ -34,13 +34,23 @@ write_table_summary(CopyTableSummary *summary, char *filename)
 	JSON_Value *js = json_value_init_object();
 	JSON_Object *jsObj = json_value_get_object(js);
 
+	char bytesPretty[BUFSIZE] = { 0 };
+	pretty_print_bytes(bytesPretty, BUFSIZE, summary->bytesTransmitted);
+
+	char transmitRate[BUFSIZE] = { 0 };
+	pretty_print_bytes_per_second(transmitRate, BUFSIZE, summary->bytesTransmitted,
+								  summary->durationMs);
+
```
hanefi, line 43: I did not want to store bytesPretty and transmitRate in CopyTableSummary structs.

1. It is inexpensive to have an integer addition in our copy loop, unlike pretty printing said value. I did not want to store bytes and bytesPretty when I can only update the primitive type and not keep the other in sync.
2. Calculating the transmitRate in real time is also an expensive operation similar to bytesPretty that was discussed in earlier item. Also, I needed to have the duration information that is not always there when the command is running.
3. If resuming an earlier operation, reading an integer value for bytes from disk makes sense. All the other values can be generated from bytes and duration information.

src/bin/pgcopydb/summary.c
```diff
@@ -750,64 +769,71 @@ void
 print_toplevel_summary(Summary *summary)
 {
 	char *d10s = "----------";
+	char *s10s = "          ";
```
hanefi, line 772: I assumed `d10s` stands for `dashes 10 string` and created `s10s` that should stand for `space 10 string`. If this naming does not make sense to other readers, feel free to suggest an alternative name.

I might as well use an empty string, as the `%10s` format would print required number of space characters. I think that seeing those spaces on the screen, near the beginning of the function helps readability.

docs/ref/pgcopydb_clone.rst
```diff
@@ -891,15 +891,16 @@ Examples
...
+   Catalog Queries (table ordering, filtering, etc)       source       842ms                         1
+                                     Prepare Schema       target       694ms                         1
+      COPY, INDEX, CONSTRAINTS, VACUUM (wall clock)         both       1s538                    8 + 12
+                                  COPY (cumulative)         both       2s771     2955 kB             8
+                         Large Objects (cumulative)         both       242ms                         1
+             CREATE INDEX, CONSTRAINTS (cumulative)       target       3s489                        12
+                                    Finalize Schema       target       654ms                         1
+ --------------------------------------------------   ----------  ----------  ----------  ------------
+                          Total Wall Clock Duration         both       3s984                    8 + 12
+ --------------------------------------------------   ----------  ----------  ----------  ------------
```
hanefi, line 906: Note: In this patch, most of the data in the top-level duration tables are made up.

dimitri: All the numbers in the docs to this point where obtained by actual runs and copy/pasted. When that make sense, runs from the test suite, ensuring they're easy to replicate, and that we're in the right ballpark. Please consider the same approach here.

src/bin/pgcopydb/string_utils.c
```diff
@@ -683,10 +683,34 @@ pretty_print_bytes(char *buffer, size_t size, uint64_t bytes)
...
+	else
+	{
+		char bytesPretty[BUFSIZE] = { 0 };
+		uint64_t bytesPerSecond = bytes * 1000 / durationMs;
+
+		pretty_print_bytes(bytesPretty, size-2, bytesPerSecond);
+		sformat(buffer, size, "%sPS", bytesPretty);
+	}
+}
+
```
dimitri, line 709: Please make sure to use official suffix for the unit, as in https://en.wikipedia.org/wiki/Data-rate_units in this case. Which I failed doing in `pretty_print_bytes` now that I read https://en.wikipedia.org/wiki/Byte again ; so that'd be for another patch then.

dimitri (changes requested):
Almost ready for merge, thanks! Still some areas to review and change, see below for details.

src/bin/pgcopydb/string_utils.c
hanefi: I read through the wiki pages and come up with a decision. I believe it is best to mimic what `pg_size_pretty` function in pgcopydb. They use powers of 2, but do not use EIC standard. That means 1024 bytes = 1 kilobytes (kB). I like to keep a similar format and thus keep `pretty_print_bytes` as it is.

However, I failed to find anywhere in the PG codebase where a Data-rate unit is calculated and printed. I think that it makes more sense to use powers of 2. Do you think it makes sense to use EIC standard for Data-rate units? I mean 1024 bit/s (bit per second) = 1 Kibit/s (Kibibit per second)?

IMO, the problem is with how PG pretty prints sizes. They used a questionable format that causes more ambiguity and confusion when we start to report transmit rates. I think it has more significance to have a consistent format than the correct one. PG already messed up, and we need to keep things consistent with their format.

dimitri: Ok, I agree with you on using the Postgres way for pg_size_pretty and pretty_print_bytes, because internally we use both in different places and being able to compare the sizes is good.

For the rates, we can decide whether we want to use multiples of 1000 or 1024 (powers of 2 or powers of 10), but then I will insist on using the proper suffix for it. I am voting for using `kbit/s`, `Mbit/s`, and al, which means powers of 10.

docs/ref/pgcopydb_clone.rst
hanefi: I created a local prototype that can produce reproducible documentation examples. I used a copy of pagila tests as a starting point and I copied commands from our current documentation. Finally I copy pasted command outputs to the docs.

Hopefully, I will open a separate PR with my changes for reproducible documentation example generation after polishing it some more.

dimitri (approved):

PR #505, opened 2023-10-20
Buffer transaction statements to disk to prevent replay deadlock

arajkumar:
This commit buffers transactions being replayed into a separate file until
the entire transaction body is received. It addresses potential deadlocks
that can occur when the BEGIN message arrives before the creation of the
transaction metadata file by the transform process.

### DIFF
--- a/src/bin/pgcopydb/copydb.c
+++ b/src/bin/pgcopydb/copydb.c
@@ -572,6 +572,10 @@ copydb_prepare_filepaths(CopyFilePaths *cfPaths,
 			"%s/lsn.json",
 			cfPaths->cdc.dir);
 
+	sformat(cfPaths->cdc.txnlatestfile, MAXPGPATH,
+			"%s/txn.latest.sql",
+			cfPaths->cdc.dir);
+
 	/*
 	 * Now prepare the "compare" files we need to compare schema and data
 	 * between the source and target instance.
--- a/src/bin/pgcopydb/ld_replay.c
+++ b/src/bin/pgcopydb/ld_replay.c
@@ -27,8 +27,17 @@
 typedef struct ReplayStreamCtx
 {
 	StreamApplyContext applyContext;
+
+	bool isTxnBuffering;
+	bool skipTxnBuffering;
+
+	FILE *txnBuffer;
+
+	char currentTxnFileName[MAXPGPATH];
 } ReplayStreamCtx;
 
+static bool streamReplayLineIntoBuffer(void *ctx, const char *line, bool *stop);
+static bool streamBufferedXID(ReplayStreamCtx *replayCtx);
 
 /*
  * stream_apply_replay implements "live replay" of the changes from the source
@@ -84,7 +93,7 @@ stream_apply_replay(StreamSpecs *specs)
 	 * Setup our PIPE reading callback function and read from the PIPE.
 	 */
 	ReadFromStreamContext readerContext = {
-		.callback = stream_replay_line,
+		.callback = streamReplayLineIntoBuffer,
 		.ctx = &ctx
 	};
 
@@ -174,7 +183,6 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
 	StreamApplyContext *context = &(replayCtx->applyContext);
 
 	LogicalMessageMetadata metadata = { 0 };
-
 	if (!parseSQLAction((char *) line, &metadata))
 	{
 		/* errors have already been logged */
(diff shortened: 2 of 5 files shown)


arajkumar:
@dimitri The latest changes uses spills to disk approach to buffer transactions in replay mode.

dimitri:
Could we store the current transaction to a file named `${xid}.sql` ; and maybe add a static filename as you did but this time as a symlink to the current transaction file? I believe that would make things more obvious when debugging live situations.

arajkumar:
> Could we store the current transaction to a file named `${xid}.sql` ; and maybe add a static filename as you did but this time as a symlink to the current transaction file? I believe that would make things more obvious when debugging live situations.

@dimitri, I've considered a different approach where we buffer transactions at the receive/prefetch level. This would remove the need for a separate metadata file for ongoing transactions, as we can include the commit_lsn in the BEGIN message while buffering the entire transaction.

arajkumar:
@dimitri I've added a symlink to the latest transaction buffer as suggested by you. Additionally KEEPALIVE, SWITCHWAL and ENDPOS messages which comes out of transaction will be replayed without buffering those.

src/bin/pgcopydb/copydb.c
```diff
@@ -572,6 +572,10 @@ copydb_prepare_filepaths(CopyFilePaths *cfPaths,
 			"%s/lsn.json",
 			cfPaths->cdc.dir);
 
+	sformat(cfPaths->cdc.latesttxnfile, MAXPGPATH,
+			"%s/latest_txn",
+			cfPaths->cdc.dir);
```
dimitri, line 577: As this is a SQL file, maybe `txn.latest.sql` would be a better name?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -27,8 +27,10 @@
 typedef struct ReplayStreamCtx
 {
 	StreamApplyContext applyContext;
+	FILE *txnReplayBuffer;
 } ReplayStreamCtx;
```
dimitri, line 31: I think we should expose the current XID file name in that context structure, so that we can use the actual file name latter when applying from file. Using the symlink at apply-from-buffer time seems too “risky” to me, for no benefits.

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +250,136 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+
+	if (strstr(line, OUTPUT_BEGIN) == line)
+	{
+		/* If there is already a replay buffer open, it indicates a partial
+		 * transaction. Close it and start a new one.
+		 */
+		if (replayCtx->txnReplayBuffer != NULL)
+		{
+			fclose(replayCtx->txnReplayBuffer);
+		}
```
dimitri, line 276: I'm not sure I follow how this is not an ERROR (or a WARN) situation? When does that happen in a normal execution?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +250,136 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+		LogicalMessageMetadata metadata = { 0 };
+		if (!parseSQLAction((char *) line, &metadata))
+		{
+			/* errors have already been logged */
+			return false;
+		}
+
+		/* Create a txid as a symlink to the buffer file */
+		char txidFile[MAXPGPATH];
+		sformat(txidFile, MAXPGPATH, "%s/%d.sql", context->paths.dir, metadata.xid);
```
dimitri, line 287: Again, let's expose the current name as `replayCtx->currentXIDfilename` or something.

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +250,136 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+
+		fformat(replayCtx->txnReplayBuffer, "%s\n", line);
+	}
+	else if (strstr(line, OUTPUT_COMMIT) == line)
+	{
+		if (replayCtx->txnReplayBuffer == NULL)
+		{
+			log_debug("Ignore COMMIT message received without a preceding BEGIN");
+			return true;
+		}
```
dimitri, line 314: Same question, does that ever happen without being an ERROR or at least a WARNing? when does that situation occur in a normal execution?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +250,136 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+		};
+
+		if (!read_from_stream(txnReplayBufferForRead, &readerContext))
+		{
+			log_error("Failed to read from replay buffer, "
+					  "see above for details");
+			return false;
+		}
+
+		fclose(txnReplayBufferForRead);
```
dimitri, line 346: I believe this should be its own function (`streamReplayXIDFromFile` or something) and receive the `replayCtx->currentXIDfilename` as an argument.

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +250,136 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+	}
+	else
+	{
+		if (replayCtx->txnReplayBuffer != NULL)
+		{
+			/*
+			 * We are in a transaction block, buffer all the messages
+			 * including KEEPALIVE, SWITCHWAL and ENPOS.
+			 */
+			fformat(replayCtx->txnReplayBuffer, "%s\n", line);
```
dimitri, line 364: Do we need to track `bool isTransaction` separately from the `FILE *` pointer being NULL, for ease of code maintenance and possibly advanced/intricate debugging?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +250,136 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+ * transation metadata file which is created by the transform process.
+ */
+static bool
+streamReplayLineIntoBuffer(void *ctx, const char *line, bool *stop)
+{
+	ReplayStreamCtx *replayCtx = (ReplayStreamCtx *) ctx;
+	StreamApplyContext *context = &(replayCtx->applyContext);
+
+	if (strstr(line, OUTPUT_BEGIN) == line)
+	{
```
dimitri, line 269: Could we maybe call `parseSQLAction` first and then use the metadata action field to sort out what's happening here, rather than comparing strings again?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +250,136 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+		{
+			/*
+			 * We are in a transaction block, buffer all the messages
+			 * including KEEPALIVE, SWITCHWAL and ENPOS.
+			 */
+			fformat(replayCtx->txnReplayBuffer, "%s\n", line);
+		}
+		else if (strstr(line, OUTPUT_KEEPALIVE) == line ||
+				 strstr(line, OUTPUT_SWITCHWAL) == line ||
+				 strstr(line, OUTPUT_ENDPOS) == line)
```
dimitri, line 368: We should most probably have some metadata action field to use here.

dimitri (changes requested):
Thanks for your work, I believe we're close to being able to merge this improvement! Please see my review and consider the changes I'm asking for.

src/bin/pgcopydb/ld_replay.c
arajkumar: The idea is to use this for prefetch mode as well, during the resume, we might encouter partially written txns. But that doesn't apply for replay. I would rather add an error message here.

src/bin/pgcopydb/ld_replay.c
arajkumar: I actually avoided because it allocates memory for jsonBuffer for EXECUTE message which we never use. I will change it to use parseSQLAction.

src/bin/pgcopydb/ld_replay.c
dimitri: Even in prefetch mode we would only encounter a single partial transaction I believe, the last one we received. Cases where we would deal with a partial transaction and then another transaction certainly indicate a bug?

arajkumar: >Even in prefetch mode we would only encounter a single partial transaction I believe, the last one we received.

This is what is happening. We have to ignore the partial transactions (i.e. either a COMMIT without a BEGIN or BEGIN without COMMIT).

arajkumar: I addressed all the review comments except this.

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +255,166 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+	LogicalMessageMetadata metadata = { 0 };
+
+	if (!parseSQLAction((char *) line, &metadata))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	/* We don't use this metadata in this function */
+	free(metadata.jsonBuffer);
```
dimitri, line 318: Could we improve the parseSQLAction in a way that allows for parsing the metadata only, or maybe split the function in two parts somehow?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +255,166 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+
+	if (metadata.action == STREAM_ACTION_BEGIN)
+	{
+		if (replayCtx->isTranscationBuffering)
+		{
+			log_debug("Received %s when transaction is already "
+					  "in buffering mode, ignoring.", line);
+			fclose(replayCtx->txnBuffer);
+		}
```
dimitri, line 327: Looks like an ERROR situation to me still. After reading your comment, I believe the way we read the current SQL file from the transform module at resume allows us to track in the context that there is already a transaction in progress, in such a way that we could avoid being surprised here, right?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +255,166 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+		replayCtx->isTranscationBuffering = true;
+	}
+	else if (metadata.action == STREAM_ACTION_COMMIT)
+	{
+		if (!replayCtx->isTranscationBuffering)
+		{
+			log_debug("Received %s when transaction is not "
+					  "in buffering mode, ignoring.", line);
+			return true;
+		}
```
dimitri, line 366: Maybe that's when we could know that a transaction is in progress even though we didn't read the BEGIN message in this session (resuming transform case etc).

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +255,166 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+		unlink_file(context->paths.txnlatestfile);
+
+		/* Early exit if we reached the end position */
+		*stop = context->reachedEndPos;
+
+		replayCtx->txnBuffer = NULL;
+	}
+	else
+	{
+		if (replayCtx->txnBuffer != NULL)
```
dimitri, line 392: I believe we should use `isTransactionBuffering` here (and fix the typo in the variable name).

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +255,166 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+}
+
+
+/* streamReplayLineIntoBuffer is a callback function for the ReadFromStreamContext
+ * and read_from_stream infrastructure. It's called on each line read from a
+ * stream such as a unix pipe and buffers the line into a file until we encounter
+ * a COMMIT. This ensures that we can replay the whole transaction in one go
+ * and avoids deadlock when the BEGIN messages arrives before the existence of
+ * transation metadata file which is created by the transform process.
+ */
```
dimitri, line 302: Is there a way to activate buffering only when a transaction does not have its COMMIT LSN attached in the metadata? So that in the case when we have a typical OTLP system with many one-command DML transactions we don't have any penalty?

dimitri (changes requested):
Getting there!

src/bin/pgcopydb/ld_replay.c
dimitri: See next series of comments. Resolving this one.

src/bin/pgcopydb/ld_replay.c
dimitri: Maybe the way to make it look like a DEBUG level information (or NOTICE actually?) is to add proper detailed comments about how/when the situation occurs and why it's not a WARN or ERROR entry?

src/bin/pgcopydb/ld_replay.c
arajkumar: Ahh, I missed to stage this change! Will do it.

src/bin/pgcopydb/ld_replay.c
arajkumar: It seem to be tightly coupled. Lets add a TODO and try to fix it in a separate PR.

src/bin/pgcopydb/ld_replay.c
arajkumar: Consider there are 2 sql files, A.sql & B.sql. B.sql has a continued transaction and the `catchup` process has exited because there is no C.sql available to apply. In the meantime C.sql was written partially by the `transform`. After switching to `replay` mode, we recognise that C.sql exists and try to apply it. But this is not going to work because  B.sql was executed in `prefetch` SQL connection/session which is no longer available in `replay` mode. Does it really makes sense to resume with the partial C.sql ?

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +257,300 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+			 * For example, lets consider the partially written C.sql file,
+			 *
+			 * BEGIN -- {"xid": 1000, "commitLSN": "0/1234"};
+			 * INSERT INTO C VALUES (1);
+			 * INSERT INTO C VALUES (2);
+			 * KEEPALIVE;
+			 *
+			 * After switching to the replay mode, logical decoding will resume
+			 * from consistent point which again starts with a valid transcation
+			 * block.
```
dimitri, line 354: Do we have the LSN value for that consistent point available in the context here? If yes, I believe we should be using it to make sure we are skipping transactions as intended. If no, then we should add to the comment why not, so that another PR later can go and fix the situation if we feel like it should be fixed.

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +257,300 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+			 *
+			 * Contents of C.sql will be streamed by stream_transform_stream
+			 * and followed by the contents from UNIX PIPE. Since both of the
+			 * contents are streamed one after the other, the second block
+			 * contains the full content of previously written
+			 * transaction. So, we can skip previous transaction buffer and
+			 * start buffering from the current transaction.
+			 */
+			log_debug("Received %s when transaction is already "
+					  "in buffering mode, ignoring.", line);
```
dimitri, line 376: Please rewrite the message so that it ends with `": %s", line`. This makes it easier to process the logs later.

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +257,300 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+			 *
+			 * Contents of C.sql will be streamed by stream_transform_stream
+			 * and followed by the contents from UNIX PIPE. Since both of the
+			 * contents are streamed one after the other, the second block
+			 * contains the full content of previously written
+			 * transaction. So, we can skip previous transaction buffer and
+			 * start buffering from the current transaction.
+			 */
+			log_debug("Received %s when transaction is not "
+					  "in buffering mode, ignoring.", line);
```
dimitri, line 459: Please rewrite the message so that it ends with `": %s", line`. This makes it easier to process the logs later.

src/bin/pgcopydb/ld_replay.c
```diff
@@ -249,3 +257,300 @@ stream_replay_line(void *ctx, const char *line, bool *stop)
...
+			 *
+			 * Contents of C.sql will be streamed by stream_transform_stream
+			 * and followed by the contents from UNIX PIPE. Since both of the
+			 * contents are streamed one after the other, the second block
+			 * contains the full content of previously written
+			 * transaction. So, we can skip previous transaction buffer and
+			 * start buffering from the current transaction.
+			 */
+			log_debug("Received %s when transaction is not "
+					  "in buffering mode", line);
```
dimitri, line 551: Please rewrite the message so that it ends with `": %s", line`. This makes it easier to process the logs later.

dimitri (changes requested):

src/bin/pgcopydb/ld_replay.c
dimitri: I think I still have some with the code (the comments are good at explaining the situation, thanks). The trouble is that we don't protect against applying this approach even in the middle of our operations, when it looks like it's only needed before we reach our own “consistent point”, never after that. Can we make sure that as soon as we start replaying transactions, then we raise the situation from a DEBUG information to an actual ERROR? (well assuming my analysis is correct here).

src/bin/pgcopydb/ld_replay.c
dimitri: Thanks for the comments in the source code. See my other review comments here for further discussion.

src/bin/pgcopydb/ld_replay.c
dimitri: Thanks!

src/bin/pgcopydb/ld_replay.c
arajkumar: @dimitri  This is a great point !  I think the current ignore logic implemented in this PR might cause inconsistency where it might ignore the partial transaction, but applies the next transaction because it has valid BEGIN/COMMIT block. 

I need to have a another deep look now.

arajkumar: @dimitri I was wrong in my previous comment. 

While switching to replay mode, follow ensures that the pending LSNs in transform queue has been [processed and applied to the target](https://github.com/dimitri/pgcopydb/blob/672ad1228278f22cd450b214afc3e0092ac47a15/src/bin/pgcopydb/follow.c#L429-L489).  This ensures everything except the last the partial transaction has been successfully committed into the target. After switching to replay mode, [stream_transform_resume](https://github.com/dimitri/pgcopydb/blob/672ad1228278f22cd450b214afc3e0092ac47a15/src/bin/pgcopydb/ld_transform.c#L223-L253) again streams the last partial file(which is already applied by the previous step) to the replay process which leads to this scenario. I think streaming last partial .sql from [stream_transform_resume](https://github.com/dimitri/pgcopydb/blob/672ad1228278f22cd450b214afc3e0092ac47a15/src/bin/pgcopydb/ld_transform.c#L223-L253) is redundant and shall be removed and this shouldn't cause any issue. WDYT?

arajkumar: I'm wondering are these PRs https://github.com/dimitri/pgcopydb/pull/348, https://github.com/dimitri/pgcopydb/pull/277 trying to do the same in different ways?

dimitri: IIRC the transform parts need to parse/load the latest file to build up the context in case we stopped in the middle of a transaction (we then need to have the start of the transaction in-memory); but I agree we don't need to send that to the replay process, which is only going to skip again.

That said the skip logic is sound and I wonder why in your PR we can't benefit from it?

arajkumar:
https://github.com/dimitri/pgcopydb/pull/525 fixes by eliminating the need of txn metadata file.

Issue #508, opened 2023-10-24
During database parameter copying ALTER SYSTEM SET seems to operate using the source database name


kbarber:
Greetings,

I noticed this since this commit: https://github.com/dimitri/pgcopydb/commit/cc702401d0b8c52a8877376b5a776bb207595407

We seem to get this error message when cloning from a source and target database that have different names:

```
2023-10-24 13:21:05 7 INFO   main.c:163                Running pgcopydb version 0.13.41.ga3a577c from "/root/pgcopydb/pgcopydb"
2023-10-24 13:21:05 7 INFO   cli_common.c:1239         [SOURCE] Copying database from "postgres://postgres@source.host:5432/sourcedb?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
2023-10-24 13:21:05 7 INFO   cli_common.c:1240         [TARGET] Copying database into "postgres://admin@target.host:5432/targetdb?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
2023-10-24 13:21:05 7 INFO   snapshot.c:100            Exported snapshot "0000000D-00000123-1" from the source database
2023-10-24 13:21:05 15 INFO   cli_clone_follow.c:530    STEP 1: fetch source database tables, indexes, and sequences
2023-10-24 13:21:05 15 INFO   copydb_schema.c:143       Fetched information for 2 extensions
2023-10-24 13:21:06 15 INFO   copydb_schema.c:282       Splitting source candidate tables larger than 47 MB
2023-10-24 13:21:06 15 INFO   copydb_schema.c:440       Fetched information for 50 tables, with an estimated total of 2990  tuples and 633 MB
2023-10-24 13:21:06 15 INFO   copydb_schema.c:466       Fetched information for 50 indexes
2023-10-24 13:21:06 15 INFO   sequences.c:41            Fetching information for 50 sequences
2023-10-24 13:21:06 15 WARN   pgsql.c:608               Failed to connect to "postgres://admin@target.host:5432/targetdb?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60", retrying until the server is ready
2023-10-24 13:21:06 15 INFO   pgsql.c:693               Successfully connected to "postgres://admin@target.host:5432/targetdb?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60" after 2 attempts in 58 ms.
[...]

PR #509, opened 2023-10-25
Use target dbname for db param setting

kbarber:
Draft fix for: https://github.com/dimitri/pgcopydb/issues/508


dimitri (approved):

kbarber:
Thanks @dimitri I've rebased and fixed the comment on the commit message.

I've tested this on about 750 transfers in our suite and it seems stable. Thanks for your help.

dimitri:
> I've tested this on about 750 transfers in our suite and it seems stable. Thanks for your help.

Wow, that sounds like a very interesting use-case for pgcopydb there!

kbarber:
> > I've tested this on about 750 transfers in our suite and it seems stable. Thanks for your help.
> 
> Wow, that sounds like a very interesting use-case for pgcopydb there!

I'm testing multi-tenant transfers from a single instance using your tool. I have 750 test databases in my dev env and i'm transferring from 1 service to 20 services to account that writes are heavier. I've got a splay of different sizes from 0 to 45 GB to try in that mix for each db so its a fun way to test things quickly. Takes about 5 hours (the big ones create a tail). I'm doing about 60 transfers at once atm but depends on how the day goes.

kbarber:
@dimitri we keep getting a test failure, I've only touched your testing a little bit. Is there something I need to fix here?

PR #514, opened 2023-10-26
Ensure NULL-termination of string when unescaping single-quotes

acritox:
The for-loop is copying a string character by character and replacing `''` with `'`
by skipping a character if the current and next one are both `'`. When this
replacement doesn't occur then the target string will be filled completly
without a `\0' terminator at the end of the string.

here's a (redacted) screenshot showing the effect of this bug:
image
which then results in an apply error like:
> ERROR:  bind message supplies 6 parameters, but prepared statement "…" requires 7

or it could result in corrupted data if if the uninitialized memory happens to be valid UTF-8 bytes/characters for JSON strings


dimitri (approved):

dimitri:
Thanks @acritox  for your contribution, much appreciated!

PR #517, opened 2023-10-28
exclude generated columns from COPY statements

### LINKED ISSUE #496: Copy error with generated columns
I'm am preparing a massive migration using pgcopydb and face this issue :

`ERROR  [TARGET 66057] DETAIL:  Generated columns cannot be used in COPY.`

Is it supported yet? 

Thanks a lot for this wonderful tool

VaibhaveS:
- add _isgenerated_ in table attribute 
- exclude generated attributes from COPY

fixes #496


VaibhaveS:
Generated columns are not supported before PG 12, so we need to follow some other approach to get the isgenerated column value 

@dimitri can you please suggest

dimitri:
> Generated columns are not supported before PG 12, so we need to follow some other approach to get the isgenerated column value
> 
> @dimitri can you please suggest, what's the best way to achieve this

In that very case I would try to see if we can use the `pg_depend` catalog instead to get the information, the same way we do for sequences, where we have the following join:

```
		"       left join pg_depend d1 on d1.objid = s.seqoid "
		"        and d1.classid = 'pg_class'::regclass "
		"        and d1.refclassid = 'pg_class'::regclass "
		"        and d1.deptype in ('i', 'a') "
```

You should be able to join pg_attribute directly to its pg_depend entry and check if it's a deptype 'i' for *internal dependency* which means *generated column* i believe, where the deptype `a` for `automatic` refers to a sequence default value dependency.

src/bin/pgcopydb/schema.c
```diff
@@ -971,12 +971,17 @@ struct FilteringQueries listSourceTablesSQL[] = {
...
+		"						  d.deptype = 'i' as attisgenerated "
 		"                    from pg_attribute a "
+				"                    from pg_attribute a "
 		"                         left join pg_index i "
 		"                                on i.indrelid = a.attrelid "
 		"                               and a.attnum = ANY(i.indkey) "
 		"                               and i.indisprimary "
+		"					 	  left join pg_depend d "
+		"					 	  		on d.refobjid = a.attrelid "
+		"					 	  		and d.refobjsubid = a.attnum "
```
dimitri, line 984: Please also use `pg_depend.classid = 'pg_class':regclass` as a join qualifier.

src/bin/pgcopydb/table-data.c
```diff
@@ -1182,12 +1183,21 @@ copydb_prepare_copy_query(CopyTableDataSpec *tableSpecs,
 
 		for (int i = 0; i < table->attributes.count; i++)
 		{
+			/* Generated columns cannot be used in COPY */
+			if (table->attributes.array[i].attisgenerated)
+			{
+				continue;
+			}
+
 			char *attname = table->attributes.array[i].attname;
```
dimitri, line 1192: Nitpicking: now that we access to more than one slot of the structure instance, could you maybe declare a new local variable and use it, as in:

``` c
    SourceTableAttribute *attribute = &(table->attributes.array[i]);
```

And then use `attribute->attname` and attribute->attisgenerated`.

dimitri (changes requested):
Almost there, mostly nitpicking comments that still require a change. The pg_depend join missing a qualifier on the `classid` column is a big deal, though.

src/bin/pgcopydb/schema.c
VaibhaveS: yep, actually I am still not sure if that's enough. It seemed like there are duplicate entries in the attributes because of the join in the pagila rental schema, probably because there are multiple entries in pg_depend for the same attribute? 

I need to refactor the query more.

dimitri: Yeah I see you're only joining to one side of the dependency entry. You need to join with the other side also (`classid` and `objid` in that case), and for that you might need to grab information from other catalogs.

src/bin/pgcopydb/schema.c
```diff
@@ -971,14 +971,19 @@ struct FilteringQueries listSourceTablesSQL[] = {
...
+		"                         i.indrelid is not null as attisprimary, "
+		"						  col.is_generated = 'ALWAYS' as attisgenerated "
 		"                    from pg_attribute a "
 		"                         left join pg_index i "
-		"                                on i.indrelid = a.attrelid "
+		"                               on i.indrelid = a.attrelid "
 		"                               and a.attnum = ANY(i.indkey) "
 		"                               and i.indisprimary "
+		"					 	  left join information_schema.columns col "
+        "                    			on col.column_name = a.attname "
```
dimitri, line 982: That's a good way to get at the information too, but I believe you need to also make sure you're targeting the column from the right table in information_schema: what if we have multiple tables using the same column name?

src/bin/pgcopydb/schema.c
```diff
@@ -971,14 +971,19 @@ struct FilteringQueries listSourceTablesSQL[] = {
...
 		"                   where a.attrelid = c.oid and not a.attisdropped "
 		"                     and a.attnum > 0 "
+		"					  and col.table_name = c.relname "
+		"          			  and col.table_schema = n.nspname "
```
VaibhaveS, line 986: @dimitri, these conditions ensure we have the right column

dimitri: Well then they need to be part of the JOIN condition (modern SQL style, from 1989 onwards I believe) rather than in the WHERE clause (legacy SQL style).

src/bin/pgcopydb/schema.c
```diff
@@ -971,14 +971,19 @@ struct FilteringQueries listSourceTablesSQL[] = {
 		"               ("
 		"                  select attnum, atttypid::integer, "
 		"                         format('%I', attname) as attname, "
-		"                         i.indrelid is not null as attisprimary "
+		"                         i.indrelid is not null as attisprimary, "
+		"						  col.is_generated = 'ALWAYS' as attisgenerated "
 		"                    from pg_attribute a "
 		"                         left join pg_index i "
-		"                                on i.indrelid = a.attrelid "
+		"                               on i.indrelid = a.attrelid "
```
dimitri, line 978: Can you cleanup your PR to avoid useless whitespace changes? (the spacing here was not a mistake btw, it was intended that way).

src/bin/pgcopydb/schema.c
```diff
@@ -971,14 +971,19 @@ struct FilteringQueries listSourceTablesSQL[] = {
...
 		"                               and a.attnum = ANY(i.indkey) "
 		"                               and i.indisprimary "
+		"					 	  left join information_schema.columns col "
+        "                    			on col.column_name = a.attname "
+		"					  			and col.table_name = c.relname "
+		"          			 			and col.table_schema = n.nspname "
 		"                   where a.attrelid = c.oid and not a.attisdropped "
-		"                     and a.attnum > 0 "
+		"                     and a.attnum > 0 "
```
dimitri, line 986: Another whitespace cleanup to do before merge, thanks!

src/bin/pgcopydb/table-data.c
```diff
@@ -1182,12 +1183,23 @@ copydb_prepare_copy_query(CopyTableDataSpec *tableSpecs,
...
 		{
-			char *attname = table->attributes.array[i].attname;
+			SourceTableAttribute *attribute = &(table->attributes.array[i]);
+			char *attname = attribute->attname;
 
-			if (i > 0)
+			/* Generated columns cannot be used in COPY */
+			if (attribute->attisgenerated)
+			{
+				log_info("Skipping %s in COPY as it is a generated column", attname);
```
dimitri, line 1192: Do we need that information at the INFO level? I suppose NOTICE would be preferable.

src/bin/pgcopydb/table-data.c
```diff
@@ -1242,12 +1254,21 @@ copydb_prepare_copy_query(CopyTableDataSpec *tableSpecs,
...
-		{
-			char *attname = table->attributes.array[i].attname;
+		{			
+			SourceTableAttribute *attribute = &(table->attributes.array[i]);
+			char *attname = attribute->attname;
+
+			/* Generated columns cannot be used in COPY */
+			if (attribute->attisgenerated)
+			{
+				log_info("Skipping %s in COPY as it is a generated column", attname);
```
dimitri, line 1264: Again, looks like a NOTICE message at most.

dimitri (changes requested):
Mostly nitpicking. Please consider a last round of changes before merge, thanks!

dimitri (approved):

arajkumar:
@dimitri We are facing issues with generated column in CDC mode. I think we can fix it by testing whether the column is generated or not from ld_transform.c, but it requires catalog access always. Do you think it is a fair tradeoff to require catalog all the time to support excluding the generated columns?

dimitri:
> @dimitri We are facing issues with generated column in CDC mode. I think we can fix it by testing whether the column is generated or not from ld_transform.c, but it requires catalog access always. Do you think it is a fair tradeoff to require catalog all the time to support excluding the generated columns?

Yes indeed. We also need catalog access in transform when using test_decoding in order to lookup attisprimary properties.

PR #522, opened 2023-10-31
fix password handling in safeURI

acritox:
when the passwords of source and target differ it will always fail on the first connection attempt because it tries to log in with the password for the other DB:

reproducible test case:
```
$ git switch main && make -s bin
Already on 'main'
Your branch is up to date with 'origin/main'.
$ env|grep PGCOPYDB_
PGCOPYDB_SOURCE_PGURI=postgres://testuser1:testPassword@localhost/testdb1
PGCOPYDB_TARGET_PGURI=postgres://testuser2:testPassword@localhost/testdb2
$ ./src/bin/pgcopydb/pgcopydb clone --restart |& grep connect
$ psql -e postgres://postgres@localhost --command="ALTER ROLE testuser2 PASSWORD 'a_different_password';"
ALTER ROLE testuser2 PASSWORD 'a_different_password';
ALTER ROLE
Time: 12.275 ms
$ export PGCOPYDB_TARGET_PGURI="postgres://testuser2:a_different_password@localhost/testdb2"
$ ./src/bin/pgcopydb/pgcopydb clone --restart |& grep connect
2023-10-31 16:21:24 2172643 WARN   pgsql.c:608               Failed to connect to "postgres://testuser2@localhost/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60", retrying until the server is ready
2023-10-31 16:21:24 2172643 INFO   pgsql.c:693               Successfully connected to "postgres://testuser2@localhost/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60" after 2 attempts in 88 ms.
2023-10-31 16:21:25 2172643 WARN   pgsql.c:608               Failed to connect to "postgres://testuser2@localhost/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60", retrying until the server is ready
2023-10-31 16:21:25 2172643 INFO   pgsql.c:693               Successfully connected to "postgres://testuser2@localhost/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60" after 2 attempts in 92 ms.
2023-10-31 16:21:25 2172662 WARN   pgsql.c:608               Failed to connect to "postgres://testuser2@localhost/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60", retrying until the server is ready
2023-10-31 16:21:25 2172662 INFO   pgsql.c:693               Successfully connected to "postgres://testuser2@localhost/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60" after 2 attempts in 102 ms.
$ git switch fix_different_passwords; make -s bin
Switched to branch 'fix_different_passwords'
Your branch is up to date with 'origin/fix_different_passwords'.
$ ./src/bin/pgcopydb/pgcopydb clone --restart |& grep connect
$ 
```

no output from grep = no issue, therefore this commit fixes the bug

### DIFF
--- a/src/bin/pgcopydb/defaults.h
+++ b/src/bin/pgcopydb/defaults.h
@@ -58,9 +58,6 @@
 
 #define POSTGRES_PORT 5432
 
-/* masqurade  passwords with that value in logs */
-#define PASSWORD_MASK "****"
-
 /* default replication slot and origin for logical replication */
 #define REPLICATION_ORIGIN "pgcopydb"
 #define REPLICATION_PLUGIN "test_decoding"
--- a/src/bin/pgcopydb/parsing_utils.c
+++ b/src/bin/pgcopydb/parsing_utils.c
@@ -936,9 +936,9 @@ escapeWithPercentEncoding(const char *str, char **dst)
 
 
 /*
- * uri_contains_password takes a Postgres connection string and checks to see
- * if it contains a parameter called password. Returns true if a password
- * keyword is present in the connection string.
+ * uri_grab_password takes a Postgres connection string and checks to see
+ * if it contains a parameter called password and if so stores a copy of it
+ * in safeURI->password.
  */
 static bool
 uri_grab_password(const char *pguri, SafeURI *safeURI)
@@ -984,8 +984,8 @@ uri_grab_password(const char *pguri, SafeURI *safeURI)
 
 /*
  * parse_and_scrub_connection_string takes a Postgres connection string and
- * populates scrubbedPguri with the password replaced with **** for logging.
- * The scrubbedPguri parameter should point to a memory area that has been
+ * populates safeURI without the password for logging purposes.
+ * The safeURI parameter should point to a memory area that has been
  * allocated by the caller and has at least MAXCONNINFO bytes.
  */
 bool
--- a/src/bin/pgcopydb/pgcmd.c
+++ b/src/bin/pgcopydb/pgcmd.c
@@ -793,7 +793,7 @@ pg_restore_db(PostgresPaths *pgPaths,
 	setenv("PGCONNECT_TIMEOUT", POSTGRES_CONNECT_TIMEOUT, 1);
 
 	/* override PGPASSWORD environment variable if the pguri contains one */
-	if (connStrings->safeSourcePGURI.password != NULL)
+	if (connStrings->safeTargetPGURI.password != NULL)
 	{
 		if (pgpassword_found_in_env && !get_env_dup("PGPASSWORD", &PGPASSWORD))
 		{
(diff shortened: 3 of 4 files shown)


dimitri:
Thanks for this contribution!

After careful consideration I see that we rely on changing the environment internally within a process, without starting a new command, and that's undefined behavior / unsupported. We should just skip the PGPASSWORD environment processing altogether and use `pgsql->connectionString` here really.

acritox:
yes that sounds good, too - I think `pgsql->connectionString` is what the second connection attempt uses and why it succeeds then

acritox:
yes, `pgsql_retry_open_connection()` uses [`pgsql->connectionString`](https://github.com/dimitri/pgcopydb/blob/main/src/bin/pgcopydb/pgsql.c#L679), so [the change in this line](https://github.com/dimitri/pgcopydb/pull/457/files#diff-84ab9d926dfb0eda4b691a03c25ba22211e96c9f749fa57ff44ca24758bb5777R533) was probably an error and I'll change it back.

I found PR #328 which might explain why there was some PGPASSWORD environment processing in the implemented first place:
> we set the password in the environment of the child processes when running pg_dump etc commands

so it makes sense having that in pgcmd.c but in pgsql.c it shouldn't be needed, or should it?

there's actually a bit of a cleanup left that should be done now from PR#328, e.g. the comments for [`uri_grab_password()`](https://github.com/dimitri/pgcopydb/blob/main/src/bin/pgcopydb/parsing_utils.c#L938-L944) 
(it's **not** `uri_contains_password()`) and [`parse_and_scrub_connection_string()`](https://github.com/dimitri/pgcopydb/blob/main/src/bin/pgcopydb/parsing_utils.c#L985-L992) are not correct anymore, [PASSWORD_MASK](https://github.com/dimitri/pgcopydb/blob/main/src/bin/pgcopydb/defaults.h#L60-L62) is not used anywhere currently

I'll change those, try it out and will update this PR if it works.

dimitri:
> so it makes sense having that in pgcmd.c but in pgsql.c it shouldn't be needed, or should it?

Not needed in pgsql.c ; it's undefined behaviour and we'd rather not count on that and simply use `.connectionString` instead.

acritox:
PR updated - took a bit because I stumbled into another bug while testing 😆 

`pgcopydb clone` just silently exited at step 3 in some cases I tested…

why? because the `pg_restore` subprocess segfaulted when the Source URI had a password but the Target URI did not (so it used $PGPASSWORD from env), i.e.:

```
export PGPASSWORD=testPassword
export PGCOPYDB_SOURCE_PGURI=postgres://testuser1:testPassword@localhost/testdb1
export PGCOPYDB_TARGET_PGURI=postgres://testuser2@localhost/testdb2
pgcopydb clone
```

found out that `safeSourcePGURI` and `safeTargetPGURI` were interchanged in two places in `pg_restore_db()`

dimitri (approved):

Issue #350, opened 2023-06-28
Parsing error with test_decoding plugin


ulo23k:
Facing parsing issues with test_decoding plugin in the below error message:

```
2023-06-24 11:41:47 20041 INFO   pgsql.c:4267              Reported write_lsn 24B1/F40FA20, flush_lsn 0/0, replay_lsn 0/0
2023-06-24 11:41:55 20042 ERROR  ld_test_decoding.c:909    Failed to parse decoding message for UPDATE on table "public"."audition" which is not in our catalogs
2023-06-24 11:41:55 20042 ERROR  ld_test_decoding.c:500    Failed to parse UPDATE new-tuple columns for logical message table public.audition: UPDATE: audition_id[text]:'5127494f-361f-4774-837a-24ca89f6916f' person_id[bigint]:96476415 created_ts[timestamp with time zone]:'2023-06-23 03:58:26.044+00' last_seen[timestamp with time zone]:'2023-06-24 11:41:37.277+00' correlation_id[text]:'4dcbe359-ecd5-461e-ba9e-dc5508a7d611:2420286' context_guid[text]:'a9cab2e8-8adf-42cd-8108-fa01137b0aa9' request_type[integer]:1 segmentation_name[text]:null audition_hash[text]:'e002a33d4f5e4b447d7ad4f1640e813f95ee34355be5fbf81014c63229ce9b16' request_hash[text]:'c524c7407b02e2f979e98ea62c5de38761afeae582108f64f8075731f99f2e92' complete_request[text]:'UkFMWUhMVEgAAAAAAAAAAErT7S+y9a6YHQPasU/Ql0ON1L1RzR5CsSbC8DdTlL/0hUjT/0DasHqC/YATpwyelDeOL/50+759V8m5IFq6+FUe4nZIASv1vwKukJRLrNU0D8ealw1sKBt3KmnqnzBVVHNG+sMs8u1e/eFGGdNPbMApzF0A5iVX8JZDie089U3/GHxU+67ocahM493QokX2ST59TzJu1NeziG62TNykLVs=' complete_audition[bytea]:unchanged-toast-datum event_date[timestamp with time zone]:null
[...]

PR #523, opened 2023-10-31
fix decoding in transform subprocess when following

acritox:
without this change `pgcopydb follow` would fail with e.g.:
> Failed to parse decoding message for UPDATE on table "public"."tablename" which is not in our catalogs

I suspect this PR fixes issue #350

how I have successfully used it with this change:
```
$ pgcopydb snapshot --follow &
$ pgcopydb clone --no-owner --no-role-passwords --no-acl --no-comments --drop-if-exists --verbose
```

and then (while the `pgcopydb clone` is still running) I ran `pgcopydb follow` in a second shell so it would prefetch the changes

### DIFF
--- a/src/bin/pgcopydb/follow.c
+++ b/src/bin/pgcopydb/follow.c
@@ -267,9 +267,10 @@ follow_main_loop(CopyDataSpec *copySpecs, StreamSpecs *streamSpecs)
 	}
 
 	/*
-	 * Read the catalogs from the source table from on-file disk.
+	 * Read the catalogs from the source table from on-file disk if
+	 * the schema dump has been done already.
 	 */
-	if (!copydb_parse_schema_json_file(copySpecs))
+	if (copySpecs->dirState.schemaDumpIsDone && !copydb_parse_schema_json_file(copySpecs))
 	{
 		/* errors have already been logged */
 		return false;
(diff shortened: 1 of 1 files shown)


acritox:
Ok, I've rebased to current main again and added a condition to it, now it will [only attempt to read catalogs if schema dump has been done already](https://github.com/dimitri/pgcopydb/pull/523/commits/212c748857f956247980ed6979982ea86dbdd678)

dimitri:
> Ok, I've rebased to current main again and added a condition to it, now it will [only attempt to read catalogs if schema dump has been done already](https://github.com/dimitri/pgcopydb/pull/523/commits/212c748857f956247980ed6979982ea86dbdd678)

Thanks for your contribution! how does the follow process work at all without a copy of the schema then? I suppose you have only tested with wal2json plugin? The test_decoding plugin needs to have a list of the primary key columns in-memory for every table to be able to parse the UPDATE statements in many cases...

acritox:
I have only been using `test_decoding`, never tried `wal2json` (not installed on my server) - that only came up during the failing tests here on Github

to be honest I don't know how/why it works then, I had just always gotten the `Failed to parse decoding message for UPDATE on table "public"."tablename" which is not in our catalogs` error and tried to solve it somehow.

I suspected that the follow process was in fact missing the schema data because I didn't see it reading it, so I figured I'll add that - and now it works

dimitri:
> I suspected that the follow process was in fact missing the schema data because I didn't see it reading it, so I figured I'll add that - and now it works

I see, I was reading it the other way around. Nice! Second round of review coming now.

dimitri (approved):

PR #524, opened 2023-10-31
compare: implement --snapshot option for source database

acritox:
to compare a source snapshot to the target

usecase: to check and make sure that data on source and target has not diverged when CDC has been ongoing for a while

1. pause the ongoing CDC
2. create new snapshot on source:
```
$ pgcopydb snapshot --follow --slot pgcopydbcompare --dir /tmp/pgcopydb-cmpsnap &
INFO   Created logical replication slot "pgcopydbcompare" with plugin "test_decoding" at 11/11111110 and exported snapshot 00000000-00000000-0
```
4. `pgcopydb follow --resume` with endpos set to the position the snapshot is at (e.g. 11/11111110)
5. wait until the target has been caught up to that position
6. now `pgcopydb compare --snapshot 00000000-00000000-0` can be used to compare

### DIFF
--- a/src/bin/pgcopydb/cli_compare.c
+++ b/src/bin/pgcopydb/cli_compare.c
@@ -33,7 +33,8 @@ static CommandLine compare_schema_command =
 		" --source ... ",
 		"  --source         Postgres URI to the source database\n"
 		"  --target         Postgres URI to the target database\n"
-		"  --dir            Work directory to use\n",
+		"  --dir            Work directory to use\n"
+		"  --snapshot       Use snapshot obtained with pg_export_snapshot on the source database\n",
 		cli_compare_getopts,
 		cli_compare_schema);
 
@@ -45,7 +46,8 @@ static CommandLine compare_data_command =
 		"  --source         Postgres URI to the source database\n"
 		"  --target         Postgres URI to the target database\n"
 		"  --dir            Work directory to use\n"
-		"  --json           Format the output using JSON\n",
+		"  --json           Format the output using JSON\n"
+		"  --snapshot       Use snapshot obtained with pg_export_snapshot on the source database\n",
 		cli_compare_getopts,
 		cli_compare_data);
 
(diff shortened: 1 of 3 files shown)


src/bin/pgcopydb/compare.c
```diff
@@ -410,6 +410,27 @@ compare_table(CopyDataSpec *copySpecs, SourceTable *source)
...
+			(void) pgsql_finish(&src);
+			return false;
+		}
+
+		if (!pgsql_set_snapshot(&src, copySpecs->sourceSnapshot.snapshot))
+		{
+			/* errors have already been logged */
+			(void) pgsql_finish(&src);
+			return false;
+		}
```
dimitri, line 429: Can we just use `copydb_set_snapshot` from snapshot.c?

src/bin/pgcopydb/compare.c
```diff
@@ -801,6 +822,10 @@ compare_fetch_schemas(CopyDataSpec *copySpecs,
 	*sourceSpecs = *copySpecs;
 	*targetSpecs = *copySpecs;
 
+	/* unset sourceSnapshot snapshot for target */
+	targetSpecs->consistent = false;
+	(void) memset(&(targetSpecs->sourceSnapshot.snapshot), 0, sizeof(targetSpecs->sourceSnapshot.snapshot));
+
```
dimitri, line 828: Why do we need to do this? can't we use `copydb_close_snapshot` ?

src/bin/pgcopydb/cli_compare.c
```diff
@@ -301,11 +312,23 @@ cli_compare_schema(int argc, char **argv)
 		exit(EXIT_CODE_INTERNAL_ERROR);
 	}
 
+	if (!IS_EMPTY_STRING_BUFFER(compareOptions.snapshot) && !copydb_set_snapshot(&copySpecs))
```
dimitri, line 315: nitpicking: please make it fit in two lines, or even with embedded if statements, for ease of maintenance and also to respect line length (does it fit within 80 chars, per project policy).

dimitri (changes requested):
Thanks for your contribution!

Would you consider adding some coverage about your use case or even an example (with an actual snapshot and all) in the documentation too?

Issue #493, opened 2023-10-12
pgcopydb hangs on while doing `pgcopydb clone --follow`


arajkumar:
## The command used
```
pgcopydb clone --filter ~/dimitri/filter-pg.ini --follow --fail-fast --skip-extensions --plugin wal2json --trace 2>&1 --restart --drop-if-exist | tee /tmp/clone.log
```

filter-pg.ini
```
[include-only-schema]
public
```

## Nature of the ingested data

Described here: https://github.com/dimitri/pgcopydb/issues/471#issuecomment-1735009507

_TLDR: 10 txn/s with each txn having 2,000 inserts, total 20,000 inserts/s._

## Expected behaviour

Pgcopydb should continue doing CDC after initial data copying.

## Observed behaviour

Pgcopydb hangs in few mins after the start of logical decoding.

Refer to attached [clone.log.zip](https://github.com/dimitri/pgcopydb/files/12879740/clone.log.zip) for more details.

Problematic log snippet, line #868756

```
...
2023-10-12 09:17:27 476016 TRACE  ld_transform.c:787        stream_transform_file[37038]:
{"action":"I","xid":"6491903","lsn":"121/618358C8","timestamp":"2023-10-12 09:17:20.281950
+0000","message":{"action":"I","xid":6491903,"schema":"public","table":"diagnostics","colu
mns":[{"name":"time","type":"timestamp with time zone","value":"2016-01-25 19:51:00+00"},{
"name":"tags_id","type":"integer","value":2501},{"name":"fuel_state","type":"double precis
ion","value":0.1},{"name":"current_load","type":"double precision","value":1476},{"name":"
status","type":"double precision","value":1},{"name":"additional_tags","type":"jsonb","val
ue":null}]}}
2023-10-12 09:17:27 476016 TRACE  ld_transform.c:787        stream_transform_file[37039]:
{"action":"I","xid":"6491903","lsn":"121/618358C8","timestamp":"2023-10-12 09:17:20.281959
+0000","message":{"action":"I","xid":6491903,"schema":"public","table":"diagnostics","colu
mns":[{"name":"time","type":"timestamp with time zone","value":"2016-01-25 19:51:00+00"},{
"name":"tags_id","type":"integer","value":2590},{"name":"fuel_state","type":"double precis
ion","value":0.1},{"name":"current_load","type":"double precision","value":1476},{"name":"
status","type":"double precision","value":2},{"name":"additional_tags","type":"jsonb","val
[...]

Issue #471, opened 2023-09-19
Resume after interrupted `pgcopydb follow` throws "Failed to parse BEGIN: transaction already in progress"


arajkumar:
I see the following error when I resume the interrupted `pgcopydb follow`

```
07:51:48.966 2311 INFO   main.c:163                Running pgcopydb version 0.13.21.gf14255a.dirty from "/home/ubuntu/dimitri/pgcopydb/src/bin/pgcopydb/pgcopydb"
07:51:49.029 2311 ERROR  ld_transform.c:1000       Failed to parse BEGIN: transaction already in progress
07:51:49.029 2311 ERROR  ld_transform.c:827        Failed to parse JSON message: {"action":"B","xid":"3169590","lsn":"9D/757932C0","timestamp":"2023-09-19 05:04:39.159606+0000","message":{"action":"B","xid":3169590}}
```

I drilled into the json file produced by the `pgcopydb receive` process, it has two BEGIN messages for the same transaction. Here is a [trimmed json file](https://github.com/dimitri/pgcopydb/files/12657316/000000020000009D00000075.json.txt) and the problem can be reproduced using `pgcopydb stream transform dir/cdc/000000020000009D00000075.json /tmp/test.sql`

## Steps to reproduce the error

1) Start pgcopydb follow
```
pgcopydb follow --dir $PGCOPY_DIR --plugin wal2json
```
2) `killall -9 pgcopydb` and resume pgcopydb
```
pgcopydb follow --dir $PGCOPY_DIR --plugin wal2json --resume
```

arajkumar:
When I restart after a simulated crash(kill -9), sometime I see errors like below,
```
08:39:31.654 5058 INFO   Resuming streaming from latest file "/home/ubuntu/.local/share/pgcopydb/00000002000000A100000091.json.partial"
08:39:31.710 5058 ERROR  Failed to parse JSON message: {"action":"I","xid":"3274811","lsn":"A1/9119D5A8","timestamp":"2023-09-19 08:38:14.853534+0000","message":{"action":"I","xid":3274811,"schema":"_timescaledb_internal","table":"_hyper_5_458_chunk","columns":[{"name":"time","type":"timestamp with time zone","value":"2016-01-30 04:34:20+00"},{"name":"tags_id","type":"integer","value":344109},{"name":"fuel_state","type":"double precision","value":0.3},{"name":"current_load","type":"double precision","value":2798},
[...]

PR #525, opened 2023-11-01
Remove the usage of txn metadata file

arajkumar:
This commit removes the usage of transaction metadata file. Initially, it was used by the apply process to bypass transactions that were already applied. However, this approach had its challenges. Specifically, in live replay mode, a transaction with numerous statements could fill the UNIX PIPE (an IPC primitive used in replay mode), leading to a potential deadlock. This is because the apply process would be waiting for the transaction metadata file. 

By eliminating the transaction metadata file, the apply process lets the transaction proceed and decides whether to apply or skip it based on the commit LSN during the commit phase.

Fixes https://github.com/dimitri/pgcopydb/issues/493 https://github.com/dimitri/pgcopydb/issues/471

### DIFF
--- a/src/bin/pgcopydb/ld_apply.c
+++ b/src/bin/pgcopydb/ld_apply.c
@@ -650,10 +650,14 @@ stream_apply_sql(StreamApplyContext *context,
 		case STREAM_ACTION_SWITCH:
 		{
 			log_debug("SWITCH from %X/%X to %X/%X",
-					  LSN_FORMAT_ARGS(context->previousLSN),
+					  LSN_FORMAT_ARGS(context->switchLSN),
 					  LSN_FORMAT_ARGS(metadata->lsn));
 
-			context->previousLSN = metadata->lsn;
+			/*
+			 * Track the SWITCH LSN, it helps to determine the next
+			 * .sql file to apply.
+			 */
+			context->switchLSN = metadata->lsn;
 
 			break;
 		}
@@ -667,33 +671,41 @@ stream_apply_sql(StreamApplyContext *context,
 				return false;
 			}
 
+			/*
+			 * Few a time, BEGIN won't have a txnCommitLSN for the txn which
+			 * spread across multiple WAL segments. We call that txn as
+			 * a continuedTxn and allow it to be replayed until we encounter
+			 * a COMMIT message.
+			 *
+			 * The lsn of a COMMIT message determines whether to keep txn or
+			 * abort.
+			 */
+			context->continuedTxn = metadata->txnCommitLSN == InvalidXLogRecPtr;
+
 			/* did we reach the starting LSN positions now? */
 			if (!context->reachedStartPos)
 			{
 				/*
 				 * compare previousLSN with COMMIT LSN to safely include
-				 * complete transactions while skipping already applied changes.
+				 * complete transactions while skipping already applied
+				 * changes.
 				 *
-				 * this is particularly useful at the beginnig where BEGIN LSN
-				 * of some transactions could be less than `consistent_point`,
-				 * but COMMIT LSN of those transactions is guaranteed to be
-				 * greater.
+				 * this is particularly useful at the beginnig where
+				 * BEGIN LSN of some transactions could be less than
+				 * `consistent_point`, but COMMIT LSN of those transactions
+				 * is guaranteed to be greater.
 				 *
-				 * in case of interruption and this is the first transaction to
-				 * be applied, previousLSN should be equal to the last
-				 * transaction's COMMIT LSN or the LSN of non-transaction
-				 * action. Therefore, this condition will still hold true.
+				 * in case of interruption and this is the first
+				 * transaction to be applied, previousLSN should be equal
+				 * to the last transaction's COMMIT LSN or the LSN of
+				 * non-transaction action. Therefore, this condition will
+				 * still hold true.
 				 */
-
-				if (!readTxnCommitLSN(context, metadata))
-				{
-					/* errors have already been logged */
-					return false;
-				}
-
 				context->reachedStartPos =
 					context->previousLSN < metadata->txnCommitLSN;
 			}
+
+			bool skip = !context->reachedStartPos && !context->continuedTxn;
 
 			log_debug("BEGIN %lld LSN %X/%X @%s, previous LSN %X/%X, COMMIT LSN %X/%X %s",
 					  (long long) metadata->xid,
@@ -701,7 +713,7 @@ stream_apply_sql(StreamApplyContext *context,
 					  metadata->timestamp,
 					  LSN_FORMAT_ARGS(context->previousLSN),
 					  LSN_FORMAT_ARGS(metadata->txnCommitLSN),
-					  context->reachedStartPos ? "" : "[skipping]");
+					  skip ? "[skipping]" : "");
 
 			/*
 			 * Check if we reached the endpos LSN already.
@@ -719,7 +731,7 @@ stream_apply_sql(StreamApplyContext *context,
 			}
 
 			/* actually skip this one if we didn't reach start pos yet */
-			if (!context->reachedStartPos)
+			if (skip)
 			{
 				return true;
 			}
@@ -743,6 +755,7 @@ stream_apply_sql(StreamApplyContext *context,
 			 */
 			bool commitLSNreachesEndPos =
 				context->endpos != InvalidXLogRecPtr &&
+				!context->continuedTxn &&
 				context->endpos <= metadata->txnCommitLSN;
 
 			GUC *settings =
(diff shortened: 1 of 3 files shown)


arajkumar:
CC: @shubhamdhama @dimitri

arajkumar:
I see a `test cdc-endpos-beteween-transaction` fails intermittently because when resuming we no longer skip the messages(if we receive again) which are buffered into the file. The case is handled in the downstream, but when we compare the interim files, there will be an extra content some times.

> diff -I BEGIN -I COMMIT -I KEEPALIVE -I SWITCH -I ENDPOS /usr/src/pgcopydb/000000010000000000000002.sql /var/lib/postgres/.local/share/pgcopydb/000000010000000000000002.sql
9,10c9,14
< -- KEEPALIVE {"lsn":"0/244B9A8","timestamp":"2023-09-17 07:04:04.9770+0000"}
< -- ENDPOS {"lsn":"0/244B9A8"}
\---
> -- KEEPALIVE {"lsn":"0/244AD40","timestamp":"2023-11-02 09:06:05.139078+0000"}
> -- ENDPOS {"lsn":"0/244AD40"}
> BEGIN; -- {"xid":492,"lsn":"0/244AC18","timestamp":"2023-11-02 09:06:05.586617+0000"}
> PREPARE 6a9e34e7 AS INSERT INTO public.category ("category_id", "name", "last_update") overriding system value VALUES ($1, $2, $3), ($4, $5, $6), ($7, $8, $9);
> EXECUTE 6a9e34e7["1005","Mystery","2022-12-13 00:00:01+00","1006","Historical drama","2022-12-14 00:00:01+00","1008","Thriller","2022-12-15 00:00:01+00"];
> -- ENDPOS {"lsn":"0/244AD40"}

Full log from test execution: 
[cdc-endpos-between-transaction.log](https://github.com/dimitri/pgcopydb/files/13237712/cdc-endpos-between-transaction.log)

dimitri:
Hi @arajkumar  ; while I'm all for simplifying the code as much as possible, we need to make sure we don't over-simplify it. Unfortunately in this PR at the moment I think it's the case that we oversimplify some aspects. As with the previous PR you worked on I am very worried about accepting a BEGIN in the middle of a transaction or a COMMIT without a BEGIN; I think we should be very cautious about that and only accept it when *we absolutely know and made sure* it's the situation we believe it is.

Another thing to keep is mind is that setting the `endpos` can be done several times by the user:
  - set the endpos, pgcopydb finishes, 
  - realize you want to replicate more,
  - reset endpos to the new (greater) LSN, 
  - restart the follow mode

In these cases we still want correct processing of all the transactions. That's the reason why we have the xid metadata files for larger transactions (or split transactions): it allows to skip a transaction as a whole even if the target endpos LSN is to be found in the middle of said transaction. We want to stop processing but also we want to register the replay_lsn to the actual last COMMIT that was processed, so that in case we restart with a new endpos the split transaction is not skipped.

I don't see how your current implementation addresses that problem...

arajkumar:
Thanks @dimitri for your feedback.

> Hi @arajkumar ; while I'm all for simplifying the code as much as possible, we need to make sure we don't over-simplify it. Unfortunately in this PR at the moment I think it's the case that we oversimplify some aspects. As with the previous PR you worked on I am very worried about accepting a BEGIN in the middle of a transaction or a COMMIT without a BEGIN; I think we should be very cautious about that and only accept it when _we absolutely know and made sure_ it's the situation we believe it is.

Partial transactions are a reality while switching from prefetch to replay mode. BEGIN after BEGIN is very much possible when you interrupt the write before the COMMIT message arrives. But I agree, COMMIT without a BEGIN should never come. 

> 
> Another thing to keep is mind is that setting the `endpos` can be done several times by the user:
> 
> * set the endpos, pgcopydb finishes,
> * realize you want to replicate more,
> * reset endpos to the new (greater) LSN,
> * restart the follow mode
> 
> In these cases we still want correct processing of all the transactions. That's the reason why we have the xid metadata files for larger transactions (or split transactions): it allows to skip a transaction as a whole even if the target endpos LSN is to be found in the middle of said transaction. We want to stop processing but also we want to register the replay_lsn to the actual last COMMIT that was processed, so that in case we restart with a new endpos the split transaction is not skipped.
> 
> I don't see how your current implementation addresses that problem...

So, when there is no commit_lsn, we let the transaction proceed and executes the body(DML), if there is an ENDPOS in the middle, it still honours it and abort the transaction as usual. We never updates the replay_lsn for the abort case, it stays with the previous value. When you change the endpos to new value and resume the pgcopydb, it would ignore all txs except the last one which was aborted  and continue the apply from from the beginning, it will ignore the old endpos.

shubhamdhama (approved):
I have thought through this and couldn't think of scenario where this PR wouldn't work. So this looks good to me.
I have also tested it using the script by which I was able to discover previous bugs, but that worked fine here. I'll sleep on it for some more time if I find some edge case.

dimitri:
Been thinking more about it, and what I think makes this approach feasible now (when it wasn't before) has been the addition of the explicit ENDPOS message in the stream. Given that, I think I see how everything works together in a way that we don't need the transaction metadata file anymore. It's a very nice simplification, thanks @arajkumar !

I still would prefer us to keep the resume filtering logic in `ld_stream.c` module for the moment, for two reasons. First, we know we can trust that code and we don't need to remove it for this idea to work, so we can review another PR later that removes that part. Second, it allows to still issue ERROR conditions when we see a BEGIN and a transaction is already in progress.

arajkumar:
Thanks @dimitri @shubhamdhama for reviewing this PR.

> I still would prefer us to keep the resume filtering logic in `ld_stream.c` module for the moment, for two reasons. First, we know we can trust that code and we don't need to remove it for this idea to work, so we can review another PR later that removes that part. Second, it allows to still issue ERROR conditions when we see a BEGIN and a transaction is already in progress.

Let me revert the changes made to ld_stream.c.

dimitri:
I just did a quick review and the PR still allows BEGIN; BEGIN; without error, or COMMIT without BEGIN, assuming it must be the case you have in mind without making sure it is. It could be any other bug that leads to that situation and we need to protect against it.

Can we remove the new processing in these areas too?

Also I don't understand the new `context->switchLSN` tracking and when it replaces `previousLSN`, but maybe we cas dispose of it while doing the previously mentioned cleanup?

arajkumar:
> Also I don't understand the new `context->switchLSN` tracking and when it replaces `previousLSN`, but maybe we cas dispose of it while doing the previously mentioned cleanup?

@dimitri This is to handle the case where the SWITCH and COMMIT have the same LSN. When we update `previousLSN` in SWITCH, it would cause the COMMIT to be ignored as we check `previousLSN < lsn_of_commit_message`.

For example, consider the following segment files, SWITCH message in 000000040000018600000096.sql and COMMIT message in 000000040000018600000097.sql has same LSN. Without having separate field to track, apply would ignore the COMMIT because previousLSN will be same.

#### 000000040000018600000096.sql
```.sql
...
BEGIN; -- {"xid":9168338,"lsn":"186/96FF0738","timestamp":"2023-11-06 08:34:39.459848+0000"}
PREPARE 4f6bacc3 AS INSERT INTO public.diagnostics ("time", "tags_id", "fuel_state", "current_load", "status","additional_tags") overriding system value VALUES ($1, $2, $3, $4, $5, $6), ($7, $8, $9, $10, $11, $12), ($13, $14, $15, $16, $17, $18)
EXECUTE 4f6bacc3["2016-01-24 01:32:50+00","482","0.700000","234","0",null,"2016-01-2401:32:50+00","2034","0.700000","234","4",null,"2016-01-24 01:32:50+00","462","0.700000","234","3",null,"2016-01-2401:32:50+00","463","0.700000","234","0",null,"2016-01-24 01:32:50+00","1166","0.700000","234","2",null,"2016-01-2401:32:50+00","465","0.700000","234","0",null,"2016-01-24 01:32:50+00","2031","0.700000","234","2",null,"2016-01-2401:32:50+00","468","0.700000","234","0",null,"2016-01-24 01:32:50+00","469","0.700000","234","0",null,"2016-01-2401:33:00+00","331","0.700000","2515","1",null,"2016-01-24 01:33:00+00","482","0.700000","2515","0",null,"2016-01-2401:33:00+00","461","0.700000","2515","0",null,"2016-01-24 01:33:00+00","2034","0.700000","2515","3",null,"2016-01-2401:33:00+00","462","0.700000","2515","1",null,"2016-01-24 01:33:00+00","1166","0.700000","2515","3",null,"2016-01-2401:33:00+00","466","0.700000","2515",null,null,"2016-01-24 01:33:00+00","465","0.700000","2515","1",null,"2016-01-2401:33:00+00","468","0.700000","2515","0",null,"2016-01-24 01:33:00+00","469","0.700000","2515",null,null,"2016-01-2401:33:10+00","331","0.700000","1736","1"]
-- SWITCH WAL {"lsn":"186/970098C0"}
```

#### 000000040000018600000097.sql
```.sql
COMMIT; -- {"xid":9168338,"lsn":"186/970098C0","timestamp":"2023-11-06 08:34:39.466938+0000"}
...
```

dimitri:
> @dimitri This is to handle the case where the SWITCH and COMMIT have the same LSN. When we update `previousLSN` in SWITCH, it would cause the COMMIT to be ignored as we check `previousLSN < lsn_of_commit_message`.

Oh that's a good find! It warrants some comments in the code to explain that, and I think it would be even best as its own separate bugfix PR, what do you think?

arajkumar:
> > @dimitri This is to handle the case where the SWITCH and COMMIT have the same LSN. When we update `previousLSN` in SWITCH, it would cause the COMMIT to be ignored as we check `previousLSN < lsn_of_commit_message`.
> 
> Oh that's a good find! It warrants some comments in the code to explain that, and I think it would be even best as its own separate bugfix PR, what do you think?

@dimitri  It wasn't a problem earlier as we checked the commitLSN in the BEGIN message., this change is only needed for the new implementation which doesn't require txn metadata file. I will try to add some comment.

dimitri (approved):

dimitri:
Thanks for all the work @arajkumar !

arajkumar:
Thanks a lot @dimitri for your time and feedback.

Next, I'm planning to address https://github.com/dimitri/pgcopydb/issues/471.

PR #528, opened 2023-11-06
Improve and fix connection timeout and total retry timeout.

shubhamdhama:
Previously, when a connection timeout occurred with the default connection timeout set to pgconnect_timeout (10 seconds), there were no retries. This was because the total retry timeout (maxT) passed to `pgsql_set_retry_policy` in `pgsql_set_interactive_retry_policy` was also set to pgconnect_timeout, which had already elapsed during the initial connection attempt.

To address this issue, the maxT value has been updated to POSTGRES_PING_RETRY_TIMEOUT. This change also means that maxT cannot be configured by altering PGCONNECT_TIMEOUT. If one wants to configure maxT they can change the values in defaults.h or we can make it configurable through some environment variable in future.

After all this change `pgconnect_timeout` seems redundant so removed it.

### DIFF
--- a/src/bin/pgcopydb/cli_root.h
+++ b/src/bin/pgcopydb/cli_root.h
@@ -12,7 +12,6 @@
 
 extern char pgcopydb_argv0[];
 extern char pgcopydb_program[];
-extern int pgconnect_timeout;
 extern int logLevel;
 
 extern Semaphore log_semaphore;
--- a/src/bin/pgcopydb/defaults.h
+++ b/src/bin/pgcopydb/defaults.h
@@ -50,9 +50,8 @@
 
 #define POSTGRES_CONNECT_TIMEOUT "10"
 
-
-/* retry PQping for a maximum of 15 mins, up to 2 secs between attemps */
-#define POSTGRES_PING_RETRY_TIMEOUT 900               /* seconds */
+/* retry PQping for a maximum of 1 min, up to 2 secs between attemps */
+#define POSTGRES_PING_RETRY_TIMEOUT 60               /* seconds */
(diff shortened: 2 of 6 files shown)


dimitri (changes requested):
LGTM.

Can we also now document PGCONNECT_TIMEOUT in our “Environment” section? My understanding is that the user defined value with be honored at face value for pg_dumpall, pg_dump and pg_restore calls, but used in a different way for the other connections?

shubhamdhama:
@dimitri 

> My understanding is that the user defined value with be honored at face value for pg_dumpall, pg_dump and pg_restore calls

Yes
>  but used in a different way for the other connections?

Even pgcopydb's connection would honor this value. Even though setting environment variable for the same process is undefined but it seems to be working good (maybe on linux at-least). Ideally we should use the `connect_timeout` parameter, but then it would override user defined environment variable value.


IMO we don't need to document this enviornment variable because it works just like any other variable present here https://www.postgresql.org/docs/current/libpq-envars.html. The only difference is we have our own default of 10 seconds since libpq doesn't mentions its https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CONNECT-TIMEOUT.

It's just my opinion, I'm happy to document it if it would help the users. What are your thoughts?

dimitri (approved):

dimitri:
> IMO we don't need to document this enviornment variable because it works just like any other variable present here https://www.postgresql.org/docs/current/libpq-envars.html. The only difference is we have our own default of 10 seconds since libpq doesn't mentions its https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CONNECT-TIMEOUT.
> 
> It's just my opinion, I'm happy to document it if it would help the users. What are your thoughts?

I see your point. I can definitely live with that ;-)

PR #533, opened 2023-11-08
Improve migration speed by using same connection in a LOB worker.

shubhamdhama:
Presently, each LOB worker creates a new connection for every object copy, impacting migration performance, especially with a large number of objects. The fix here is simple: use the same connection within each worker.

However, while doing so, we encountered an unexpected error:

`BUG: call to pgsql_commit() without holding an open multi statement connection`

Upon investigation, we discovered that the `connectionStatementType` was set to an unusual value of 32767. In GDB, it was
(PGSQL_CONNECTION_MULTI_STATEMENT | unknown: 32766). This happens only after the call of `msgrcv` function. There are a few theories; one suggests that because both `msg` and `dst` are on the stack, `msgrcv` is overwriting `dst` data when it should only change `msg`. Despite having `-fstack-protector` enabled, we did not observe a "stack smashing detected" error, so maybe this theory is incorrect.

In any case, moving `msg` to the heap seemed to resolve the issue. However, it's important to fully understand this problem, as it may have implications in other areas of the code as well.

Results for this schema
```sql
BEGIN;

CREATE TABLE files(
    id bigserial PRIMARY KEY,
    name text,
    file oid
);

\set n 100000
INSERT INTO files(name, file)
SELECT
    array_to_string(ARRAY (
            SELECT
                chr((65 + round(random() * 25))::integer)
    FROM generate_series(1,(random() * 10)::int)), ''),
    lo_from_bytea(0, '\xffffff00')
FROM
    generate_series(1, :n);

COMMIT;
```
Before
```
                         Large Objects (cumulative)         both       4m31s            10
```

After
```
                         Large Objects (cumulative)         both       1m17s            10
```


src/bin/pgcopydb/blobs.c
```diff
@@ -253,44 +253,73 @@ copydb_blob_worker(CopyDataSpec *specs)
...
+	}
+
 	int errors = 0;
 	bool stop = false;
 
 	while (!stop)
 	{
-		QMessage mesg = { 0 };
-		bool recv_ok = queue_receive(&(specs->loQueue), &mesg);
+		QMessage *mesg = malloc(sizeof(QMessage));
```
dimitri, line 279: Let's use `calloc` to ensure that we initialize that memory area with zeroes.

dimitri (changes requested):
Thanks for that work. I think we need to make another PR where we fix all the `queue_receive` calls to use the heap instead of the stack. We might even make yet another PR later that does the same thing with all the PGSQL instances.

dimitri (approved):

PR #542, opened 2023-11-16
Use same connection in table copy worker.

shubhamdhama:
Currently for each table we are creating and closing a new connection to
the target. Using a same connection would save us some time and resource,
both at pgcopydb and target side.

Also, seting GUCs only once for a given target connection session would
again save us resources. This would also reduce 8 extra log lines that are
currently produced for each table which otherwise would produce noise if we
have say, 10K tables.

Follow-up: Use the same connection for `INDEX` worker where same connection is even more useful.

### DIFF
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -52,7 +52,9 @@ static bool build_parameters_list(PQExpBuffer buffer,
 
 static void parseIdentifySystemResult(void *ctx, PGresult *result);
 static void parseTimelineHistoryResult(void *ctx, PGresult *result);
-
+static bool pg_copy_data(PGSQL *src, PGSQL *dst, const char *srcQname,
+						 const char *dstQname, bool truncate,
+						 uint64_t *bytesTransmitted);
 static bool pg_copy_send_query(PGSQL *pgsql,
 							   const char *qname,
 							   ExecStatusType status,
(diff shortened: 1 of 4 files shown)


shubhamdhama:
It's a follow-up of #541 I'll remove https://github.com/dimitri/pgcopydb/pull/542/commits/07c8b6604838ac2898a5ac8967ab3db2df7694f9 commit once that PR is merged.

src/bin/pgcopydb/pgsql.c
```diff
@@ -2273,13 +2270,37 @@ pg_copy(PGSQL *src, PGSQL *dst, const char *srcQname, const char *dstQname,
...
+
+/*
+ * pg_copy implements a COPY operation from a source Postgres instance (src) to
+ * a target Postgres instance (dst), for the data found in the table referenced
+ * by the qualified identifier name srcQname on the source, into the table
+ * referenced by the qualified identifier name dstQname on the target.
+ */
+bool
+pg_copy_data(PGSQL *src, PGSQL *dst, const char *srcQname, const char *dstQname,
+			 bool truncate)
```
shubhamdhama, line 2297: Here I have extracted out the function from `pg_copy` to make the setting up and clean-up of connection from failure simpler. 
It's sort of a private function but since C doesn't have such construct is there any convention to show it's not meant for other files to use.

dimitri: Make it static, you will find other functions in there that are static. Then it needs its prototype in the beginning of the same .c file.

src/bin/pgcopydb/table-data.c
```diff
@@ -484,6 +484,24 @@ copydb_table_data_worker(CopyDataSpec *specs)
 		return false;
 	}
 
+	PGSQL *src = &(specs->sourceSnapshot.pgsql);
+	PGSQL dst = { 0 };
```
dimitri, line 488: Can we use `calloc` to allocation the `dst` PGSQL structure instance?

src/bin/pgcopydb/table-data.c
```diff
@@ -558,6 +581,11 @@ copydb_table_data_worker(CopyDataSpec *specs)
 	/* terminate our connection to the source database now */
 	(void) copydb_close_snapshot(specs);
 
+	if (errors > 0)
+	{
+		pgsql_finish(&dst);
+	}
+
```
dimitri, line 588: I think we should call `pgsql_finish` unconditionally here.

dimitri (changes requested):
Small changes before we can merge. Thanks!

src/bin/pgcopydb/table-data.c
dimitri: Let's revisit that idea in another issue/PR.

PR #544, opened 2023-11-17
Rollback incomplete txns on graceful exit

arajkumar:
Before this update, exiting the process (either reaching ENDPOS or user-initiated abort like Ctrl-C) could leave us in the midst of an unfinished transaction. Upon resumption, PostgreSQL begins streaming from a consistent point, which might be the start of this incomplete transaction or from an earlier one. Our previous method tried to bypass all earlier messages until hitting the final message of the incomplete transaction, but this approach had its shortcomings, especially when the messages originated from several transactions back.

This commit introduces a more straightforward solution. Now, if the process is aborted and the last transaction is not complete, we'll issue a rollback command for that transaction. This ensures a clean and consistent state for a graceful exit and subsequent resumption.

### DIFF
--- a/src/bin/pgcopydb/ld_stream.c
+++ b/src/bin/pgcopydb/ld_stream.c
@@ -338,14 +338,12 @@ stream_init_context(StreamSpecs *specs)
 
 	privateContext->endpos = specs->endpos;
 	privateContext->startpos = specs->startpos;
-	privateContext->startposActionFromJSON = specs->startposActionFromJSON;
 
 	privateContext->mode = specs->mode;
 
 	privateContext->transformQueue = &(specs->transformQueue);
 
 	privateContext->paths = specs->paths;
-	privateContext->startpos = specs->startpos;
 
 	privateContext->connStrings = specs->connStrings;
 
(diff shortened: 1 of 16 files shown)


dimitri (commented):
I think given all your tries of solving it a different way, without adding a ROLLBACK to the current half-transaction on-disk, this is the most promising approach. I'm surprised that the test-case does not embed a ROLLBACK message in the JSON or the SQL files, do you think it would be possible to make that happen?

src/bin/pgcopydb/ld_stream.c
```diff
@@ -896,6 +849,15 @@ stream_write_json(LogicalStreamContext *context, bool previous)
...
 
+	if (metadata->action == STREAM_ACTION_BEGIN)
+	{
+		privateContext->transactionInProgress = true;
+	}
+	else if (metadata->action == STREAM_ACTION_COMMIT)
+	{
+		privateContext->transactionInProgress = false;
+	}
+
```
dimitri, line 860: Do we need to add a case for STREAM_ACTION_ROLLBACK? if not, we need to add a comment about why...

src/bin/pgcopydb/ld_stream.c
```diff
@@ -1237,6 +1199,28 @@ streamCloseFile(LogicalStreamContext *context, bool time_to_abort)
 		}
 	}
 
+	/*
+	 * ROLLBACK transactions that are still in progress when we abort.
+	 * When resuming, we always start from a consistent point, so we don't need
+	 * worry about partial transactions which are rollbacked here.
+	 */
```
dimitri, line 1206: What happens in case of a crash? I think we need a short reference to that in the comment (the code to handle restart after crash might not be here, but rather where we read a JSON file from disk and resume from it).

dimitri (changes requested):

dimitri (approved):

arajkumar:
Fixes https://github.com/dimitri/pgcopydb/issues/471

PR #549, opened 2023-11-21
Recover from crash using undo log

arajkumar:
This commit remembers the file write position of a last successful transaction in
a undo file. If there's a crash and we need to restart, we use this undo file to get rid of any partial work from the last transaction that wasn't finished.

TODO: Can we use the undo log mechanism to even for graceful exit?
This would simplify the code and make it uniform for both graceful
and ungraceful exits/crashes. The idea is to undo the changes made
by the last incomplete transaction by reading the undo log file and
write other messages like ENDPOS.

### DIFF
--- a/src/bin/pgcopydb/copydb.c
+++ b/src/bin/pgcopydb/copydb.c
@@ -572,6 +572,10 @@ copydb_prepare_filepaths(CopyFilePaths *cfPaths,
 			"%s/lsn.json",
 			cfPaths->cdc.dir);
 
+	sformat(cfPaths->cdc.undofile, MAXPGPATH,
+			"%s/undo",
+			cfPaths->cdc.dir);
+
 	/*
 	 * Now prepare the "compare" files we need to compare schema and data
 	 * between the source and target instance.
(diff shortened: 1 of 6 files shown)


arajkumar:
I think implementing a proper undo log will become tricky to unless we resort to a transactional file write. Maybe opening the latest file and appending a ROLLBACK(if last txn is incomplete) will be much easier to implement.

dimitri:
> I think implementing a proper undo log will become tricky to unless we resort to a transactional file write. Maybe opening the latest file and appending a ROLLBACK(if last txn is incomplete) will be much easier to implement.

What if the latest/current transaction is incomplete and started several JSON files ago?

arajkumar:
> > I think implementing a proper undo log will become tricky to unless we resort to a transactional file write. Maybe opening the latest file and appending a ROLLBACK(if last txn is incomplete) will be much easier to implement.
> 
> What if the latest/current transaction is incomplete and started several JSON files ago?

It doesn't matter, all we do is appending a rollback message to latest file if we find the last statement is part of the incomplete txn. It is same as what we do for graceful exit.

PR #559, opened 2023-11-30
Add restart support to snapshot

arajkumar:
This will drop the existing replication slot and recreates a new one.

### DIFF
--- a/src/bin/pgcopydb/cli_snapshot.c
+++ b/src/bin/pgcopydb/cli_snapshot.c
@@ -28,12 +28,12 @@ CommandLine snapshot_command =
 	make_command(
 		"snapshot",
 		"Create and export a snapshot on the source database",
-		" --source ... ",
 		"  --source         Postgres URI to the source database\n"
 		"  --dir            Work directory to use\n"
 		"  --follow         Implement logical decoding to replay changes\n"
 		"  --plugin         Output plugin to use (test_decoding, wal2json)\n"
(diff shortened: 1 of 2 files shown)


arajkumar:
I'm not sure the default behavior of `pgcopydb snapshot --follow` is useful. When we run second time, it skips creating replication slot which is fine, but prints the snapshot-id which was closed already. 

Should we keep the default behavior of `pgcopydb snapshot` as always restart?

dimitri (changes requested):
Good idea, some details need attention.

src/bin/pgcopydb/cli_snapshot.c
```diff
@@ -290,6 +308,23 @@ cli_create_snapshot(int argc, char **argv)
...
 
+		/*
+		 * When using pgcopydb snapshot --follow --restart we first cleanup the
+		 * previous setup, and that includes dropping the replication slot.
+		 */
+		if (createSNoptions.restart)
+		{
+			log_info("Clean-up replication setup, per --restart");
+
+			if (!stream_cleanup_databases(&copySpecs,
```
arajkumar, line 319: @dimitri I see this function call also drops replication origin on the target., the snapshot never dealt with the target so far right? I'm not sure we should clean up the origin & sentinel tables as part of this.? 

I think this option adds a bit of confusion around the responsibilities of the subcommand(snapshot, stream cleanup).

Issue #539, opened 2023-11-12
Can foreign key creation be parallelized?


troelsarvin:
I'm looking at a clone situation where things have gone rather fast so far, but suddenly things seem to be proceeding in a single-task-at-a-time:

```
12:55:06 999542 INFO   STEP 10: restore the post-data section to the target database
12:55:06 999542 INFO    /usr/bin/pg_restore --dbname postgres:///mydb --single-transaction --use-list /tmp/pgcopydb/schema/post.list /tmp/pgcopydb/schema/post.dump
```

When I look at what's going on in the database, there is a single ALTER TABLE statement running:

`ALTER TABLE ONLY schemaname.table1?    ADD CONSTRAINT table1_othercolid_fkey FOREIGN KEY (somecol) REFERENCES schemaname.othertable(id) ON UPDATE RESTRICT ON DELETE RESTRICT;???`

Maybe things would be faster, if the pg_restore command be given a "-j X" argument? X could perhaps be equal to PGCOPYDB_INDEX_JOBS (or a PGCOPYDB_POST_JOBS parameter could be introduced). For this to have an effect, the "--single-transaction" argument would have to be removed; that would be perfectly fine for most of my pgcopydb use cases, but maybe not for everyone. So maybe it would require an extra optional argument allowing not to use --single-transaction.

PR #561, opened 2023-11-30
Parallelize pg_restore operations

hanefi:
This commit adds a new option --restore-jobs to pgcopydb that allows specifying how many jobs can be used to run pg_restore operations in parallel. This option can also be set using the PGCOPYDB_RESTORE_JOBS environment variable.

When this option is set to 1, pgcopydb will run pg_restore with the --single-transaction option, and pgcopydb will behave the same as it used to. Otherwise, pg_restore will be run with the --jobs option set to the number of jobs specified by the user.

Fixes: #539

TODO:
- [ ] Discuss a sane default with team. Should we keep using `pg_restore --single-transaction` as default instead?
- [ ] Go over the list of steps in `pgcopydb_clone.rst` and make sure the changes are correct.


hanefi:
> I think a good default value for `--restore-jobs` should be the current value for `--index-jobs` (dynamic, from command line or environment) , because this impacts the target server only.

Agreed. This was not a trivial change as some `pgcopydb restore` commands did not have `--index-jobs` parameter, but we need to set a default value for `--restore-jobs`. I ended up checking for the PGCOPYDB_INDEX_JOBS, or `--index-jobs` and if both failed I used the default value for index jobs which is 4. I think this is acceptable.

> Also I'm not sure we need `--single-transaction` on the target server at all, actually, but I think I like the way to chose to still have it when `--restore-jobs` is 1.

I think we should keep a workaround for users that desire to keep the old behavior where we always used `--single-transaction`. This may not be really needed, but it is nice to let them have that option.

src/bin/pgcopydb/defaults.h
```diff
@@ -43,6 +44,7 @@
 /* default values for the command line options */
 #define DEFAULT_TABLE_JOBS 4
 #define DEFAULT_INDEX_JOBS 4
+#define DEFAULT_RESTORE_JOBS 0
```
hanefi, line 47: As discussed changed the default to 0 from -1.

docs/ref/pgcopydb_clone.rst
```diff
@@ -423,6 +430,12 @@ The following options are available to ``pgcopydb clone``:
   Postgres target system, minus some cores that are going to be used for
   handling the COPY operations.
 
+--restore-jobs
+
+  How many threads or processes can be used during pg_restore. A good is to set
+  this option to the number of threads that are available on the Postgres target
+  system.
```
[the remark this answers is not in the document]
dimitri: Also add text to mention that `--index-jobs` is used as the default when it's been provided and `--restore-jobs` has been omitted. Also please mention that the option is passed down as-is to `pg_restore --jobs` option.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -819,7 +819,16 @@ pg_restore_db(PostgresPaths *pgPaths,
 	args[argsIndex++] = (char *) pgPaths->pg_restore;
 	args[argsIndex++] = "--dbname";
 	args[argsIndex++] = (char *) connStrings->safeTargetPGURI.pguri;
-	args[argsIndex++] = "--single-transaction";
+
+	if(options.jobs == 1)
```
dimitri, line 823: Nitpicking: space between `if` and the open paren.

dimitri (changes requested):

docs/ref/pgcopydb_clone.rst
```diff
@@ -432,9 +432,12 @@ The following options are available to ``pgcopydb clone``:
 
 --restore-jobs
 
-  How many threads or processes can be used during pg_restore. A good is to set
-  this option to the number of threads that are available on the Postgres target
-  system.
+  How many threads or processes can be used during pg_restore. A good option is
+  to set this option to the countq of CPU cores that are available on the
```
dimitri, line 436: s/countq/count/

docs/ref/pgcopydb_restore.rst
```diff
@@ -210,6 +210,15 @@ The following options are available to ``pgcopydb restore schema``:
   ``${TMPDIR}/pgcopydb`` when the environment variable is set, or
   then to ``/tmp/pgcopydb``.
 
+--restore-jobs
+
+  How many threads or processes can be used during pg_restore. A good option is
+  to set this option to the countq of CPU cores that are available on the
```
dimitri, line 216: s/countq/count/

dimitri (changes requested):

docs/ref/pgcopydb_clone.rst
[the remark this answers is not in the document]
hanefi: Updated this section and added another paragraph:

    If this value is not set, we reuse the ``--index-jobs`` value. If that value
    is not set either, we use the the default value for ``--index-jobs``.

src/bin/pgcopydb/pgcmd.c
hanefi: Fixed.

I also went ahead and fix all the current indentation issues so that I can use citus_indent for formatting in this project. See https://github.com/dimitri/pgcopydb/pull/567

PR #565, opened 2023-12-04
Automate updating help messages for commands

hanefi:
This commit is the first phase of a plan to automate updating our documentation. It adds a script that can be used to update the help messages for commands. Instead of copy pasting `--help` output into the documentation, we can now use this script to generate these messages for us. All that is left is to include relevant parts of these generated output in the documentation.

In the near future, I plan to run this script as part of the build process to make sure that the documentation is always up to date.

Closes: #588

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ maintainer-clean: clean
 docs:
 	$(MAKE) -C docs clean man html
 
+update-docs: install
+	bash ./docs/update-help-messages.sh
+
 test: build
 	$(MAKE) -C tests all
 
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -65,3 +65,13 @@ jobs:
 
     - name: Check C style
       run: citus_indent --check
+
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Build docs
+        uses: ammaraskar/sphinx-action@master
+        with:
+          docs-folder: "docs/"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,7 +58,7 @@ html_theme = "sphinx_rtd_theme"
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = []
 
 latex_engine = 'xelatex'
 
--- a/docs/include/clone.rst
+++ b/docs/include/clone.rst
@@ -1,0 +1,39 @@
+::
+
+   pgcopydb clone: Clone an entire database from source to target
+   usage: pgcopydb clone  --source ... --target ... [ --table-jobs ... --index-jobs ... ] 
+   
+     --source                      Postgres URI to the source database
+     --target                      Postgres URI to the target database
+     --dir                         Work directory to use
+     --table-jobs                  Number of concurrent COPY jobs to run
+     --index-jobs                  Number of concurrent CREATE INDEX jobs to run
+     --restore-jobs                Number of concurrent jobs for pg_restore
+     --large-objects-jobs          Number of concurrent Large Objects jobs to run
+     --split-tables-larger-than    Same-table concurrency size threshold
+     --drop-if-exists              On the target database, clean-up from a previous run first
+     --roles                       Also copy roles found on source to target
+     --no-role-passwords           Do not dump passwords for roles
+     --no-owner                    Do not set ownership of objects to match the original database
+     --no-acl                      Prevent restoration of access privileges (grant/revoke commands).
+     --no-comments                 Do not output commands to restore comments
+     --skip-large-objects          Skip copying large objects (blobs)
+     --skip-extensions             Skip restoring extensions
+     --skip-ext-comments           Skip restoring COMMENT ON EXTENSION
+     --skip-collations             Skip restoring collations
+     --skip-vacuum                 Skip running VACUUM ANALYZE
+     --requirements <filename>     List extensions requirements
+     --filters <filename>          Use the filters defined in <filename>
+     --fail-fast                   Abort early in case of error
+     --restart                     Allow restarting when temp files exist already
+     --resume                      Allow resuming operations after a failure
+     --not-consistent              Allow taking a new snapshot on the source database
+     --snapshot                    Use snapshot obtained with pg_export_snapshot
+     --follow                      Implement logical decoding to replay changes
+     --plugin                      Output plugin to use (test_decoding, wal2json)
+     --wal2json-numeric-as-string  Print numeric data type as string when using wal2json output plugin
+     --slot-name                   Use this Postgres replication slot name
+     --create-slot                 Create the replication slot
+     --origin                      Use this Postgres replication origin node name
+     --endpos                      Stop replaying changes when reaching this LSN
+   
--- a/docs/include/compare-data.rst
+++ b/docs/include/compare-data.rst
@@ -1,0 +1,10 @@
+::
+
+   pgcopydb compare data: Compare source and target data
+   usage: pgcopydb compare data  --source ... 
+   
+     --source         Postgres URI to the source database
+     --target         Postgres URI to the target database
+     --dir            Work directory to use
+     --json           Format the output using JSON
+   
--- a/docs/include/compare-schema.rst
+++ b/docs/include/compare-schema.rst
@@ -1,0 +1,9 @@
+::
+
+   pgcopydb compare schema: Compare source and target schema
+   usage: pgcopydb compare schema  --source ... 
+   
+     --source         Postgres URI to the source database
+     --target         Postgres URI to the target database
+     --dir            Work directory to use
+   
--- a/docs/include/compare.rst
+++ b/docs/include/compare.rst
@@ -1,0 +1,9 @@
+::
+
+   pgcopydb compare: Compare source and target databases
+   
+   Available commands:
+     pgcopydb compare
+       schema  Compare source and target schema
+       data    Compare source and target data
+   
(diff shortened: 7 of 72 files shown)


hanefi:
Several improvement areas:
- [ ] Remove whitespaces from file names. Consider replacing them with underscores instead.
- [ ] Automate generating all available subcommands. 
- [ ] Remove unused commands or add them to the documentation somehow.

Current list of generated help outputs that are unused:
- copy.rst
- copy-db.rst
- fork.rst
- stream.rst
- stream sentinel.rst
- stream sentinel set.rst

hanefi:
It is not so trivial to add a CI check for this script for several reasons:

- We can not build the project outside a docker environment as we are missing package dependencies
- We can not easily update an existing test to run the script and verify all the docs are current as current test containers do not contain docs or necessary makefiles and shell scripts.

I will postpone adding CI capabilities to a later time.

gokhangulbiz:
Great work, @hanefi! The script to automate help message updates is a game-changer for our documentation. This will make our lives much easier and ensure docs are always in sync.

/cc @dimitri

dimitri:
Can we add a new CI workflow that would only build the docs? That would be a good starting point for any docs contents automation I believe...

hanefi:
I cleaned up and removed all the new warnings that I introduced in this branch. The only warning left is about static files.

    copying static files... WARNING: html_static_path entry '/home/hanefi/code/pgcopydb/docs/_static' does not exist

We can easily fix that warning as well by the following patch:

```diff
diff --git a/docs/conf.py b/docs/conf.py
index e020e9a..ea35f8c 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,7 +58,7 @@ html_theme = "sphinx_rtd_theme"
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = []

 latex_engine = 'xelatex'
```

I am fairly certain that this is a boilerplate code that is not really applicable to our project as we do not have that `_static` file. Just to be safe, I did not apply this change in this PR as I feared I might brake readthedocs builds.

dimitri:
> I am fairly certain that this is a boilerplate code that is not really applicable to our project as we do not have that `_static` file. Just to be safe, I did not apply this change in this PR as I feared I might brake readthedocs builds.

Please add the change in here, I can arrange RTD to produce docs for a branch too. We will see what happens on that branch build.

update-help-messages.sh
```diff
@@ -0,0 +1,61 @@
...
+        "clone" "follow" "copy" "snapshot" "compare" "dump" "restore" "list" "stream" "ping" \
+        "compare schema" "compare data" \
+        "copy db" "copy roles" "copy extensions" "copy schema" "copy data" "copy table-data" "copy blobs" "copy sequences" "copy indexes" "copy constraints" \
+        "dump schema" "dump pre-data" "dump post-data" "dump roles" \
+        "list databases" "list extensions" "list collations" "list tables" "list table-parts" "list sequences" "list indexes" "list depends" "list schema" "list progress" \
+        "restore schema" "restore pre-data" "restore post-data" "restore roles" "restore parse-list" \
+        "stream setup" "stream cleanup" "stream prefetch" "stream catchup" "stream replay" "stream sentinel" "stream receive" "stream transform" "stream apply" \
+        "stream sentinel create" "stream sentinel drop" "stream sentinel get" "stream sentinel set" \
+        "stream sentinel set startpos" "stream sentinel set endpos" "stream sentinel set apply" "stream sentinel set prefetch" \
+    )
```
dimitri, line 48: Is there a way to generate that list from the `pgcopydb help` command?

update-help-messages.sh
```diff
@@ -0,0 +1,61 @@
...
+
+    # Print the main pgcopydb help output to pgcopydb.rst
+    print_help "" >"$INCLUDE_DIR/pgcopydb.rst"
+
+    # Print the help output for each command to their separate file
+    for cmd in "${cmds[@]}"; do
+        print_help "$cmd" >"$INCLUDE_DIR/$cmd.rst"
+    done
```
dimitri, line 56: It would be nice to replace the spaces by another separation in the filename. Traditionally that would probably be an underscore (`_`), but in our case maybe a dash (`-`) would look better overall, given that some commands use the dash already?

update-help-messages.sh
```diff
@@ -10,10 +10,18 @@ INCLUDE_DIR="docs/include"
...
+    local file_path
+    if [ "$cmd" == "" ]; then
+        file_path="$INCLUDE_DIR/pgcopydb.rst"
+    else
+        file_path="$INCLUDE_DIR/${cmd// /-}.rst"
+    fi
 
+    local help_text
     # shellcheck disable=SC2086
     help_text="$(pgcopydb $cmd --help 2>&1)"
```
dimitri, line 26: Can you double-quote `"${cmd}"` and remove the shellcheck disable comment? Also please use `${name}` notation, makes it easier to review/maintain.

update-help-messages.sh
```diff
@@ -25,37 +33,45 @@ function print_help() {
...
+            continue;
+        # Parse subcommands
+        elif [[ $l =~ ^([a-z-]+) || $l =~ ^\+\ ([a-z-]+) ]]; then
+            subcmd="${BASH_REMATCH[1]}"
+        else
+            continue
+        fi
+
+        # shellcheck disable=SC2086
+        print_help_to_file $cmd $subcmd
```
dimitri, line 67: Again, can we use `print_help_to_file "${cmd} ${subcmd}" and remove the comment to disable a shellcheck?

update-help-messages.sh
```diff
@@ -25,37 +33,45 @@ function print_help() {
...
+        if [[ $l =~ ^pgcopydb$ ]]; then
+            cmd=""
+            subcmd=""
+        # Parse other sections of the help text that contain `pgcopydb <cmd>`
+        elif [[ $l =~ ^pgcopydb\ (.+) ]]; then
+            cmd="${BASH_REMATCH[1]}"
+            subcmd=""
+            continue;
+        # Parse subcommands
+        elif [[ $l =~ ^([a-z-]+) || $l =~ ^\+\ ([a-z-]+) ]]; then
```
dimitri, line 60: I think these regexps needs comments to explain what you expect from them, and why you had to separate them in two rounds of processing instead of one. Usually I would put in the main comment for the first `if` a list of examples of lines that are parsed and what we get from the regexps etc, in a pretty visual way.

Again, it's just a shell script, but we have to maintain it as soon as it's merged.

dimitri (changes requested):
Please see comments. Almost there, I like the overall approach, but we still need to make that shell script easier to maintain in the long-run.

update-help-messages.sh
hanefi: I just realized that having a dash to replace space is not acceptable here. For example, the following are both acceptable commands that produce different results:

- `pgcopydb copy-db --help`
- `pgcopydb copy db --help`

I will use underscores instead.

- `pgcopydb copy-db --help` output will be in `copy-db.rst`
- `pgcopydb copy db --help` output will be in `copy_db.rst`

dimitri: It's just that using both underscore and dash in the same file name looks ugly to me. Maybe we could double the pre-existing dashes and have `pgcopydb-copy--db.rst` and `pgcopydb-copy-db.rst` for that example? Or better yet maybe it's time to remove the `pgcopydb copy-db` command now that we have `pgcopydb clone` and its alias `pgcopydb fork` too.

update-help-messages.sh
```diff
@@ -0,0 +1,129 @@
...
+        help_cmd="pgcopydb ${cmd} --help 2>&1"
+        file_path="${INCLUDE_DIR}/${cmd// /_}.rst"
+    fi
+
+    # Generate help text by running the command, removing the line with version
+    # information and adding 3 spaces at the beginning of each line
+    local help_text
+    help_text="$( eval "${help_cmd}"  |
+        sed '/.*Running pgcopydb version.*/d' |
+        sed 's/^/   /'
```
dimitri, line 39: IIRC you can use multiple sed expressions in a single sed invocation, using `sed -e 's///' -e 's///'`.

update-help-messages.sh
```diff
@@ -0,0 +1,129 @@
...
+    # `pgcopydb compare data --help` to `compare-data.rst`
+    local file_path
+    local help_cmd
+    if [ "${cmd}" = "" ]; then
+        help_cmd="pgcopydb --help 2>&1"
+        file_path="${INCLUDE_DIR}/pgcopydb.rst"
+    else
+        help_cmd="pgcopydb ${cmd} --help 2>&1"
+        file_path="${INCLUDE_DIR}/${cmd// /_}.rst"
+    fi
```
dimitri, line 32: Can we transform the existing dashes and double them, and then replace the space with a dash again?

dimitri (changes requested):
Last change before approve and merge: can we move the `update-help-messages.sh` script from the top-level directory to either the `ci` subdir or maybe the `docs` subdir?

Makefile
```diff
@@ -21,6 +21,9 @@ maintainer-clean: clean
 docs:
 	$(MAKE) -C docs clean man html
 
+update-docs:
+	bash ./docs/update-help-messages.sh
+
```
dimitri, line 26: Do I read the script correctly that this target depends on the `install` target?

hanefi: Makes sense. I added the dependency to build target.

dimitri: Another way to do it would be to use a PGCOPYDB variable that points to the binary we're using, and would default to `./src/bin/pgcopydb/pgcopydb` so that the default dependency would be to `make bin` rather than `make install`, or could be completely bypassed with `make PGCOPYDB=/usr/local/bin/pgcopydb` docs` for instance.

That said in any case the pgcopydb binary built from the current repository state needs to be available when running the `update-help-messages.sh` script. Also, I'm not sure that the CI system needs to be able to run that command, as the computed files are checked-in to the repository?

dimitri (approved):

Makefile
hanefi: I think the CI system should raise an error when the developer updates a command interface and forgets to update the docs. The purpose of said test should not be updating files, but letting the developer know that there is something that needs fixing.

PR #567, opened 2023-12-06
Run make indent target

hanefi:
This commit will allow developers to run `make indent` to indent their code in their development branches by solving all the indentation issues in the codebase.


arajkumar (approved):
lgtm, I was about to create a PR with this :D Thank you.

@hanefi Can we also add the check into CI, so that it can fail on missing indentation.?

Basically we need to add another job into the [workflow](https://github.com/dimitri/pgcopydb/blob/295f2d1710a1b8713e490253f0546903c494abfd/.github/workflows/run-tests.yml#L13), which installs citus_indent and runs `make indent` followed by `git diff --exit-code` to throw error if there is a missing indent?

dimitri:
There is already `make tests/ci` for this kind of things, but I suppose it's lacking the indentation check at the moment. I think it's worth adding it there within this PR.

hanefi:
I think that it does not really makes sense to clone another git repo, build and install. Instead it makes more sense to reuse the docker images that already have citus_indent set up.

I added another CI job that uses that docker image, and runs the checks on CI where I used `citus_indent --check` that exits with an error code if there are some problems with the indentation.

hanefi:
FYI, the Dockerfile that we use lives in https://github.com/citusdata/the-process/blob/master/circleci/images/stylechecker/Dockerfile

It is not an image that is updated frequently.

dimitri (approved):

dimitri:
Nice, I didn't know we could borrow the other infrastructure from Citus that easily. Thanks @hanefi!

PR #568, opened 2023-12-06
Compare endpos only if it is valid

arajkumar:
Endpos can be set to 0/0 which is an invalid LSN.


dimitri (approved):

dimitri:
> Endpos can be set to 0/0 which is an invalid LSN.

Is there a case to be made for better parsing of endpos in our command lines? (we would still need this PR, so it's going to be merged in any case).

PR #569, opened 2023-12-07
Use SQLite to store our internal copy of the source catalogs.

### LINKED ISSUE #532: Design proposal for fixing the "Failed to fork a worker process: Cannot allocate memory" error
a.Add jitter during process forking
b.Set min_job (default limit as 1) and max_jobs for tables and indexes in pgcopydb . Also expose these as arguments, so that they can be altered if required.

### LINKED ISSUE #545: Support very large catalogs
At the moment pgcopydb allocates the source database catalogs in-memory with both a set of arrays (for tables, indexes, constraints, sequences, etc) and a hash-table construct on-top of it to enable oid based lookups, and restore-list-names lookups in some cases.

When given databases with a very large number of tables (e.g. 1,500,000 tables in 9,000 schemas would qualify), it could easily be that the needed catalogs won't fit in-memory anymore. We would need a way to spill-to-disk the giant array of tables (and other objects), and a way to implement a hash-table lookup that also knows how to spill-to-disk.

Fixes #532

dimitri:
This allows to implement an in-memory array with the capabilities to
spill-to-disk and also hash-table like lookup operations, and with good
multi-process concurrency handling.

Using SQLite internally will allow reducing the memory usage of pgcopydb and
make it possible to address databases with more than a million of tables
defined.

Fixes #545 
Fixes #532

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,7 @@ RUN dpkg --add-architecture ${TARGETARCH:-arm64} && apt update \
     libpq5 \
     libreadline-dev \
     libselinux1-dev \
+    libsqlite3-dev \
     libssl-dev \
     libxslt1-dev \
     lsof \
@@ -65,7 +66,8 @@ RUN dpkg --add-architecture ${TARGETARCH:-arm64} && apt update \
 	passwd \
     ca-certificates \
     libpq5 \
-    lsof \
+	libsqlite3-0 \
+	lsof \
     tmux \
     watch \
     psmisc \
--- a/docs/design.rst
+++ b/docs/design.rst
@@ -397,3 +397,76 @@ to use this feature:
     using same-table COPY concurrency.
 
     Use your usual Postgres configuration editing for testing.
+
+Internal Catalogs (SQLite)
+--------------------------
+
+To be able to implement pgcopydb operations, a list of SQL objects such as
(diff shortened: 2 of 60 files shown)


arajkumar:
Believe it or not, I was even contemplating the idea of using SQLite as a storage layer for .json and .sql data instead of files.

I think adding SQLite into pgcopydb will open up avenues of opportunities to improve its storage consistency.

dimitri:
> Believe it or not, I was even contemplating the idea of using SQLite as a storage layer for .json and .sql data instead of files.
> 
> I think adding SQLite into pgcopydb will open up avenues of opportunities to improve its storage consistency.

While I believe using SQLite for the JSON and SQL file could be a net win in terms of finding where we are in the file (we could use an LSN int64 column for that), I'm still not sure of how to design the SQLite database itself. We might have to store in there metadata + logical replication output plugin raw data, the same we do at the moment in our own JSON format.

More interestingly, the split-file approach that we have now and is the core of all the complexity on the replay side has been designed to avoid having a single giant 1TB JSON file to deal with. I suppose we would prefer it the same way if we had the data in a SQLite database, so we will have to design something that allows for splitting the data in several files again, I suppose. Unless dealing with a 1TB SQLite database is easy enough?

PR #572, opened 2023-12-11
Support skipping extensions & filtering in pgcopydb dump/restore cli

arajkumar:
Currently, pgcopydb dump & restore doesn't support skipping extensions or filtering because we don't populate the context with extension details & skip list information.

### DIFF
--- a/src/bin/pgcopydb/cli_dump.c
+++ b/src/bin/pgcopydb/cli_dump.c
@@ -38,6 +38,7 @@ static CommandLine dump_schema_command =
 		"  --source          Postgres URI to the source database\n"
 		"  --target          Directory where to save the dump files\n"
 		"  --dir             Work directory to use\n"
+		"  --skip-extensions Skip restoring extensions\n" \
 		"  --snapshot        Use snapshot obtained with pg_export_snapshot\n",
 		cli_dump_schema_getopts,
 		cli_dump_schema);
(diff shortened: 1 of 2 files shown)


dimitri:
Please rebase to the current main branch, now that the SQLite catalogs are merged you will see some impact in how to handle this. It might be that we don't need this PR anymore, but I'm not sure. The main difference is that we have an on-disk cache of the catalogs now.

arajkumar:
@dimitri It looks we still needthis, it enables --skip-extensions cli flag to dump & restore. Additionally I see quite a few issues while testing pgcopydb dump, pgcopydb restore cli programs. One of them is https://github.com/dimitri/pgcopydb/pull/585.

src/bin/pgcopydb/cli_dump.c
```diff
@@ -370,7 +380,7 @@ cli_dump_schema_section(CopyDBOptions *dumpDBoptions,
 		exit(EXIT_CODE_INTERNAL_ERROR);
 	}
 
-	if (!copydb_init_specs(&copySpecs, dumpDBoptions, DATA_SECTION_NONE))
+	if (!copydb_init_specs(&copySpecs, dumpDBoptions, DATA_SECTION_ALL))
```
dimitri, line 383: I'm not sure that we need DATA_SECTION_ALL here, but thinking about it, we might. After all we want to be able to implement the whole filtering.

dimitri (approved):

arajkumar:
@dimitri I found few more issues which are independent of this work. I created a separate PR to fix them: https://github.com/dimitri/pgcopydb/pull/587

dimitri (approved):

PR #573, opened 2023-12-12
Add quotes for identifiers during apply phase, fix duplicate begin transaction.

VaibhaveS:
- Add quotes for identifiers during apply phase
```
2023-12-11 20:54:32 20 ERROR pgsql.c:1949 [TARGET 8125] ERROR:  relation "public.abc" does not exist
2023-12-11 20:54:32 20 ERROR pgsql.c:1949 [TARGET 8125] LINE 1: INSERT INTO public.abc("", "...
2023-12-11 20:54:32 20 ERROR pgsql.c:1949 [TARGET 8125]                     ^
2023-12-11 20:54:32 20 ERROR ld_apply.c:603 Failed to apply SQL from file "/root/.local/share/pgcopydb/00000002000002D200000040.sql", see above for details
```

- Remove redundant begin
```
2023-12-06 17:50:58 41 WARN   pgsql.c:831               WARNING:  there is already a transaction in progress
```

### DIFF
--- a/src/bin/pgcopydb/ld_transform.c
+++ b/src/bin/pgcopydb/ld_transform.c
@@ -2061,7 +2061,7 @@ stream_write_insert(FILE *out, LogicalMessageInsert *insert)
 		/*
 		 * First, the PREPARE part.
 		 */
-		appendPQExpBuffer(buf, "INSERT INTO %s.%s ",
+		appendPQExpBuffer(buf, "INSERT INTO \"%s\".\"%s\" ",
 						  insert->nspname,
 						  insert->relname);
 
@@ -2198,7 +2198,7 @@ stream_write_update(FILE *out, LogicalMessageUpdate *update)
 		/*
 		 * First, the PREPARE part.
 		 */
-		appendPQExpBuffer(buf, "UPDATE %s.%s SET ",
+		appendPQExpBuffer(buf, "UPDATE \"%s\".\"%s\" SET ",
 						  update->nspname,
 						  update->relname);
 		int pos = 0;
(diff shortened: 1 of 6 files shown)


VaibhaveS:
@dimitri need your help here. How to handle double quote names like these -

```PREPARE 5eff0dcd AS INSERT INTO "public"."""dqname""" ("id") overriding system value VALUES ($1);```

File: src/bin/pgcopydb/ld_transform.c [hunk 1]
```diff
@@ -2061,7 +2061,7 @@ stream_write_insert(FILE *out, LogicalMessageInsert *insert)
 		/*
 		 * First, the PREPARE part.
 		 */
-		appendPQExpBuffer(buf, "INSERT INTO %s.%s ",
+		appendPQExpBuffer(buf, "INSERT INTO \"%s\".\"%s\" ",
```
shubhamdhama, line 2064: how about using `quote_identifier` from `"utils/builtins.h"`? @dimitri thoughts?

dimitri: If that's available to pgcopydb through libpq I think we should see about that, otherwise we can use [PQescapeIdentifier](https://www.postgresql.org/docs/current/libpq-exec.html#LIBPQ-PQESCAPEIDENTIFIER).

shubhamdhama: Oh I think it isn't available through libpq and I have some wrong impression of using postgres internal libraries for our usage.
I think `PQescapeIdentifier` is our obvious choice then.

PR #576, opened 2023-12-13
Add option to output numeric as string on wal2json

hanefi:
This change adds a new option `--wal2json-numeric-as-string` that changes wal2json plugin output format to print numeric data types as strings. This is accomplished by passing the
`--numeric-data-types-as-string` option to wal2json plugin.

This is useful to prevent precision loss when using wal2json plugin to stream changes from a database that uses numeric data types.

wal2json plugin version that supports `--numeric-data-types-as-string` option is required to use this pgcopydb option. As of today there is no official wal2json release that supports this option, but it is available on master branch of the project.

Relevant changes in wal2json plugin is at
https://github.com/eulerto/wal2json/pull/255


src/bin/pgcopydb/ld_stream.c
```diff
@@ -91,7 +91,29 @@ stream_init_specs(StreamSpecs *specs,
...
+					"true",
+					"true",
+					"true",
+					"true",
+					"true",
+					"pgcopydb.*"
+				}
+			};
+
+			KeyVal optionsNumericAsNumber = {
```
dimitri, line 116: Is there a way to share that block of code? Usually I would use a `#define` and refer to it in both arrays, like it's done in other places in the code (have a look at `srcSettings` definition in `copydb.c` for instance.

dimitri (changes requested):
I think we should add support for the `--plugin` and the `--wal2json-numeric-as-string` options in the environment too, as discussed.

src/bin/pgcopydb/ld_stream.c
hanefi: I see your point. I actually took a look at `srcSettings` before writing this piece of code, and I gave up after writing some even uglier piece of code.

I now have an alternative implementation where I have a single struct that I update by appending one option and increment the counter if a condition holds. I guess this is as clean as it can get.

If I were to use `#if` or similar directive on 3 lines, I'd have 6 lines in total. Now I have 7 non-empty lines and 2 of those only contain curly brackets and spaces. I guess this is the cleanest it can get.

src/bin/pgcopydb/ld_stream.c
```diff
@@ -111,6 +111,15 @@ stream_init_specs(StreamSpecs *specs,
...
 
+			if (specs->slot.wal2jsonNumericAsString)
+			{
+				int index = options.count;
+
+				options.keywords[index] = "numeric-data-types-as-string";
+				options.values[index] = "true";
+				options.count += 1;
+			}
+
```
dimitri, line 122: Where is the memory allocated? I think I would prefer the following approach, if we're going to make the code as simple as possible:

``` c
			KeyVal options = {
				.count = specs->slot.wal2jsonNumericAsString ? 7 : 6,
				.keywords = {
					"format-version",
					"include-xids",
					"include-schemas",
					"include-transaction",
					"include-types",
					"filter-tables",
					"numeric-data-types-as-string"
				},
				.values = {
					"2",
					"true",
					"true",
					"true",
					"true",
					"pgcopydb.*",
					"true"
				}
			};
```

hanefi: Your suggestion is definitely cleaner, although a bit hacky. I applied it after adding some comments.

KeyVal struct contains an array of char pointers with 64 elements. We already have reserved space for 64 char* here. My choice to set one such pointer to a string literal address is a weird choice. As long as we do not try to update the string and cause undefined behavior, I thought that it was ok. 

https://github.com/dimitri/pgcopydb/blob/2c14034531f29c690830e4dd5e04f14771cc0f70/src/bin/pgcopydb/parsing_utils.h#L40-L59

dimitri: That comment seems outdated, we don't use MAXCONNINFO anymore (because libpq does not use it either), we use `strdup` instead.

dimitri (approved):

PR #585, opened 2023-12-18
Fix args index bug due to incorrect evaluation order

arajkumar:
This happens while using `--skip-extensions` with `pgcopydb dump schema` command.


arajkumar:
CC: @dimitri

dimitri (approved):

dimitri:
Thanks for your continued testing and fixes!

PR #587, opened 2023-12-19
Fix broken --skip-extensions flag after #569

arajkumar:
There were few issues with --skip-extensions flag after #569 

1) Excluded extensions were not part of the filter 
2) Invalid memory access from pg_dump_db_extension_namespace_hook 
3) Filter table populated without checking whether it is already done or not.


src/bin/pgcopydb/pgcmd.c
```diff
@@ -504,25 +535,13 @@ pg_dump_db_extension_namespace_hook(void *ctx, SourceExtension *ext)
...
-					  "too many schema are excluded: "
-					  "argsIndex %d > %d",
-					  *(context->argsIndex) + 2,
-					  PG_CMD_MAX_ARG);
-			return false;
-		}
-
-		args[(*context->argsIndex)++] = "--exclude-schema";
-		args[(*context->argsIndex)++] = nspname;
+		extNamespaces[(*context->extNamespaceCount)++] = strdup(nspname);
```
arajkumar, line 544: This fixes the invalid memory access.

arajkumar:
cc: @dimitri

src/bin/pgcopydb/pgcmd.c
dimitri: Yeah but it's still missing a buffer overflow check before adding to the new array, right?

dimitri (changes requested):

PR #589, opened 2023-12-19
Fix crash on exit with --debug and PGCOPYDB_LOG_FILENAME

arajkumar:
We close the FILE which is used to log into file before calling
the `copydb_cleanup_sysv_resources` which still uses log functions leads
to crash due to invalid FILE* access.

This commit fixes the crash by closing the FILE after calling the
cleanup function.

### DIFF
--- a/src/bin/pgcopydb/main.c
+++ b/src/bin/pgcopydb/main.c
@@ -264,6 +264,13 @@ set_logger()
 static void
 unlink_system_res_atexit(void)
 {
+	(void) copydb_cleanup_sysv_resources(&system_res_array);
+
+	/*
+	 * close the log file after cleaning up System V resources. This ordering
+	 * is important because we want to log any errors that might happen during
+	 * System V resources clean-up.
+	 */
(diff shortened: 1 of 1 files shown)


src/bin/pgcopydb/lock_utils.c
```diff
@@ -208,7 +208,14 @@ semaphore_unlink(Semaphore *semaphore)
 
 	semun.val = 0;              /* unused, but keep compiler quiet */
 
-	log_debug("ipcrm -s %d", semaphore->semId);
+	/*
+	 * log has dependency on semaphore, so we can't use log_debug() here.
+	 * We use fformat() instead to avoid crash while exiting.
+	 */
```
dimitri, line 214: Well the semaphore has not been deleted yet, and the `log_debug` function is sync (returns when the logging has been done already), so I'm not sure I follow?

arajkumar: Wrong call!. I just fixed the actual issue in the new commit. PTAL.

PR #595, opened 2023-12-22
Fix escaping of identifiers while transforming for wal2json plugin.

shubhamdhama:
wal2json doesn't escape identifiers as test_decoding do. To fix this issue
we are using PQescapeIdentifier function. Even though this function doesn't
require PGconn to be connected, but creating an empty PGconn object was
unfeasible. Some fields of PGconn that are required by new versions of
libpq for PQescapeIdentifier are not available in older PGconn. So decided
to do this the right way, by creating PGconn with an actual server
connection. It turned out more simpler than managing a dummy PGconn object.


src/bin/pgcopydb/ld_transform.c
```diff
@@ -43,11 +44,44 @@ typedef struct TransformStreamCtx
...
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	if (!pgsql_begin(privateContext->transformPGSQL))
+	{
+		/* errors have already been logged */
+		return false;
+	}
```
dimitri, line 80: Why do we need to open a (long-running? idle?) transaction here? It needs to be properly commented/documented, or better yet, avoided if that's possible...

src/bin/pgcopydb/pgsql.c
```diff
@@ -5627,3 +5627,28 @@ parseSentinel(void *ctx, PGresult *result)
...
 	context->parsedOK = true;
 }
+
+/*
+ * escape_identifier escapes the (postgres) identifiers and the generated string
+ * is always quoted. It uses PQescapeIdentifier
+ * (https://www.postgresql.org/docs/current/libpq-exec.html#LIBPQ-PQESCAPEIDENTIFIER).
+ */
+char *
+escape_identifier(PGSQL *pgsql, char *src)
```
dimitri, line 5637: Please follow the per-module C nomenclature that we use here, which means naming that function `pgsql_escape_identifier`. Also add a note in the function comment about the memory allocation here, and how to free it. I believe `PQfreemem` should be used, right?

src/bin/pgcopydb/ld_transform.c
```diff
@@ -1579,6 +1662,12 @@ FreeLogicalTransaction(LogicalTransaction *tx)
 	tx->first = NULL;
 }
 
+void
+FreeLogicalMessageRelation(LogicalMessageRelation *table)
+{
+	free(table->nspname);
+	free(table->relname);
+}
```
dimitri, line 1670: Missing function level comment. I believe the function `PQfreemem` should be used on these pointers?

src/bin/pgcopydb/pgsql.c
```diff
@@ -5627,3 +5627,28 @@ parseSentinel(void *ctx, PGresult *result)
...
+{
+	PGconn *conn = pgsql->connection;
+	if (conn == NULL)
+	{
+		return NULL;
+	}
+
+	char *dst = PQescapeIdentifier(conn, src, strlen(src));
```
dimitri, line 5645: Please use a better variable name here, maybe `escapedIdentifier` or something like that. We tend to use `dst` everywhere as a PGSQL object that manages a connection to the target (destination) database.

src/bin/pgcopydb/ld_stream.h
```diff
@@ -135,32 +135,34 @@ typedef struct LogicalMessageTupleArray
 	LogicalMessageTuple *array; /* malloc'ed area */
 } LogicalMessageTupleArray;
 
+typedef struct LogicalMessageRelation
+{
+	char *nspname;
+	char *relname;
```
dimitri, line 141: Add inline comment that this is a malloc'ed area, please, as we do in the other cases.

src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -338,13 +339,16 @@ parseTestDecodingMessageHeader(TestDecodingHeader *header, const char *message)
...
 
-	/* grab the table schema.name */
-	strlcpy(header->nspname, idp, dot - idp + 1);
-	strlcpy(header->relname, dot + 1, sep - dot);
+	/*
+	 * Grab the table schema.name, which is already escaped by appropriately by
+	 * the plugin
+	 */
+	header->table.nspname = strndup(idp, dot - idp);
+	header->table.relname = strndup(dot + 1, sep - dot - 1);
```
dimitri, line 347: I think you meant *quoted* rather than *escaped*?

src/bin/pgcopydb/ld_transform.c
```diff
@@ -43,11 +44,44 @@ typedef struct TransformStreamCtx
...
 												   LogicalTransactionStatement *new);
 static bool coalesceLogicalTransactionStatement(LogicalTransaction *txn,
 												LogicalTransactionStatement *new);
 
+/*
+ * stream_transform_init_context_pgsql initializes StreamContext's
+ * transformPGSQL and opens a connection to the target. This is required to use
+ * PQescapeIdentifier API of libpq when escaping identifiers
+ */
+bool stream_transform_init_context_pgsql(StreamSpecs *specs)
```
dimitri, line 61: I wonder if the name is the best we can find. I think `stream_transform_context_init_pgsql` is better to me.

src/bin/pgcopydb/ld_transform.c
dimitri: Apparently it depends where they come from... so maybe we need to track the information with a new slot in the LogicalMessageRelation structure: `bool pqMemory` or something?

src/bin/pgcopydb/ld_wal2json.c
```diff
@@ -32,6 +33,9 @@
 #include "summary.h"
 
 
+static bool SetSchemaAndTableName(JSON_Object *jsobj,
+								  LogicalMessageRelation *table,
+								  PGSQL *pgsql);
```
dimitri, line 38: I think I would prefer the name `SetMessageRelation` for this function here.

src/bin/pgcopydb/ld_wal2json.c
```diff
@@ -274,6 +268,42 @@ parseWal2jsonMessage(StreamContext *privateContext,
...
+	table->nspname = escape_identifier(pgsql, schema);
+	if (table->nspname == NULL)
+	{
+		return false;
+	}
+
+	table->relname = escape_identifier(pgsql, relname);
+	if (table->relname == NULL)
+	{
+		free(table->nspname);
```
dimitri, line 301: We need to use `PQfreemem` here certainly, right?

dimitri (changes requested):

src/bin/pgcopydb/ld_wal2json.c
shubhamdhama: Yes documentation suggests to use `PQfreemem` but it felt out of place (I mean most libpq stuff are used within pgsql.c) so I simply used `free` after looking at it's implementation (below). But I agree and understand we should the suggested API

```c
/*
 *		PQfreemem - safely frees memory allocated
 *
 * Needed mostly by Win32, unless multithreaded DLL (/MD in VC6)
 * Used for freeing memory from PQescapeBytea()/PQunescapeBytea()
 */
void
PQfreemem(void *ptr)
{
	free(ptr);
}
```

src/bin/pgcopydb/ld_transform.c
shubhamdhama: Yeah while developing I did this to save 10 seconds of moving `pgsql_open_connection` as a library function. Fixing this.

dimitri: I see. Well please expose something else, a kind of a new wrapper to it, I think we should not expose `pgsql_open_connection` to the outside of the module, because we want to hide some of our PGResults smarts in there and automate closing the connection after the query is complete.

shubhamdhama:
Follow-up
- [ ] Fix escaping of column names
- [ ] Debug flaky cdc-endpos... tests. Most likely unrelated to this change.

dimitri (approved):

src/bin/pgcopydb/ld_transform.c
dimitri: Thinking more about it, ending up with an alias for the same function is not going to do any good to the project. Let's make progress with exposing it and maybe revisit later.

src/bin/pgcopydb/ld_wal2json.c
```diff
@@ -120,18 +125,13 @@ parseWal2jsonMessage(StreamContext *privateContext,
...
 
-	char *schema = NULL;
-	char *table = NULL;
-
-	schema = (char *) json_object_dotget_string(jsobj, "message.schema");
-	table = (char *) json_object_dotget_string(jsobj, "message.table");
+	LogicalMessageRelation table = { 0 };
 
-	if (schema == NULL || table == NULL)
+	if (!SetMessageRelation(jsobj, &table, privateContext->transformPGSQL))
```
arajkumar, line 130: @dimitri @shubhamdhama Since we do this for every DML operation, I'm worried about the additional network latency introduced by this. Can we just copy the function from Postgres source which performs escaping?

arajkumar: Ahh, never mind. It doesn't do any networking.! Sorry for the false alarm.

PR #596, opened 2023-12-22
Attach source catalog even for follow only mode

arajkumar:
test_decoding transform relies on source catalog to decode the update message.

Prior to this commit, update while using test_decoding would work only while doing `pgcopydb clone --follow`. This commit enables support for update message decoding while using test_decoding plugin in `pgcopydb follow` too.

### DIFF
--- a/tests/cdc-endpos-between-transaction/copydb.sh
+++ b/tests/cdc-endpos-between-transaction/copydb.sh
@@ -36,9 +36,6 @@ pgcopydb stream setup
 # pgcopydb copy db uses the environment variables
 pgcopydb clone
 
-kill -TERM ${COPROC_PID}
-wait ${COPROC_PID}
-
 # now that the copying is done, inject some SQL DML changes to the source
 psql -d ${PGCOPYDB_SOURCE_PGURI} -f /usr/src/pgcopydb/dml.sql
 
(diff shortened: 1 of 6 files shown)


tests/cdc-endpos-between-transaction/copydb.sh
```diff
@@ -105,5 +102,8 @@ pgcopydb follow --resume --trace
 sql="select count(*) from category"
 test 26 -eq `psql -AtqX -d ${PGCOPYDB_SOURCE_PGURI} -c "${sql}"`
 
+kill -TERM ${COPROC_PID}
+wait ${COPROC_PID}
+
```
arajkumar, line 107: Now the snapshot is needed until follow finishes!, is it okay? or better to go with --not-consistent snapshots?

dimitri (approved):

tests/cdc-endpos-between-transaction/copydb.sh
dimitri: It's okay! We need a variety of tests covering multiple usages, I don't have a strong opinion on whether a particular test should hold on to its snapshot for what period of time. Tests should also be good implementation references when possible, so using consistency here is good too.

dimitri:
Thanks @arajkumar ! I believe this also fixes a use-case that wasn't very clean before when it comes to only using the *follow* parts of pgcopydb.

arajkumar:
Thanks @dimitri. Indeed, it will be useful for follow only use-cases.

PR #601, opened 2023-12-27
Use our SQLite catalogs for the pgcopydb sentinel table.

dimitri:
The main use-case for the sentinel table is to communicate progress made by
the three independent worker processes for stream, transform, and replay.
It is also needed to communicate new startpos, endpos and apply values with
those processes from another pgcopydb (interactive) command, a process
that's not in the process tree.
    
This was previously solved by creating and using a pgcopydb.sentinel table
on the source database. Now that we have a SQLite database internal to the
pgcopydb run, we can use that instead.


src/bin/pgcopydb/sentinel.c
```diff
@@ -0,0 +1,582 @@
...
+		log_error("BUG: sentinel_get: db is NULL");
+		return false;
+	}
+
+	char *sql =
+		"select startpos, endpos, apply, write_lsn, flush_lsn, replay_lsn "
+		"  from sentinel "
+		" where id = 1";
+
+	if (!semaphore_lock(&(catalog->sema)))
```
arajkumar, line 428: Instead of explicit serialisation, can't we rely on the default sqlite behavior while accessing the DB from multiple processes?

According to the https://www.sqlite.org/faq.html, sqlite would return SQLITE_BUSY when the the DB file is already opened by the other process, maybe we could retry few a times on SQLITE_BUSY and abort? WDYT?

dimitri: That's a good idea and that's what the code used to be doing up to recently. In a single-writer multiple-readers scenario, that'd be all we need for happily benefiting from SQLite concurrency facilities. Unfortunately our code is not single-writer, and we need to discipline it in a way that it is: that's why we have explicit serialization. It's up to pgcopydb code to ensure reliable operations by having a single SQLite writer at all times.

At least I failed to make it work any other way, and I tried.

PR #612, opened 2024-01-03
Makefile: Add DEBUG variable to produce debug friendly build.

shubhamdhama:
Usage: `make DEBUG=1 install`


shubhamdhama:
Using the `Og` flag while debugging has proven helpful to me on several occasions. Alternatively, are there any other methods to achieve this?

arajkumar (approved):
Nice. I used to fiddle with makefile directly :)

dimitri (approved):

PR #613, opened 2024-01-04
Add libgc (Boehm-Demers-Weiser GC) for garbage collection.

shubhamdhama:
With growth of pgcopydb project we should also have some smart memory management as often we miss free'ing dynamically allocated memory. The easiest way to get started here is to use [Boehm-Demers-Weiser Garbage Collector](https://github.com/ivmai/bdwgc) which is pretty convenient to integrate.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN dpkg --add-architecture ${TARGETARCH:-arm64} && apt update \
     libpam-dev \
     libpq-dev \
     libpq5 \
+	libgc-dev \
     libreadline-dev \
     libselinux1-dev \
     libsqlite3-dev \
--- a/src/bin/pgcopydb/Makefile
+++ b/src/bin/pgcopydb/Makefile
@@ -80,6 +80,7 @@ LIBS += $(shell $(PG_CONFIG) --libs)
 LIBS += -lpq
 LIBS += -lncurses
 LIBS += -lsqlite3
+LIBS += -lgc
 
 all: $(PGCOPYDB) ;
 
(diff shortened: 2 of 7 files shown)


shubhamdhama:
TODOs before publishing this PR
- [ ] Validation/metrics that this really works
- [ ] Plan/caution for using `GC_malloc` e.g. https://www.hboehm.info/gc/conservative.html and _"It is usually best not to mix garbage-collected allocation with the system malloc-free. If you do, you need to be careful not to store pointers to the garbage-collected heap in memory allocated with the system malloc."_

arajkumar:
I don't have any opinion on this.

Unrelated: While researching about libgc, I found Postgres memory management which is very interesting[1][2].

[1] https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob_plain;f=src/backend/utils/mmgr/README;hb=HEAD
[2] https://www.cybertec-postgresql.com/en/memory-context-for-postgresql-memory-management

dimitri:
Two comments:

 - is it possible to use the Garbage Collector in a transparent way, where calloc and malloc are actually calling GC_malloc under the hood?
 - yes Postgres memory management system is interesting, but it's backend only, I don't know that we can import and use that code within a client-side application even when written in C just like Postgres... there is libgc, but there is no libpostgres, unfortunately.

arajkumar:
@shubhamdhama  I would be interested to see the difference in memory usage(RSS?) between the manual & libgc approach. I can give a try on my side too.

@dimitri As you said, we may not able to reuse Postgres implemenation, but there are few [arena allocators](https://github.com/tsoding/arena) which we can reuse if we want to go down that path. But as of now, libgc approach looks simple to me, hope it works!

PR #614, opened 2024-01-04
Implement filtering logic for logical messages

### LINKED ISSUE #510: Add filtering support for “online” migrations (logical decoding)
The wal2json plugin allows for filtering but pgcopydb does not fill-in that setting entry with the list of filtered-out tables from its own setup at the moment. The test_decoding plugin lacks the feature entirely, so in that case we would need to filter-out events on the receiving end.

It might be worth filtering on the receiving end to get started with, so that we share a single implementation for both plugins.

gokhangulbiz:
This pull request adds the implementation to filter out logical messages based on source filters for `--follow` operations.

It introduces the `ShouldFilterOutMessage(..)` function in `ld_stream.c` to determine whether a message needs to be filtered out based on the provided source filters. When a logical message is received, the logical message parsers (for both test_decoding and wal2json) can set the message `metadata.filterout` field by calling the introduced function to prevent further processing.

Fixes #510


arajkumar:
@gokhangulbiz  Thanks for your contribution. Could you please rebase against latest main?

src/bin/pgcopydb/ld_stream.c
```diff
@@ -35,6 +35,150 @@
...
+	if (relname == NULL || strlen(relname) == 0)
+	{
+		log_error("BUG: relname is NULL or empty");
+		return false;
+	}
+
+	/*
+	 * Filter-out pgcopydb.* in any case
+	 */
+	if (strcmp(nspname, "pgcopydb") == 0)
```
arajkumar, line 67: Use streq instead of strcmp across the change.

```suggestion
	if (streq(nspname, "pgcopydb"))
```

src/bin/pgcopydb/ld_stream.c
```diff
@@ -35,6 +35,150 @@
...
+ * provided filters. The caller function should have already parsed and
+ * validated the filters.
+ */
+bool
+ShouldFilterOutMessage(SourceFilters *filters, char *nspname, char *relname)
+{
+	/*
+	 * Validate nspname is not NULL or empty
+	 */
+	if (nspname == NULL || strlen(nspname) == 0)
```
arajkumar, line 49: Use IS_EMPTY_STRING_BUFFER to check whether the given C string is empty.
```suggestion
	if (nspname == NULL || IS_EMPTY_STRING_BUFFER(nspname))
```

tests/filtering/exclude.ini
```diff
@@ -4,12 +4,5 @@ bar
 schema_name_20_chars
 
 [exclude-table]
-; public.actor
-; public.category
-; public.film
-; public.film_actor
-; public.film_category
-; public.language
-; public.rental
```
arajkumar, line 13: May I know why this has been removed? I think this has been kept to check the commenting aspect of ini file.

arajkumar (commented):
Ive added a few trivial comments not related to implementation logic.

.github/workflows/docker-publish.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
arajkumar, line 36: I would recommend to keep only the relevant changes to the problem.

tests/filtering/exclude.ini
gokhangulbiz: Good catch! I initially removed the commented lines, thinking they were leftovers. However, I understand now that they may serve a purpose in checking the commenting aspect of the INI file. I added back the commented lines and included a clear comment on top. [5a9d1c9](https://github.com/dimitri/pgcopydb/pull/614/commits/5a9d1c90ce0cd3e125bf2df958d6bef0d0c9276f)

.github/workflows/docker-publish.yml
gokhangulbiz: Aah, agreed with you. My bad :/ Reverted it ([0548fd3](https://github.com/dimitri/pgcopydb/pull/614/commits/0548fd367b2c44d31e51f3c74c1fa671b5a04d89)) and created a new PR for this change. See #620

src/bin/pgcopydb/ld_stream.c
```diff
@@ -35,6 +35,150 @@
...
+	}
+
+	/*
+	 * Filter-out pgcopydb.* in any case
+	 */
+	if (streq(nspname, "pgcopydb"))
+	{
+		log_info("Filtering out message for pgcopydb internal tables \"%s\"", nspname);
+		return true;
+	}
```
dimitri, line 71: Now that the sentinel table is created in the SQLite database catalogs, rather than on the source server, do we still have `pgcopydb.*` tables to filter out at all?

src/bin/pgcopydb/ld_stream.c
```diff
@@ -35,6 +35,150 @@
...
+
+	/*
+	 * If no filters are set, then we don't need to filter out any messages
+	 */
+	if (filters->type == SOURCE_FILTER_TYPE_NONE)
+	{
+		return false;
+	}
```
dimitri, line 79: We can't use the return value of the function for two separate meanings: either it means the function could operate as intended (see the previous return false cases) or it means that `ShouldFilterOutMessage` is true/false. I advice doing like we do in most of the other code where that happens, return true/false to indicate that the function could operate as intended, and add a `bool *shouldFitlterOutMessage` argument to the function.

src/bin/pgcopydb/ld_stream.c
```diff
@@ -35,6 +35,150 @@
...
+
+	/*
+	 * Filtering based on the sourceFilters.exclude-table-data
+	 * return true if the table is in the list
+	 * exclude-table-data filters should be processed first, otherwise
+	 * include-only-table will be able to filter the message before
+	 * we can check if the table is in the exclude-table-data list
+	 * and we will not be able to filter out the message.
+	 */
```
dimitri, line 88: Please review the filtersDB SQLite catalog, I believe you could implement a single look-up in the main filter table, or maybe in the s_table table and have all the information you need in a single lookup.

src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -171,15 +171,16 @@ parseTestDecodingMessageActionAndXid(LogicalStreamContext *context)
...
+
+		/*
+		 * Check if the message should be filtered out based on namespace and relation name
+		 */
+		if (ShouldFilterOutMessage(&(privateContext->filters), header.table.nspname,
+								   header.table.relname))
 		{
-			log_debug("Filtering out message for schema \"%s\": %s",
-					  header.table.nspname,
-					  context->buffer);
```
dimitri, line 178: Maybe we should keep a `log_trace` about that decision?

src/bin/pgcopydb/ld_wal2json.c
```diff
@@ -95,6 +95,40 @@ parseWal2jsonMessageActionAndXid(LogicalStreamContext *context)
 		return false;
 	}
 
+	/*
+	 * if metadata action is one of INSERT, UPDATE, DELETE, TRUNCATE
+	 * then we need to parse the schema and table name from the message
+	 * and check if the message should be filtered out
+	 */
```
dimitri, line 102: I thought we could use the wal2json filtering abilities and append the list of tables we filter-out to the proper option when creating the replication slot with the wal2json plugin. Did you try that approach? Is there a learning where we can't use that option and have to filter-out in our code here? Do you think maybe we need to do both (defensive programming)?

tests/filtering/copydb.sh
```diff
@@ -49,16 +49,53 @@ select NULL as oid, s.restore_list_name, 'sequence owned by'
        (select 1 from source.s_table st where st.oid = s.ownedby);
 EOF
 
-pgcopydb clone --filters /usr/src/pgcopydb/exclude.ini --resume --not-consistent
+# Starts a coprocess to clone the database using the 'wal2json' plugin, applying filters from 'exclude.ini'
+coproc (pgcopydb clone --follow --plugin wal2json --filters /usr/src/pgcopydb/exclude.ini --restart)
```
dimitri, line 53: That's a little strange to me. Also we're not calling wait on the PID of the coproc to make sure to wait until the command is over with. In the other test-cases we use a separate docker service when we have to run such commands in the background, but as it's just a single one maybe using coproc is ok... as long as we then use `wait`.

dimitri (changes requested):
First round of review! Thanks

src/bin/pgcopydb/ld_stream.c
gokhangulbiz: [0e4dc40](https://github.com/dimitri/pgcopydb/pull/614/commits/0e4dc403e4790c38da8173b042d652ee8fab0f48)

src/bin/pgcopydb/ld_test_decoding.c
gokhangulbiz: [62ecc5a](https://github.com/dimitri/pgcopydb/pull/614/commits/62ecc5aaf46db0797e3768e17c4f180bb20668e2)

tests/filtering/copydb.sh
gokhangulbiz: [179ff2a](https://github.com/dimitri/pgcopydb/pull/614/commits/179ff2ade7a2618c46bbbc333e563e2a83326afd)
[80fb72c](https://github.com/dimitri/pgcopydb/pull/614/commits/80fb72c660381586325ae043f1852f9957148855)

src/bin/pgcopydb/ld_stream.c
gokhangulbiz: [9250199](https://github.com/dimitri/pgcopydb/pull/614/commits/925019998251e5634944189a9975af6e75c8172a)

src/bin/pgcopydb/ld_stream.c
```diff
@@ -35,6 +35,93 @@
...
+										nspname,
+										relname,
+										table))
+	{
+		log_error("Failed to lookup for table \"%s\".\"%s\" in our "
+				  "internal source catalogs",
+				  nspname,
+				  relname);
+
+		free(table);
```
arajkumar, line 104: You can either use stack allocation or remove free here, it is no longer necessary after https://github.com/dimitri/pgcopydb/pull/647

src/bin/pgcopydb/ld_stream.c
```diff
@@ -35,6 +35,93 @@
...
+	 * message if exclude-data is set to true. Otherwise, we should filter out.
+	 */
+	if (table->oid != 0)
+	{
+		*shouldFilterOutMessage = table->excludeData;
+	}
+	else
+	{
+		*shouldFilterOutMessage = true;
+	}
```
arajkumar, line 119: nit: I prefer to do this, but I leave it upto you to decide which one is more readable.

```suggestion
	*shouldFilterOutMessage = table->oid != 0 && table->excludeData;
```

src/bin/pgcopydb/ld_stream.c
```diff
@@ -100,29 +190,84 @@ stream_init_specs(StreamSpecs *specs,
 
 		case STREAM_PLUGIN_WAL2JSON:
 		{
+			/* Initialize empty strings for tables to add and filter */
+			char *commaSeperatedAddTables = { 0 };
+			char *commaSeperatedFilterTables = { 0 };
+
```
arajkumar, line 196: I would suggest to declare the PQExpBuffer addTablesBuffer & PQExpBuffer filterTablesBuffer here and destroy it at the end of function or on error.
```suggestion
			PQExpBuffer addTablesBuffer = createPQExpBuffer();
			PQExpBuffer filterTablesBuffer = createPQExpBuffer();

```

src/bin/pgcopydb/ld_stream.c
```diff
@@ -100,29 +190,84 @@ stream_init_specs(StreamSpecs *specs,
...
+			if (filters->type != SOURCE_FILTER_TYPE_NONE)
+			{
+				/* convert filters to wal2json options. */
+
+				PQExpBuffer addTablesBuffer = createPQExpBuffer();
+				PQExpBuffer filterTablesBuffer = createPQExpBuffer();
```
arajkumar, line 202:
```suggestion
```

src/bin/pgcopydb/ld_stream.c
```diff
@@ -100,29 +190,84 @@ stream_init_specs(StreamSpecs *specs,
...
+				{
+					log_error("Failed to create wal2json filters");
+					return false;
+				}
+
+				commaSeperatedAddTables = strdup(addTablesBuffer->data);
+				commaSeperatedFilterTables = strdup(filterTablesBuffer->data);
+
+				destroyPQExpBuffer(addTablesBuffer);
+				destroyPQExpBuffer(filterTablesBuffer);
```
arajkumar, line 216:
```suggestion
```

src/bin/pgcopydb/ld_stream.c
```diff
@@ -100,29 +190,84 @@ stream_init_specs(StreamSpecs *specs,
...
+
+				PQExpBuffer addTablesBuffer = createPQExpBuffer();
+				PQExpBuffer filterTablesBuffer = createPQExpBuffer();
+
+				if (!convertFiltersToWal2JsonOptions(filters,
+													 addTablesBuffer,
+													 filterTablesBuffer))
+				{
+					log_error("Failed to create wal2json filters");
+					return false;
```
arajkumar, line 209:
```suggestion
                                         destroyPQExpBuffer(addTablesBuffer);
				        destroyPQExpBuffer(filterTablesBuffer);
					return false;
```

src/bin/pgcopydb/ld_stream.c
```diff
@@ -221,6 +366,106 @@ stream_init_specs(StreamSpecs *specs,
...
+						  filteredRelName);
+	}
+
+	if (PQExpBufferBroken(commaSeperatedAddTables) ||
+		PQExpBufferBroken(commaSeperatedFilterTables))
+	{
+		log_error("Failed to convert source filters to wal2json filters: out of memory");
+
+		destroyPQExpBuffer(commaSeperatedAddTables);
+		destroyPQExpBuffer(commaSeperatedFilterTables);
```
arajkumar, line 460: I would rather leave the caller to deallocate these because it was allocated by the caller.

src/bin/pgcopydb/ld_stream.c
```diff
@@ -100,29 +190,84 @@ stream_init_specs(StreamSpecs *specs,
...
+			}
+
+			/* Add the "filter-tables" option to the wal2json plugin options */
+			if (commaSeperatedFilterTables != NULL && !IS_EMPTY_STRING_BUFFER(
+					commaSeperatedFilterTables))
+			{
+				options.keywords[options.count] = "filter-tables";
+				options.values[options.count] = commaSeperatedFilterTables;
+				options.count++;
+			}
```
arajkumar, line 253:
```suggestion
			/* Add the "add-tables" option to the wal2json plugin options */
			if (addTablesBuffer.len > 0)
			{
				options.keywords[options.count] = "add-tables";
				options.values[options.count] = addTablesBuffer.data;
				options.count++;
			}

			/* Add the "filter-tables" option to the wal2json plugin options */
			if (filterTablesBuffer.len > 0)
			{
				options.keywords[options.count] = "filter-tables";
				options.values[options.count] = filterTablesBuffer.data;
				options.count++;
			}
			
			destroyPQExpBuffer(addTablesBuffer);
			destroyPQExpBuffer(filterTablesBuffer);
```

tests/filtering/copydb.sh
```diff
@@ -49,16 +49,47 @@ select NULL as oid, s.restore_list_name, 'sequence owned by'
...
+psql -d "${PGCOPYDB_SOURCE_PGURI}" ${pgopts} --file ./follow-mode/dml-for-exclude.sql
+
+# make sure the inject service has had time to see the final sentinel values
+sleep 2
+
+# set the end position to the current position to complete the follow operation
+pgcopydb stream sentinel set endpos --current
+
+# the follow command ends when reaching endpos
+wait ${COPROC_PID}
```
arajkumar, line 65: I would suggest to use follow subcommand here.

```suggestion
pgcopydb clone --filters /usr/src/pgcopydb/exclude.ini --resume --not-consistent

# execute DML queries against the source database to test filtering for exclude.ini
psql -d "${PGCOPYDB_SOURCE_PGURI}" ${pgopts} --file ./follow-mode/dml-for-exclude.sql

# set the end position to the current position to complete the follow operation
pgcopydb stream sentinel set endpos --current

# the follow command ends when reaching endpos
pgcopydb follow --filters /usr/src/pgcopydb/exclude.ini --resume --not-consistent

```

tests/filtering/copydb.sh
```diff
@@ -49,16 +49,47 @@ select NULL as oid, s.restore_list_name, 'sequence owned by'
...
+
+# cleanup the stream
+pgcopydb stream cleanup
```
arajkumar, line 92:
```suggestion
# now another migration with the "include-only" parts of the data by using text_decoding plugin
pgcopydb clone --filters /usr/src/pgcopydb/include.ini --resume --not-consistent

# execute DML queries against the source database to test filtering for include.ini
psql -d "${PGCOPYDB_SOURCE_PGURI}" ${pgopts} --file ./follow-mode/dml-for-include.sql

# set the end position to the current position to complete the follow operation
pgcopydb stream sentinel set endpos --current

# the follow command ends when reaching endpos
pgcopydb follow --filters /usr/src/pgcopydb/include.ini --resume --not-consistent

# cleanup the stream
pgcopydb stream cleanup
```

arajkumar (changes requested):

arajkumar (changes requested):
Please rebase against the current main & fix style issues. You can use `make indent` to check all the styling/formatting issues, which actually needs https://github.com/citusdata/tools/tree/develop/uncrustify

arajkumar:
@gokhangulbiz Is this PR on your radar still?

dcupif:
Hi, I'm interested in knowing if this MR is going to see through? We want to run an online migration with CDC to copy over only a subset of tables from a database.
Right now, after doing some testing, it seems that the SQL files generated from the WAL consumption contain statements that belong to tables excluded by the filter I'm using.
Since these excluded tables won't be recreated on the target instance, replaying these SQL files will result in errors.

arajkumar:
@dcupif As an workaround, we use [exclude-table-data], which avoids copying initial data and avoids errors during CDC. Once the migration is over, the unwanted table can be dropped. This might be useful when the candidate table has a lots of exists data, but fewer DMLs during migration.

Of course, this won't be useful if the excluded table attracts lots of DML during CDC.

dcupif:
> Of course, this won't be useful if the excluded table attracts lots of DML during CDC.

Yes, that's my concern. We have mutualized Postgres servers (e.g. dozens of databases hosted on the same server) with a lot of activity. Not filtering the DML statements means that pgcopydb can store a lot of changes, which could ultimately incur storage issues. I believe we could use an infinite storage mount like S3 or something, but this seems extremely inefficient.

PR #608, opened 2024-01-02
Refactor code to fetch current LSN position from Postgres.

dimitri:

PR #615, opened 2024-01-04
Problem: Incorrect flush LSN from source being used to find durable LSN

arajkumar:
Solution: Use flush LSN from target to find the durable LSN.

To find a durable LSN with async commit, we maintain a list of LSN mapping between source(sourceLSN) and target(insertLSN). New mapping will be inserted into the list on COMMIT or KEEPALIVE message.

While reporting the replay_lsn to sentinel we have to get the flush lsn from target and compare it against the insertLSN from the list to find the sourceLSN.

This has been discussed in https://github.com/dimitri/pgcopydb/pull/608#issuecomment-1875144153


dimitri (approved):

dimitri:
Thanks @arajkumar ! That looks almost like a typo that I did, all with the function comments saying the right thing and the code using the wrong connection string then... nice catch!

PR #617, opened 2024-01-04
Fix numerous issues with argument parsing

hanefi:
The following issues were fixed:
- some options were not recognized
- some nonexistent options were recognized
- some options with required arguments were not expecting an argument
- some commands did not print help messages when given invalid options

I also removed `root_with_debug` global variable that is identical to root CommandLine.

### DIFF
--- a/src/bin/pgcopydb/cli_restore.c
+++ b/src/bin/pgcopydb/cli_restore.c
@@ -153,7 +153,6 @@ cli_restore_schema_getopts(int argc, char **argv)
 		{ "source", required_argument, NULL, 'S' },
 		{ "target", required_argument, NULL, 'T' },
 		{ "dir", required_argument, NULL, 'D' },
-		{ "schema", required_argument, NULL, 's' },
 		{ "drop-if-exists", no_argument, NULL, 'c' }, /* pg_restore -c */
 		{ "no-owner", no_argument, NULL, 'O' },       /* pg_restore -O */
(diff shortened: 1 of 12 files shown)


src/bin/pgcopydb/cli_restore.c
```diff
@@ -153,7 +153,6 @@ cli_restore_schema_getopts(int argc, char **argv)
 		{ "source", required_argument, NULL, 'S' },
 		{ "target", required_argument, NULL, 'T' },
 		{ "dir", required_argument, NULL, 'D' },
-		{ "schema", required_argument, NULL, 's' },
```
hanefi, line 156: @dimitri please let me know if that is ok to remove this line.

This particular option does not seem to exist anywhere in the codebase today. I suspect, this is a remnant of a removed argument.

hanefi: We have `--schema-name` option in `cli_list_db_getopts`. The trace log of `--schema` in there makes me think that we renamed it at some point.

Shall I rename this command to `--schema-name` and add the case block in the same function?

dimitri (approved):

src/bin/pgcopydb/cli_restore.c
dimitri: I can't remember why we have `--schema` in there, seems ok to just remove it now as part of that cleanup PR. Thanks!

PR #618, opened 2024-01-04
Improve documentation templates

hanefi:
- Adds CI to check that the documentation templates are up-to-date.
- Add a new `help` command to the documentation templates. This command prints the help message for all the sub-commands of `pgcopydb` in a single file. This is useful for users who want to see all the sub-commands in one place.
- Remove dependency on installing the binary before updating the documentation templates. This allows users to update the documentation templates without installing the binary. One can set the `PGCOPYDB` environment variable to point to the binary to use for updating the documentation templates.
- Fix a casing issue in command descriptions. The first letter of the first word in the description of each command should be capitalized.


Dockerfile
```diff
@@ -49,6 +49,21 @@ COPY . .
 
 RUN make -s clean && make -s -j$(nproc) install
 
+# Now the "docs-template-testing" image, to check if the docs are up to date
+FROM --platform=${TARGETPLATFORM} build as docs-template-testing
```
dimitri, line 53: Does this means we run the docs-template-testing bits each time we want to run a test case (using e.g. `make tests/pagila`) on our local systems?

Makefile
```diff
@@ -21,8 +21,8 @@ maintainer-clean: clean
 docs:
 	$(MAKE) -C docs clean man html
 
-update-docs: install
-	bash ./docs/update-help-messages.sh
+update-docs: bin
+	PGCOPYDB=${TOP}src/bin/pgcopydb/pgcopydb bash ./docs/update-help-messages.sh
```
dimitri, line 25: Please define the PGCOPYDB make variable earlier in the file in a way that it can be overloaded from the outside, typically:

```
PGCOPYDB ?= $(TOP)src/bin/pgcopydb/pgcopydb
```

docs/update-help-messages.sh
```diff
@@ -18,17 +19,26 @@ function print_help_to_file() {
...
+    #   `pgcopydb compare data --help` to `compare-data.rst` etc.
+    #
+    #  One exception is the `pgcopydb help` command. We print the output of
+    #  `pgcopydb help` to `help.rst` instead of supplying a `--help` argument to
+    #  the command.
     local file_path
     local help_cmd
     if [ "${cmd}" = "" ]; then
-        help_cmd="pgcopydb --help 2>&1"
+        help_cmd="$PGCOPYDB --help 2>&1"
```
dimitri, line 35: Please use `${PGCOPYDB}` notation with the `{}` as everywhere else in our shell scripts.

dimitri: Also please install a default value in the case that the environment variable has not been set.

dimitri (changes requested):

Dockerfile
hanefi: No. Docker is clever enough to skip build stages that are not used. All thanks to the default backend (since last February) BuildKit for Docker.

See the list at the beginning of the page for [BuildKit](https://docs.docker.com/build/buildkit/)

dimitri: I see. I think I would still prefer it if we had a separate Dockerfile for the docs, does it make sense to split things that way?

hanefi: Docker does a poor job managing dependencies across multiple Dockerfiles. I created a new `Dockerfile.tests` file with a copy of our build stage so that we can hit docker build cache.

.github/workflows/run-tests.yml
```diff
@@ -75,3 +75,7 @@ jobs:
         uses: ammaraskar/sphinx-action@master
         with:
           docs-folder: "docs/"
+      - name: Check that docs are up to date
+        run: |
+          docker build --file=Dockerfile.tests --tag tests . && \
+          docker run tests
```
dimitri, line 81: We now have two main Dockerfile, the one that we use to run tests and the one that we use to build documentation. Why is the documentation one named `Dockerfile.tests`?

hanefi: Currently it does not build docs. It only checks (or tests) if the documentation templates are up to date.

Is there any other name you'd prefer? Let me move this file to `ci/Dockerfile` if it will make it clear that it is not a main Dockerfile.

dimitri: Well I think I would prefer `ci/Dockerfile.docs` (the move to ci/ is nice). And if the contents have to contain a copy/paste version of the main Dockerfile, can we script that, use a template/include mechanism (only in the ci/ version, leave the main Dockerfile alone/unchanged)?

hanefi: Moved the file to `ci/Dockerfile.docs.template`. One can run the following command to generate the final Dockerfile that is ignored in the git worktree:

```sh
cat Dockerfile ci/Dockerfile.docs.template > ci/Dockerfile.docs
```

I do not expect users to actually use this dockerfile in any way. I think `make update-docs` should be the command to run on development environments.

dimitri (approved):

PR #631, opened 2024-01-12
Problem: float values loses precision after SQL transformation

arajkumar:
Solution: Use format string `%.15f` which would retain 15 digits after the decimal point.

### DIFF
--- a/src/bin/pgcopydb/ld_transform.c
+++ b/src/bin/pgcopydb/ld_transform.c
@@ -2607,7 +2607,7 @@ stream_add_value_in_json_array(LogicalMessageValue *value, JSON_Array *jsArray)
 				}
 				else
 				{
-					sformat(string, sizeof(string), "%f", value->val.float8);
+					sformat(string, sizeof(string), "%.15f", value->val.float8);
 				}
 
 				json_array_append_string(jsArray, string);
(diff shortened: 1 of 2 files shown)


arajkumar:
@dimitri I'm not sure this is a right solution, I wanted to understand why we settled on `%f` which outputs floating point values with low resolution.?

I had gone through [1][2] and I agree that [1] will be the right fix for the problem, I would still like to understand why not just bump up the resolution as proposed in this fix? I can add tests if required.

[1] https://github.com/eulerto/wal2json/pull/255
[2] https://github.com/dimitri/pgcopydb/commit/ebe3f3327d4b078fad436740ad1e1fe955b903c8 (edited)

shubhamdhama:
@arajkumar The actual problem is here,
```
			case JSONNumber:
			{
				double x = json_value_get_number(jsval);

				valueColumn->oid = FLOAT8OID;
				valueColumn->val.float8 = x;
				valueColumn->isNull = false;
				break;
			}
```

I believe JSON doesn't have limit on the length of numeric numbers. But when it's converted to a `parson` JSON object it detects its type as number. Hence the precision of `numeric` data type, which is pretty much endless, is lost when stored in `double` of C. Long time back I tried to patch `parson` to return numeric data type as string as well which worked fine (or maybe there is a way out of box in parson to read numeric as string in the first place). But then you realize Postgres supports +/-`NaN` and +/- `Inf` which comes as `null` from wal2json. That would lead to loss of data integrity. I guess you already know this from linked discussion.

That's why for `wal2json` can't be used reliable without `numeric-data-types-as-string` so the code path you pointed above is not used practically (and perhaps we can even delete it).

arajkumar:
Thanks @shubhamdhama.

PR #633, opened 2024-01-14
Fix large sequence lastValue conversion to 32-bit integer

marc-money:
When testing clone on a database with BIGSERIAL sequences containing `last_value`s that don't fit into 32-bit integers, I encountered overflow/wrapping, e.g.:
```
[TARGET 27517] ERROR:  setval: value -1249415229 is out of bounds for sequence "image_cid_seq"
```
Adjusting the sqlite routine used resolved the issue, and the correct value was set on the target database.


shubhamdhama (approved):
LGTM! It would be also helpful if you can add a test case for this.

dimitri (approved):

PR #645, opened 2024-01-19
Fix database name identifier in ALTER DATABASE properties using PQescapeIdentifier.

VaibhaveS:
**Logs**

```
2024-01-18 14:10:58 53 ERROR  pgsql.c:1978              [TARGET 1664] ERROR:  syntax error at or near "-"
2024-01-18 14:10:58 53 ERROR  pgsql.c:1989              [TARGET 1664] SQL query: ALTER DATABASE vaibhave-test SET "search_path" TO '$user', 'public';
```

### DIFF
--- a/src/bin/pgcopydb/dump_restore.c
+++ b/src/bin/pgcopydb/dump_restore.c
@@ -18,7 +18,6 @@
 #include "env_utils.h"
 #include "lock_utils.h"
 #include "log.h"
-#include "pgsql.h"
 #include "pidfile.h"
 #include "schema.h"
 #include "signals.h"
--- a/src/bin/pgcopydb/dump_restore.c
+++ b/src/bin/pgcopydb/dump_restore.c
@@ -281,11 +281,12 @@ copydb_copy_database_properties_hook(void *ctx, SourceProperty *property)
 	PGSQL *dst = context->dst;
 	PGconn *conn = dst->connection;
 
-	const char *t_pguri = specs->connStrings.safeTargetPGURI.uriParams.dbname;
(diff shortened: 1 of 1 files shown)


dimitri (approved):

src/bin/pgcopydb/dump_restore.c
```diff
@@ -281,7 +281,13 @@ copydb_copy_database_properties_hook(void *ctx, SourceProperty *property)
...
 	PGconn *conn = dst->connection;
 
-	const char *t_dbname = specs->connStrings.safeTargetPGURI.uriParams.dbname;
+	const char *t_pguri = specs->connStrings.safeTargetPGURI.uriParams.dbname;
+	const char *t_dbname = pgsql_escape_identifier(dst, t_pguri);
+
+	if (t_dbname == NULL)
+	{
+		return false;
+	}
```
dimitri, line 290: I think it's worth adding a log message about the situation actually. Because we don't expect it to ever happen, something like we have in similar unexpected places:

```
    log_error("BUG: failed to escape identifier \"%s\"", t_pguri);
```

Oh and also we went to fast with the naming of the intermediate variable. That's not a Postgres connection string (pguri), it's a database name... Please review naming!

dimitri (changes requested):

VaibhaveS:
Looks like the `filtering` is failing randomly.

dimitri (approved):

dimitri:
> Looks like the `filtering` is failing randomly.

That's the SQLite `PRAGMA journal_mode = WAL failed: database is locked` problem that we need to look into separately, unrelated to your changes.

PR #646, opened 2024-01-21
Improve sqlite catalog setup performance.

shubhamdhama:
Intentionally this PR contains two independent commits to achieve a single
goal mentioned in the title

1. Use appropriate sqlite PRAGMA commands which significantly improved the
   performance. Code comments are self explanatory.
2. Batched writes within one big transaction.

**Results**

For a setup of 13K+ tables below is the time taken by pgcopydb list tables
command excluding the time of running postgres query on source.

WAL => `PRAGMA journal_mode = WAL`
Normal => `PRAGMA synchronous = NORMAL`
Txn => `BEGIN`/`COMMIT`

No optimization | WAL | WAL + Normal | WAL + Normal + Txn | WAL + Txn
-|-|-|-|-
1h 18 mins | 8 mins | 53s | 18s | 18s

So finally decided to go with `journal_mode` as WAL along with one big insert transaction.

### DIFF
--- a/src/bin/pgcopydb/catalog.c
+++ b/src/bin/pgcopydb/catalog.c
@@ -732,28 +732,17 @@ catalog_init(DatabaseCatalog *catalog)
 	}
 
 	/*
-	 * Because the source catalog also registers the activity information,
-	 * which generates a non-trivial amount of small writes with some level of
-	 * concurrency, we turn the WAL mode on that one.
-	 *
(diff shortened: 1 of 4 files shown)


dimitri (approved):

dimitri:
Thanks a lot @shubhamdhama for the investigation and the PR!

PR #652, opened 2024-01-25
Overhaul of EditorConfig rules

hanefi:
Before this change we had 153 files on the main branch that did broke at least one of our editorconfig rules. After this change, all of our files honor the rules that are partially updated.

There are 2 directories that are now exempt from our rules:
- external libraries
- debian directory


dimitri:
Is there a command line tool that would apply editor config to all the files, ala `make indent`? Otherwise I'm not sure my Emacs setup will be able to apply the current set, in particular the bash changes look weird...

hanefi:
> Is there a command line tool that would apply editor config to all the files, ala `make indent`?

There are several alternatives. I tried 3 most popular implementations, and I do not think that I can suggest someone to install them as they have their quirks.

1. https://github.com/jednano/eclint/
   - archived repo
   - last release was 5 years ago
   - can infer rules from codebase
   - can fix problems in files 
   - the npm package is installed 19k times last week.
2. https://gitlab.com/greut/eclint
   - can fix problems in files
   - I never liked having a single dash in arguments ( e.g. `-fix` instead of `--fix` )
3. https://github.com/editorconfig-checker/editorconfig-checker
   - can not fix problems in files

> Otherwise I'm not sure my Emacs setup will be able to apply the current set, in particular the bash changes look weird...

I was not able to come up with a rule that worked for our sql files as well. Maybe I can add make shell scripts exempt from indentation rules just like what I had for sql files.

dimitri:
> I was not able to come up with a rule that worked for our sql files as well. Maybe I can add make shell scripts exempt from indentation rules just like what I had for sql files.

I think you've done that in the current version of the PR, right? looks better to me already

hanefi:
> > I was not able to come up with a rule that worked for our sql files as well. Maybe I can add make shell scripts exempt from indentation rules just like what I had for sql files.
> 
> I think you've done that in the current version of the PR, right? looks better to me already

Yes, that's right. I expect the editorconfig to ignore shell scripts except for adding a single line ending character at the end, etc.

tests/extensions/countries.sql
```diff
@@ -469,4 +469,3 @@ COPY capitals (id, name, country_id, geom) FROM stdin;
 --
 -- PostgreSQL database dump complete
 --
-
```
dimitri, line 472: Not our file, is it possible to avoid processing it?

src/bin/lib/parson/tests/test_2.txt
```diff
@@ -1,4 +1,4 @@
-﻿{
+{
```
dimitri, line 1: Not our file, can you remove that change from the PR?

tests/extensions/countries.sql
hanefi: I did not know about this file being from an outside source. Thanks for bringing this to my attention.

I added a new .editorconfig in this directory to have an exclusion for this file. All the other files in the same directory are still affected by the root .editorconfig.

src/bin/lib/parson/tests/test_2.txt
hanefi: This was my mistake. I verified that this PR now has the minimal set of changes required to pass the current editorconfig rules.

`src/bin/lib/.editorconfig` file allows us to ignore all the files under `/src/bin/lib` now.

I removed this change from the PR.

PR #653, opened 2024-01-25
Introduce CONTRIBUTING.md

hanefi:
I used a copy of Citus' CONTRIBUTING.md as a starting point.

I removed sections that are not relevant to pgcopydb, and added some more sections about documentation and running tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,82 @@
+# Contributing to pgcopydb
+
+We're happy you want to contribute! You can help us in different ways:
+
+* Open an [issue](https://github.com/dimitri/pgcopydb/issues) with suggestions
+  for improvements, potential bugs, etc.
+* Fork this repository and submit a pull request
+
+### Building from source code
+
+Follow the relevant docs at our
+[documentation](https://pgcopydb.readthedocs.io/en/latest/install.html#build-from-sources).
+
+### Following our coding conventions
+
+CI pipeline will automatically reject any PRs which do not follow our coding
+conventions. The easiest way to ensure your PR adheres to those conventions is
+to use the
+[citus_indent](https://github.com/citusdata/tools/tree/develop/uncrustify) tool.
+This tool uses `uncrustify` under the hood.
+
+
+```bash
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+We're happy you want to contribute! You can help us in different ways:
+
+* Open an [issue](https://github.com/dimitri/pgcopydb/issues) with suggestions
+  for improvements, potential bugs, etc.
+* Fork this repository and submit a pull request
+
+### Building from source code
+
+Follow the relevant docs at our
+[README.md](https://github.com/dimitri/pgcopydb/#installing-pgcopydb)
```
dimitri, line 12: I think the README should be as short as possible, and now that we have build/install coverage in the main documentation at https://pgcopydb.readthedocs.io/en/latest/install.html#build-from-sources I'm wondering if we should remove the section from the README, and in this PR point to the docs. What do you think?

CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+This tool uses `uncrustify` under the hood.
+
+
+```bash
+# Use your favorite package manager, and install `uncrustify` by running one of
+# the following
+apt-get install uncrustify
+brew install uncrustify
+apk add --upgrade uncrustify
+dnf install uncrustify # may require epel-release rpm
```
dimitri, line 29: Let's avoid showing the alternatives within a single bash script, (some) people will inevitably copy/paste the whole thing and wonder about it. Might even report an issue...

dimitri (changes requested):
Thanks a lot Hanefi, we need such a file!

CONTRIBUTING.md
hanefi: I read through that piece of the README, and realized that yum installation is outdated for a while. I think it is a good idea to remove installation instructions completely, and add a reference to our docs.

I also took the liberty of adding a reference in our docs for building from source to CONTRIBUTING.md.

dimitri (approved):

PR #655, opened 2024-01-26
Refactor pgcopydb for standby server support

gokhangulbiz:
This pull request refactors the pgcopydb to add support for running on a standby server. It includes changes to the copydb_export_snapshot() function to handle standby server mode by setting isolation level and readOnly mode properly and creates a new test case named pagila-standby in the GHA workflow.

Please note that, with this PR, `pgcopydb` skips partitioning on standby servers (where pg_is_in_recovery() is true) because it cannot execute `ANALYZE` on a read-only database.


src/bin/pgcopydb/snapshot.c
```diff
@@ -79,6 +73,30 @@ copydb_export_snapshot(TransactionSnapshot *snapshot)
...
+		/* errors have already been logged */
+		(void) pgsql_finish(pgsql);
+		return false;
+	}
+
+	if (pg_is_in_recovery)
+	{
+		level = ISOLATION_REPEATABLE_READ;
+		readOnly = true;
+	}
```
dimitri, line 92: Do we need actually ISOLATION_SERIALIZABLE for normal operations on a primary server? I mean, it might be that we only ever need ISOLATION_REPEATABLE_READ in all cases?

dimitri (commented):
Thanks for this work @gokhangulbiz !

I see the test is failing because of the pgcopydb table size table creation. I think we could also just run a SELECT query to fetch the information and populate our SQLite catalogs with the result of it, in a new table `s_table_size` on the source catalogs. This would need to be done in yet another PR.

tests/pagila-standby/copydb.sh
```diff
@@ -0,0 +1,35 @@
...
+
+grep -v "OWNER TO postgres" /usr/src/pagila/pagila-schema.sql > /tmp/pagila-schema.sql
+
+psql -o /tmp/s.out -d ${PGCOPYDB_SOURCE_PGURI} -1 -f /tmp/pagila-schema.sql
+psql -o /tmp/d.out -d ${PGCOPYDB_SOURCE_PGURI} -1 -f /usr/src/pagila/pagila-data.sql
+
+pgcopydb clone --skip-ext-comments --notice \
+         --source ${PGCOPYDB_SOURCE_STANDBY_PGURI} \
+         --target ${PGCOPYDB_TARGET_PGURI}
+
```
arajkumar, line 28: Can we test CDC(i.e. --follow) as well?

dimitri: Not yet, because that requires Postgres 16. Let's first focus on the clone operations, and when we have that in we will be in a great position to see about supporting follow operations too.

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -721,9 +710,13 @@ copydb_prepare_table_specs_hook(void *ctx, SourceTable *source)
 	 * When the Table Access Method used is not "heap" we don't know if the
 	 * CTID range scan is supported (see columnar storage extensions), so
 	 * we skip partitioning altogether in that case.
+	 *
+	 * Also, we cannot execute ANALYZE on a database that is in recovery mode,
+	 * so we skip partitioning in that case too.
```
dimitri, line 715: I think we should assume that the statistics are up-to-date when hitting a read-only source database, or at least have support for COPY partitioning in that case. An option to disable COPY partitioning would then be good to have for users who want to opt-out of single-table concurrency. This option would be useful in more cases.

dimitri (changes requested):
This now needs a rebase on-top of the current main branch.

dimitri (changes requested):

dimitri (approved):

PR #658, opened 2024-01-29
Fix confusion around reporting operations in progress.

dimitri:
The SQLite refactor meant rewriting all the SourceTableArray loops to use a query result iterator instead. Well there is a place for a SourceTableArray loop still, and that's the pgcopydb list progress thing.

### DIFF
--- a/src/bin/pgcopydb/catalog.c
+++ b/src/bin/pgcopydb/catalog.c
@@ -6911,9 +6911,7 @@ catalog_iter_s_table_in_copy_init(SourceTableIterator *iter)
 		"  select t.oid, qname, nspname, relname, amname, restore_list_name, "
 		"         relpages, reltuples, t.bytes, t.bytes_pretty, "
 		"         exclude_data, part_key, "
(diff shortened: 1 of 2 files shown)


shubhamdhama (changes requested):
It still has a problem majorly here,
https://github.com/dimitri/pgcopydb/blob/ad8f20ec06aa1c843c1ea5553f19ab1e6bbef43e/src/bin/pgcopydb/catalog.c#L6910-L6930

The `sum` functions are ran over all the rows and resulting a single row. Instead we should run it over groups by table oid or don't run sum at all if we are showing all parts.

dimitri:
> The `sum` functions are ran over all the rows and resulting a single row. Instead we should run it over groups by table oid or don't run sum at all if we are showing all parts.

Good finding! That might be why the query always returns a single row even when there is no table copy currently in progress. The solution looks indeed like adding a GROUP BY clause to the query, but here we want to track COPY progress, which is per-partition, not per-table. Looking at the consuming part of the code in `progress.c` I realize that we don't use that extra information at the moment anyway, so I simplified by removing it.

PR #660, opened 2024-01-31
Vendor SQLite amalgamation distribution.

dimitri:
This allows to ensure we're always using a known version of the SQLite lib and allows pgcopydb code to use some "recent" features on all OS targets.

Such features include the RETURNING clause for DMLs, the FILTER clause for aggregate function calls, and time formatting function or JSON functions.

See https://sqlite.org/amalgamation.html for details about that distribution mode for SQLite.


shubhamdhama:
@dimitri how about putting sqlite based code in a separate repo (fork) and add it as a submodule here?

Issue #426, opened 2023-08-15
syntax error during follow process


ulo23k:
I am currently using the latest commit and running into a syntax error.

```
2023-08-15 06:09:13 5373 ERROR  pgsql.c:1913              [TARGET 20344] ERROR:  syntax error at or near "WHERE"
2023-08-15 06:09:13 5373 ERROR  pgsql.c:1913              [TARGET 20344] LINE 1: UPDATE "public"."device_credentials" SET  WHERE "device_id" ...
2023-08-15 06:09:13 5373 ERROR  pgsql.c:1913              [TARGET 20344]                                                   ^
2023-08-15 06:09:13 5373 ERROR  ld_apply.c:576            Failed to apply SQL from file "/home/ssm-user/.local/share/pgcopydb/00000001000016C600000023.sql", see above for details
2023-08-15 06:09:13 5370 ERROR  follow.c:902              Subprocess catchup with pid 5373 has exited with error code 12
2023-08-15 06:09:13 5371 INFO   pgsql.c:4309              Reported write_lsn 16C6/F440E1C0, flush_lsn 16C6/F440E1C0, replay_lsn 16C6/88E34FA8
2023-08-15 06:09:13 5371 INFO   ld_stream.c:455           Streamed up to write_lsn 16C6/F440E1C0, flush_lsn 16C6/F440E1C0, stopping
2023-08-15 06:09:13 5370 ERROR  follow.c:581              Some sub-process exited with errors, see above for details
2023-08-15 06:09:13 5370 ERROR  follow.c:288              Failed to follow changes from source, see above for details
2023-08-15 06:09:14 5365 ERROR  cli_clone_follow.c:722    follow process 5370 has terminated [12]
```

I took a look at the query it was complaining about and its quite normal

```
BEGIN; -- {"xid":942370231,"lsn":"16C6/8C0000D0","timestamp":"2023-08-15 03:59:30.290077+0000","commit_lsn":"16C6/8C0021A8"}
PREPARE 8766489e AS UPDATE "public"."device_credentials" SET "accessed_at" = $1 WHERE "device_id" = $2 and "pin" = $3 and "sally_id" = $4 and "failures" = $5 and "pin_token" = $6 and "product" = $7 and "accessed_at" = $8 and "created_at" = $9;
[...]

PR #664, opened 2024-02-02
Support CDC on tables with generated columns

arajkumar:
During the CDC (i.e. pgcopydb clone --follow), replication of INSERT/DELETE on tables with generated column fails following error log.

> 11:53:41.475 94 ERROR  [TARGET 197718] [428C9] ERROR:  cannot insert a non-DEFAULT value into column "term_vector"
11:53:41.475 94 ERROR  [TARGET 197718] DETAIL:  Column "term_vector" is a generated column.

**Solution**: Find tables with generated column from the sqlite schema catalog and use `DEFAULT` as value on those columns instead of the value from logical message.

Accessing the SQLite catalog for each DML message might incur performance overhead. To avoid this, the transform process prepares an in-memory cache using hash table to hold all the tables with generated columns during the initialization of the transform. Generally, we don't expect to have many tables with generated columns, so keeping them in memory is a better trade-off and avoids performance bottlenecks.

Related PR which skips generated column for table-data copy: https://github.com/dimitri/pgcopydb/pull/517


arajkumar:
I see `table missing in catalog error` due to quoting added by https://github.com/dimitri/pgcopydb/pull/595. i.e. the catalog doesn't escape the nspname, relname; but it has been escaped for wal2json by https://github.com/dimitri/pgcopydb/pull/595.

src/bin/pgcopydb/catalog.c
```diff
@@ -2339,7 +2339,8 @@ catalog_lookup_s_table_by_name(DatabaseCatalog *catalog,
 		"           group by oid"
 		"         ) p "
 		"        on p.oid = t.oid"
-		"   where nspname = $1 and relname = $2 ";
+		"   where trim(nspname, '\"') = trim($1, '\"') "
+		"   and trim(relname, '\"') = trim($2, '\"') ";
```
arajkumar, line 2343: @dimitri I'm not sure how it scales well when we have lots of entries in catalog., Should we consider creating an index for nspname + relname?

dimitri: When using wal2json I believe we have the table OID available to us, we should probably use that. I didn't understand if the problem you're trying to solve is just in wal2json or both with test_decoding too?

dimitri (changes requested):
Lots of changes in ld_transform and I'm not sure why... do we really have to change all the internal function signatures in the context of this bugfix?

arajkumar:
> Lots of changes in ld_transform and I'm not sure why... do we really have to change all the internal function signatures in the context of this bugfix?

@dimitri We need privateContext to get access to catalog from all the functions which generates SQL statement. I think the alternate approach can be accessing the catalog from ld_wal2json.c & ld_test_decoding.c, ignore the generated column while transforming to the data object.

src/bin/pgcopydb/catalog.c
arajkumar: I'm trying to solve this for both wal2json & test_decoding.

dimitri:
> I think the alternate approach can be accessing the catalog from ld_wal2json.c & ld_test_decoding.c, ignore the generated column while transforming to the data object.

This sounds like it would make more sense to me yes, and as you're highlighting the ld_test_decoding module already has to implement some SQLite lookup for attribute properties. Also, again, this would allow skipping the quoting issue from wal2json for nspname and relname altogether, because I think we have the table OID available there already, right?

arajkumar:
@dimitri 

> > I think the alternate approach can be accessing the catalog from ld_wal2json.c & ld_test_decoding.c, ignore the generated column while transforming to the data object.
> 
> This sounds like it would make more sense to me yes, and as you're highlighting the ld_test_decoding module already has to implement some SQLite lookup for attribute properties. Also, again, this would allow skipping the quoting issue from wal2json for nspname and relname altogether, because I think we have the table OID available there already, right?

@dimitri I couldn't find table oid in wal2json, Do you mean wal2json has an option/setting to include table oid as part of the message? I do see `include-type-oids`, but that is only for column types, doesn't include table oid.

src/bin/pgcopydb/defaults.h
```diff
@@ -16,6 +16,8 @@
 #define calloc(m, n) GC_malloc((m) * (n))
 #define free(p) GC_free(p)
 #define realloc(p, n) GC_realloc((p), (n))
+#define strdup(p) GC_strdup(p)
+#define strndup(p, n) GC_strndup(p, n)
```
arajkumar, line 20: @dimitri I didn't see the strdup being overridden to use GC_ variant. Was there a specific reason?

dimitri: No reason other than it was missed, please can you make a separate PR for that?

arajkumar: @dimitri FYI https://github.com/dimitri/pgcopydb/pull/691

arajkumar:
@dimitri It would be useful if can review this. Thank you.

dimitri:
> @dimitri It would be useful if can review this. Thank you.

I should be able to review that early next week now. The work I'm doing to refactor the JSON/SQL storage and the transform and replay services around using an internal file format that allow random access (thanks SQLite!) might have some intersection with how you're implementing this PR. I have to have that in mind at review time, next week.

arajkumar:
@dimitri, Did you find some time to review this? Thank you.

dimitri (commented):
I fail to understand why we build a new in-memory hash-table for only generated columns instead of doing a SQLite lookup per column at transform time?

arajkumar:
@dimitri Please take a look. thank you.

src/bin/pgcopydb/catalog.c
```diff
@@ -3143,6 +3146,190 @@ catalog_iter_s_table_part_finish(SourceTablePartsIterator *iter)
 }
 
 
+/*
+ * catalog_iter_s_generated_column_next fetches the next GeneratedColumn entry
+ * in our catalogs.
+ */
+bool
+catalog_iter_s_generated_column_next(GeneratedColumnIterator *iter)
+{
```
dimitri, line 3155: Nitpicking: the iterators are all declared in the same ordering (iter, init, next, fetch, finish, if I remember correctly). Could you follow the convention here?

src/bin/pgcopydb/ld_stream.h
```diff
@@ -275,6 +275,26 @@ typedef enum
 } LogicalStreamMode;
 
 
+/*
+ * Fully Qualified Postgres column name: "nspname"."relname"."attname".
+ * We need to account for the dots hence add 2 more bytes.
+ */
+typedef char FQColumnName[(PG_NAMEDATALEN * 3) + 2];
```
dimitri, line 282: I think we do not need that. See the docs for UTHASH support for both [compound keys](https://troydhanson.github.io/uthash/userguide.html#_compound_keys) and also for [structure keys](https://troydhanson.github.io/uthash/userguide.html#_structure_keys).

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2653,268 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+ * Here is an example:
+ * foo -> "foo"
+ * "foo" -> "foo"
+ * foo"bar -> "foo"bar"
+ * "foo -> ""foo"
+ *
+ * The goal of this normalization is to make sure that the identifiers are
+ * comparable in the context of Hash Table.
+ */
+#define NORMALIZE_IDENTIFIER(buf, bufsize, offset, ident) \
```
dimitri, line 2677: Not needed if we use compound or structure keys from our hash lib.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2653,268 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+	return isGenerated;
+}
+
+
+/*
+ * populateGeneratedColumnCache is a callback function that populates the
+ * generated columns cache from the catalog.
+ */
+static bool
+populateGeneratedColumnCache(void *ctx, GeneratedColumn *column)
```
dimitri, line 2765: Nitpicking: all the callback function names ends the same way, with `_hook` as a postfix. Please consider doing the same here.

dimitri (changes requested):
This seems to be based on the SOURCE metadata/catalogs, meaning that it will only work when the source and target schema are the same. This can be a nice first step towards solving that problem, but then I believe we should add a note of it in our documentation of the limitations of our CDC implementation.

src/bin/pgcopydb/ld_transform.c
arajkumar: Done.

src/bin/pgcopydb/ld_stream.h
arajkumar: Done.

src/bin/pgcopydb/catalog.c
arajkumar: Done

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2653,269 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+		/* Only INSERT and UPDATE statements might have generated columns */
+		return true;
+	}
+
+	if (columns->count > 1)
+	{
+		log_error("BUG: removeGeneratedColumnsFromStatement called on a statement "
+				  "with multiple VALUES clauses");
+		return false;
+	}
```
dimitri, line 2868: Does that mean we don't have support for multi-values INSERTS at all anymore? when does that test kicks-in really? That needs at least a good comment about the situation.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2653,269 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+
+	LogicalMessageTuple *tuple = &(columns->array[0]);
+
+	for (int c = 0; c < tuple->cols; c++)
+	{
+		if (isGeneratedColumn(cache,
+							  nspname,
+							  relname,
+							  tuple->columns[c]))
+		{
```
dimitri, line 2878: Is there a way to bypass walking through the list of attribute of a table for every INSERT and UPDATE operation we want to transform? Maybe another level of caching to know if a table has any generated columns?

dimitri (changes requested):
Please see detailed review below. In short: 
  - can we add the regression testing to already existing test suites?
  - can we avoid looping through all columns and instead cache the list of tables that have generated columns?
  - let's rename the GeneratedColumn data structure, it's just a fully qualified column name that one

src/bin/pgcopydb/catalog.c
```diff
@@ -3143,6 +3146,190 @@ catalog_iter_s_table_part_finish(SourceTablePartsIterator *iter)
...
+	{
+		log_error(ALLOCATION_FAILED_ERROR);
+		return false;
+	}
+
+	char *sql =
+		"  select nspname, relname, attname "
+		"    from s_attr a"
+		"      join s_table t on t.oid = a.oid "
+		"   where a.attisgenerated";
```
dimitri, line 3176: Is there a distinction to be made between GENERATED ALWAYS and GENERATED BY DEFAULT? (which only is for IDENTITY columns, if I read the docs correctly).

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2653,269 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+			{
+				tuple->columns[j] = tuple->columns[j + 1];
+			}
+
+			/* number of colums have reduced now */
+			tuple->cols--;
+
+			/* remove the value from the tuple */
+			for (int r = 0; r < tuple->values.count; r++)
+			{
```
dimitri, line 2893: We could keep the column in there and replace the value with the DEFAULT keyword, I believe. Does that work? does it behave better in terms of performance characteristics? Is it easier to debug?

dimitri (commented):
Some aspects of the PR needs more comments/docs/explaining and maybe there is a better way to approach the replacement of the given values, using the DEFAULT keyword?

src/bin/pgcopydb/ld_transform.c
arajkumar: Yeah, it was one of the candidates. But, since we pass the EXEC parameters as a JSON string, we may not be able to distinguish between a DEFAULT keyword and DEFAULT as a string unless we make changes to the way we handle it.

.github/workflows/run-tests.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           - follow-9.6
           - follow-data-only
           - endpos-in-multi-wal-txn
+          - cdc-generated-column
```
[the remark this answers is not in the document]
arajkumar: I can do that, but I afraid we will end up duplicating the tests logics.

src/bin/pgcopydb/ld_transform.c
arajkumar: I think we can use a special marker object to encode DEFAULT keyword, something like below,

> EXECUTE 7fa747a0["20","2024-06-04 11:41:19.279866+00", {"keyword": "DEFAULT"}];

dimitri: I don't know if keeping the number of attributes is important enough to warrant that change. What I like about it is that we don't need to edit the columns and values arrays (run-time performance characteristics should be better), and it makes it easier to debug too. How does it look like in terms of complexity? do you think such a change is a good idea, and that it's worth it?

.github/workflows/run-tests.yml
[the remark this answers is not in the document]
arajkumar: Done

src/bin/pgcopydb/catalog.c
```diff
@@ -8144,3 +8147,108 @@ catalog_stop_timing(TopLevelTiming *timing)
 						 INTSTRING_MAX_DIGITS);
 	}
 }
+
+
+/*
+ * catalog_iter_s_table_attisgenerated iterates over the list of tables that
```
dimitri, line 8153: nitpicking: function name is wrong now

src/bin/pgcopydb/ld_stream.h
```diff
@@ -275,6 +275,42 @@ typedef enum
...
+{
+	/* The table which has generated columns */
+	char nspname[PG_NAMEDATALEN];
+	char relname[PG_NAMEDATALEN];
+} TableWithGeneratedColumns;
+
+
+typedef struct GeneratedColumns
+{
+	char column[PG_NAMEDATALEN];
```
dimitri, line 292: What about using `attname` here?

src/bin/pgcopydb/ld_stream.h
```diff
@@ -275,6 +275,42 @@ typedef enum
...
+ * This is a multi-level hash table, where the first level is the table which
+ * has generated columns and the second level is the column name itself.
+ */
+typedef struct GeneratedColumnsCache
+{
+	/* The key to lookup the generated columns */
+	TableWithGeneratedColumns table;
+
+	/* The generated columns cache */
+	GeneratedColumns *columns;
```
dimitri, line 308: Is this an array or a linked list or something else? I think the comments should say.

src/bin/pgcopydb/ld_stream.h
```diff
@@ -309,6 +345,9 @@ typedef struct StreamContext
 	/* transform needs some catalog lookups (pkey, type oid) */
 	DatabaseCatalog *sourceDB;
 
+	/* hash table acts as a cache for tables with generated columns */
+	GeneratedColumnsCache *generatedColumnsCache;
```
dimitri, line 349: Why is this a pointer? Is this an array? a linked-list? a single item that is malloced later?

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2650,323 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+	if (item->columns == NULL)
+	{
+		log_error("BUG: Table \"%s.%s\" is in the cache but columns are NULL",
+				  nspname, relname);
+		return false;
+	}
+
+	*columns = item->columns;
+
+	log_debug("Table \"%s.%s\" has generated columns", nspname, relname);
```
dimitri, line 2733: I suppose this would be `log_trace` if we think we want to keep it around in the merge?

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2650,323 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+
+	NORMALIZE_IDENTIFIER(column, attname, PG_NAMEDATALEN);
+
+	GeneratedColumns *generatedColumns = NULL;
+
+	HASH_FIND_STR(columns, column, generatedColumns);
+
+	if (generatedColumns != NULL)
+	{
+		log_debug("Column \"%s\" is generated", attname);
```
dimitri, line 2759: log_trace?

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2650,323 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+	return false;
+}
+
+
+/*
+ * generated_column_iter_hook is a callback function that populates the
+ * generated columns cache from the catalog.
+ */
+static bool
+generated_column_iter_hook(void *ctx, SourceTable *table)
```
dimitri, line 2772: The name should be prepareGeneratedColumnsCache_hook I believe.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2650,323 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+												calloc(1, sizeof(GeneratedColumns));
+			if (generatedColumn == NULL)
+			{
+				log_error(ALLOCATION_FAILED_ERROR);
+				return false;
+			}
+
+			NORMALIZE_IDENTIFIER(generatedColumn->column,
+								 attr->attname, PG_NAMEDATALEN);
+			HASH_ADD_STR(item->columns, column, generatedColumn);
```
dimitri, line 2813: We now have a hash-table per table with generated column. I believe the previous design had a single hash-table with all the generated columns in there, and I believe I prefer that option. How could you use the hash-table API that we have to maintain a single hash-table that would contain `nspname.relname` and also another single hash-table that would contain `nspname.relname.attname`?

dimitri: Or maybe the hash-table overhead is much less compared to having to repeat the nspname and relname in memory for each generated column?

tests/follow-data-only/copydb.sh
```diff
@@ -83,3 +83,18 @@ psql -d ${PGCOPYDB_SOURCE_PGURI} -c "${sql}" > /tmp/s.out
...
 diff /tmp/s.out /tmp/t.out
+
+# count the number of rows in the generated_column_test table
+# on the source and target databases
+sql="select count(*) from generated_column_test"
+src_count=`psql -AtqX -d ${PGCOPYDB_TARGET_PGURI} -c "${sql}"`
+tgt_count=`psql -AtqX -d ${PGCOPYDB_SOURCE_PGURI} -c "${sql}"`
+test ${src_count} -gt 0
+test ${tgt_count} -gt 0
+test ${src_count} -eq ${tgt_count}
```
dimitri, line 94: Can we use output files and then `diff`? When the test fail, it makes it easier to debug.

dimitri (changes requested):

src/bin/pgcopydb/ld_stream.h
arajkumar: @dimitri, it is a hash table as explained in the comment above.

src/bin/pgcopydb/ld_transform.c
arajkumar: @dimitri, It depends on the nature of the schema. We can compromise both of the implementation using micro benchmarks. For e.g. 10 tables with each having 3 generated columns each vs 3 tables having 10 generated columns each.

I prefer the current approach, i.e. one hash table to find `nspname.relname` and another hash table to find `attname`.

src/bin/pgcopydb/ld_stream.h
dimitri: Yeah but a hash-table over what? We might have more than one column per table, and we have a single pointer. What do we point to? Certainly the hash-table has one entry per column name, and also points to somewhere... where does it point to?

src/bin/pgcopydb/ld_transform.c
dimitri: I can live with that too, but then it needs to be explained in the comments why that choice was made, and how we did conclude that the “one hash-table per table with generated columns” is an acceptable overhead.

src/bin/pgcopydb/ld_transform.c
arajkumar: @dimitri It has been simplified now and uses DEFAULT as a value instead of removing the column.

src/bin/pgcopydb/catalog.c
```diff
@@ -8144,3 +8147,108 @@ catalog_stop_timing(TopLevelTiming *timing)
...
+ * catalog of SourceTable entries which has generated columns.
+ */
+bool
+catalog_iter_s_table_generated_columns_init(SourceTableIterator *iter)
+{
+	sqlite3 *db = iter->catalog->db;
+
+	if (db == NULL)
+	{
+		log_error("BUG: Failed to initialize s_table iterator: db is NULL");
```
dimitri, line 8218: Function name is wrong here.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2669,297 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+			GeneratedColumnSet *generatedColumn = (GeneratedColumnSet *)
+												  calloc(1, sizeof(GeneratedColumnSet));
+			if (generatedColumn == NULL)
+			{
+				log_error(ALLOCATION_FAILED_ERROR);
+				return false;
+			}
+
+			NORMALIZED_PG_NAMEDATA_COPY(generatedColumn->attname, attr->attname);
+			HASH_ADD_STR(item->columns, attname, generatedColumn);
```
dimitri, line 2833: I think I see now. I am used to building a hash-table over an array or a linked-list, so that I would have a sequential path and a direct lookup path. Here you're saying we only need the hash-table lookup path to access the data, so that we don't even need to maintain a linked-list or anything.

src/bin/pgcopydb/ld_stream.h
dimitri: Same as the other comment, there is never a need for a sequential access to the data allocated, so all we have to access the malloc'ed areas is the hash-table.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2669,297 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+}
+
+
+/*
+ * isGeneratedColumn checks whether the given "attname" is a generated column.
+ *
+ * Returns true if the column is generated, false otherwise.
+ *
+ * NOTE: There is no error condition, if the cache is NULL, it means that we
+ * don't have any generated columns in the catalog.
```
dimitri, line 2766: Do we still need to do a hash lookup if the cache (`*columns`) is null here? is it safe to use the API that way?

dimitri (changes requested):
We're onto nitpicking only comments now, almost ready to merge. Thanks for exploring the DEFAULT idea, it looks much simpler now thanks to that.

dpunkturban:
Hi @arajkumar, 

Thanks for your PR. I was hitting these problems and searching for a solution. Just to be sure, does your MR also support `pgcopydb clone --follow`? I tried to sync with your latest commit on your branch, but I encountered another error. It's not working for me. It always fails on the auto-generated column based on `tsvector` when I add a row where a tsvector auto-generated column is included.

Error with latest version (0.16) from dimitri:
```
11:53:38.381 94 INFO   Replaying changes from LSN 0/32D4538
11:53:41.475 94 ERROR  [TARGET 197718] [428C9] ERROR:  cannot insert a non-DEFAULT value into column "term_vector"
11:53:41.475 94 ERROR  [TARGET 197718] DETAIL:  Column "term_vector" is a generated column.
11:53:41.475 94 ERROR  [TARGET 197718] Context: Failed to receive pipeline sync
free(): double free detected in tcache 2
```

Error with your latest commit: 
```
11:46:34.645 349 INFO   Replaying changes from LSN 0/32B3B40
11:46:40.892 349 ERROR  [TARGET 197362] [42601] ERROR:  syntax error in tsvector: "''test'':1"
11:46:40.892 349 ERROR  [TARGET 197362] CONTEXT:  unnamed portal parameter $13 = '...'
11:46:40.892 349 ERROR  [TARGET 197362] Context: Failed to receive pipeline sync
```

Thanks :)

arajkumar:
Hey, thanks for testing this. Could you please share a simple reproducer ?
Or schema for that table would be helpful too.

dpunkturban:
Hi @arajkumar,

I looked deeper into the errors, and from what I see, it's not coming from your changes. The old "error message" only hides the real error. I cannot reproduce it with normal SQL commands; only if I use prepared statements. This was already reported in #426, so my assumption is that it's not coming from your changes.

Sorry for the noise! :/

arajkumar:
No worries, thanks for the update!

I would still recommend to attach the schema for the problematic table for
the ease of debugging.

src/bin/pgcopydb/ld_transform.c
arajkumar: That is correct, we have a higher level hash-table which contains tables which has generated columns and with in that, a set of columns which are generated(again a hash-table).

src/bin/pgcopydb/ld_transform.c
arajkumar: Fixed it.

src/bin/pgcopydb/catalog.c
arajkumar: Fixed it

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2609,3 +2669,302 @@ LogicalMessageValueEq(LogicalMessageValue *a, LogicalMessageValue *b)
...
+ * generated columns set, false otherwise.
+ *
+ * NOTE: There is no error condition, if the cache is NULL, it means that we
+ * don't have any generated columns in the catalog.
+ */
+static bool
+fetchGeneratedColumnsForTable(GeneratedColumnsCache *cache,
+							  const char *nspname,
+							  const char *relname,
+							  GeneratedColumnSet **columns)
```
dimitri, line 2720: Given the style in the rest of the project, I'm not super happy with this implementation. It looks to me like that caller could have a look at the columns parameter and NULL would indicate that there is no generated columns. We could then use the return value to indicate success, as usual.

Also I think a better name would start with “lookup” because we don't fetch anything to anywhere, we just implement a hash-table lookup, which is in line with the output value being NULL when we don't find anything.

dimitri (changes requested):
See the last detail about the hash-table lookup function name and signature.

Also about the unit tests, can we move the current test from follow-wal2json to cdc-wal2json? I think it would be beneficial to review the JSON and SQL changes that are caused by this PR.

src/bin/pgcopydb/ld_transform.c
dimitri: Thanks for that extra work!

src/bin/pgcopydb/catalog.c
dimitri: ping?

arajkumar: @dimitri Yes, we need to differentiate between IDENTITY vs normal types.

The implementation works only for GENERATED ALWAYS with EXPRESSION.
i.e. `information_schema.columns.is_generated` must be `ALWAYS`.

For the IDENTITY types, `information_schema.columns.is_generated` is always `NEVER`. INSERT works fine with `IDENTITY` types, but UPDATE would fail.

tests/cdc-test-decoding/docker-compose.yml
```diff
@@ -14,7 +14,7 @@ services:
       -c ssl_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
       -c ssl_key_file=/etc/ssl/private/ssl-cert-snakeoil.key
   target:
-    image: postgres:13-bullseye
+    image: postgres:16-bullseye
```
arajkumar, line 17: postgres:13-bullseye has issues when dealing with multi-value insert with generated columns.

e.g.

the following fails on PG13

```
insert into generated_column_test(id, name, greet_hello, email) values
(10, 'Tiger', DEFAULT, 'tiger@wild.com'), (2, 'Tiger', DEFAULT, 'tiger@wild.com');
```

but works!

```
insert into generated_column_test(id, name, greet_hello, email) values
(10, 'Tiger', DEFAULT, 'tiger@wild.com');
```

dimitri: Ok, but we still need/want support for Postgres 13 even on the target side of things. Are you saying that the problem is with how the unit-test is written, or is the problem in how we generate the SQL for the target to apply?

arajkumar: The problem happens with multi value insert + generated column place holder as DEFAULT.

If we want to support PG 13, either we need to stop using DEFAULT place holder for generated columns(i.e. remove generated column from insert statement) or disable multi value insert optimisation.

arajkumar: @dimitri FYI, few tests uses PG16 as target => https://github.com/dimitri/pgcopydb/blob/590e72703dc7e701a5cbd9dd42282e102f7f3a17/tests/follow-wal2json/docker-compose.yml#L19

dimitri: Ideally we would have test coverage for all supported Postgres versions, both as source and target. Having some diversity is better than just using a single Postgres version. Thanks to that we discovered the issue here.

Then sorry about the DEFAULT work you have added, please revert back to actually removing columns and values from the statements.

arajkumar: > or is the problem in how we generate the SQL for the target to apply?

The problem is how we generate the multi value INSERT for the target with generated column. It works fine with we exclude columns if it is generated instead of using DEFAULT as a place holder.

arajkumar: @dimitri How about disabling multi value INSERT optimization if the target is <PG14?

dimitri: Is it possible to selectively do that only when the target table contains generated columns? if yes, then I think that's a good idea.

arajkumar: Other option is to exclude the generated column name and DEFAULT value from INSERT statement based on isGenerated flag. I will explore that.

arajkumar: @dimitri I updated the PR to add a new type called LogicalMessageAttribute which holds column name and attributes (currently it has isgenerated). For insert, the new code discards generated column based on isgenerated, for update it is still uses DEFAULT as value.

src/bin/pgcopydb/ld_stream.h
```diff
@@ -116,17 +116,30 @@ typedef struct LogicalMessageValues
...
 typedef struct LogicalMessageValuesArray
 {
 	int count;
 	int capacity;
 	LogicalMessageValues *array; /* malloc'ed area */
 } LogicalMessageValuesArray;
 
+typedef struct LogicalMessageAttribute
+{
+	char *name; /* malloc'ed area */
```
dimitri, line 129: Maybe too generic? If that's only going to host column/attribute names, maybe it could be `char *attname`?

dimitri (approved):
Small comment only, I like the new shape of this PR. Please consider fixing the name thing before merge...

src/bin/pgcopydb/ld_stream.h
arajkumar: @dimitri Sure. I've updated PR as per your suggestion.

dimitri (approved):

arajkumar:
Thanks @dimitri for your valuable feedback. I’m happy to see this is getting merged.

PR #683, opened 2024-02-26
Add skip tablespaces in restore options.

VaibhaveS:


src/bin/pgcopydb/cli_restore.c
```diff
@@ -174,7 +177,7 @@ cli_restore_schema_getopts(int argc, char **argv)
 		{ "trace", no_argument, NULL, 'z' },
 		{ "quiet", no_argument, NULL, 'q' },
 		{ "help", no_argument, NULL, 'h' },
-		{ NULL, 0, NULL, 0 }
+		{ "no-tablespaces", no_argument, NULL, 'y' },
```
VaibhaveS, line 180: how do we choose the characters?

hanefi: We usually try to pick a letter in the beginning of the most significant word in the long form. You can also pick capital letters.

In this case, my first picks are already taken, `t`, `T`, `s`, `S` to name a few.

I do not really understand the logic behind some of those letters. For example `X` is short for "no-comments".

I guess it is enough to pick a letter that was not used before.

dimitri: In pgcopydb we don't expose short letters to the users, we only maintain long option names. But the getopt_long API in Unix requires a short option. So yeah we try to make life easy on us developers, using short names that help us relate to the long names, but if we can't, anything goes.

dimitri (approved):

PR #685, opened 2024-02-29
Add ARM based OSX support to Makefile

gokhangulbiz:
This is necessary because ARM-based macOS devices have a different default library path (ex: installing libgc via `brew install`) compared to other architectures. By including this line, the Makefile ensures that the necessary libraries are correctly linked for ARM-based macOS devices.


dimitri:
Can we make this addition dependent on actually running macOS, at least, or maybe dependant on the `/opt/homebrew` directory existence?

dimitri (approved):

PR #691, opened 2024-03-04
Problem: Still old variant of strdup has been used

arajkumar:
pgcopydb recently switched to Boehm-Demers-Weiser Garbage Collector[1] and replaced malloc family with GC_malloc and friends. However, strdup is not replaced to GC_strdup.

Solution: Use macros to replace strdup/strndup with it's GC variant.

[1] https://github.com/dimitri/pgcopydb/pull/647


arajkumar:
CC: @dimitri

dimitri (approved):

Issue #692, opened 2024-03-04
Problem:  `pgcopydb clone` fails to resume on docker


arajkumar:
Running pgcopydb inside docker causes trouble while using pid to determine who is owning the table for copying. Docker container will be running on their own pid namespace causing it to allocate deterministic pids. 

> 2024-03-04 17:43:57.496 50 ERROR  table-data.c:831          Failed to start table-data COPY worker for table public.readings (63475),  already being processed by pid 43

Workaround: Run the container with `--pid=host` to make use of host's pid namespace. Otherwise don't run inside docker.

My question is, should we use the pid to determine the COPY owners of table or index object?

dimitri:
I think the problem might be that we fail to clean-up the `process` table in some cases. We might also want to TRUNCATE said table at start-up when resuming operations?

arajkumar:
@dimitri , I believe truncating the table on startup would be simple. Let me attempt a fix. Thank you.

PR #698, opened 2024-03-08
Problem: Resume after interrupted clone always fails

arajkumar:
We use `done_time_epoch` from `timings` to populate `runState`, which helps to skip already processed table sections.

However, we call `summary_start_timing` regardless of whether the section has been already done or not., which causing insertion(insert or replace) of default value to `done_time_epoch` because `summary_start_timing` won't be called for already done section.

Without this fix, while resuming after the interruption pgcopydb would do pre-data restore again leading to table already exists error.

Solution: Check whether the section is already completed using `runState` and exit early without calling `summary_start_timing` for already done section.

Related to https://github.com/dimitri/pgcopydb/issues/692

### DIFF
--- a/src/bin/pgcopydb/summary.c
+++ b/src/bin/pgcopydb/summary.c
@@ -434,7 +434,7 @@ summary_add_table(DatabaseCatalog *catalog, CopyTableDataSpec *tableSpecs)
 	}
 
 	char *sql =
-		"insert into summary(pid, tableoid, partnum, start_time_epoch, command)"
+		"insert or replace into summary(pid, tableoid, partnum, start_time_epoch, command)"
 		"values($1, $2, $3, $4, $5)";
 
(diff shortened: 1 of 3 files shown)


src/bin/pgcopydb/summary.c
```diff
@@ -434,7 +434,7 @@ summary_add_table(DatabaseCatalog *catalog, CopyTableDataSpec *tableSpecs)
 	}
 
 	char *sql =
-		"insert into summary(pid, tableoid, partnum, start_time_epoch, command)"
+		"insert or replace into summary(pid, tableoid, partnum, start_time_epoch, command)"
```
dimitri, line 437: I'm wondering if that could case issues wrt hiding errors.

dimitri (approved):

PR #699, opened 2024-03-10
Problem: interrupted clone resume fails with index relation already exists error

arajkumar:
### How to produce this problem?

Run the following command on a relative large source with quite a few tables with indexes.

```
pgcopydb clone --skip-extension --no-acl --no-owner --table-jobs=1 --index-jobs=1 --notice
```

When you see index creation notice message, interrupt the clone operation using Ctrl + C. You might get log lines like below once interrupted.

```
...
05:46:11.552 3527984 INFO   Skipping sequences, already done on a previous run
05:46:11.557 3527979 NOTICE COPY public.readings
05:46:14.136 3527979 NOTICE COPY public.diagnostics
05:46:14.140 3527983 NOTICE CREATE INDEX IF NOT EXISTS readings_tags_id_time_idx ON public.readings USING btree (tags_id, "time" DESC);
05:46:15.582 3527983 NOTICE CREATE INDEX IF NOT EXISTS readings_time_idx ON public.readings USING btree ("time" DESC);
05:46:15.638 3527979 NOTICE COPY public.tags
05:46:15.681 3527979 NOTICE COPY public.metrics
05:46:15.720 3527982 NOTICE VACUUM ANALYZE public.metrics;
05:46:16.037 3527983 NOTICE CREATE INDEX IF NOT EXISTS diagnostics_tags_id_time_idx ON public.diagnostics USING btree (tags_id, "time" DESC);
^C05:46:16.580 3527983 ERROR  Postgres query was interrupted: CREATE INDEX IF NOT EXISTS diagnostics_tags_id_time_idx ON public.diagnostics USING btree (tags_id, "time" DESC);
05:46:16.580 3527983 ERROR  Failed to create index with oid 65679, see above for details
05:46:16.580 3527983 ERROR  CREATE INDEX worker has been interrupted
05:46:16.680 3527978 ERROR  Sub-process 3527983 exited with code 12
05:46:16.680 3527978 ERROR  Some INDEX worker process(es) have exited with error, see above for details
...
```

Then, try to resume the clone,

```
pgcopydb clone --skip-extension --no-acl --no-owner --table-jobs=1 --index-jobs=1 --notice --resume
```

It fails with following error logs,

```
...
05:46:40.965 3528009 INFO   Skipping table public.tags (65657), already done on a previous run
05:46:40.965 3528009 NOTICE Skipping table public.metrics, already done on a previous run
05:46:40.965 3528009 INFO   Skipping table public.metrics (65569), already done on a previous run
05:46:41.153 3528006 INFO   STEP 10: restore the post-data section to the target database
05:46:41.155 3528006 NOTICE  /usr/bin/pg_restore -f /tmp/pgcopydb/schema/post-out.list -l /tmp/pgcopydb/schema/post.dump
05:46:41.187 3528006 NOTICE Skipping already processed dumpId 6653: INDEX 65672 public readings_tags_id_time_idx tsdbadmin
05:46:41.187 3528006 NOTICE Skipping already processed dumpId 6654: INDEX 65673 public readings_time_idx tsdbadmin
05:46:41.187 3528006 NOTICE Write filtered pg_restore list file at "/tmp/pgcopydb/schema/post-filtered.list"
05:46:41.187 3528006 INFO    /usr/bin/pg_restore --dbname 'postgres://xxxx:36898/tsdb?sslmode=require&keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --single-transaction --no-owner --no-acl --use-list /tmp/pgcopydb/schema/post-filtered.list /tmp/pgcopydb/schema/post.dump
05:46:41.272 3528006 ERROR  pg_restore: error: could not execute query: ERROR:  relation "diagnostics_tags_id_time_idx" already exists
05:46:41.272 3528006 ERROR  Command was: CREATE INDEX diagnostics_tags_id_time_idx ON public.diagnostics USING btree (tags_id, "time" DESC);
05:46:41.273 3528006 ERROR  Failed to run pg_restore: exit code 1
05:46:41.273 3528006 ERROR  Failed to finalize schema on the target database, see above for details
05:46:41.273 3528006 ERROR  Failed to clone source database, see above for details
05:46:41.318 3528001 ERROR  clone process 3528006 has terminated [6]
...
```

### Root cause:

When we interrupt during index creation, index might have been created in the target but failed to update the done status on the sqlite source.db. Since the done status is not updated, the problematic index is not filtered and pg_restore attempts to recreate them.

### Solution:

On resume, let the table-data skip data copy if it is done but still signal the index workers to proceed with creating indexes. Index worker shall ignore if the index is already exists, else creates one.

### Alternatives

- pg_restore post-data can be called with `--clean --if-exists --jobs` which will take care of removing indexes those are created on target, but not updated to sqlite summary table. However, it might do double work and explicit vacuum is needed.
- Instead of using sys v message queue, probably we can use sqlite as a persistent message queue. Once table copy completes, it writes indexes to be created on a sqlite and index workers can pick from sqlite table. A similar approach can be used with in index worker to vacuum worker communication too.

Related to https://github.com/dimitri/pgcopydb/issues/692

### DIFF
--- a/src/bin/pgcopydb/table-data.c
+++ b/src/bin/pgcopydb/table-data.c
@@ -904,30 +904,36 @@ copydb_copy_data_by_oid(CopyDataSpec *specs, PGSQL *src, PGSQL *dst,
 		return false;
 	}
 
+	/*
+	 * Skip only table-data copy when it it has been done already on a previous
+	 * run. We still need to process the indexes, constraints, and vacuum.
+	 * So, signal the index and vacuum workers as usual.
+	 */
 	if (isDone)
 	{
-		log_info("Skipping table %s (%u), already done on a previous run",
+		log_info("Skipping table-data %s (%u), already done on a previous run",
 				 tableSpecs->sourceTable->qname,
 				 tableSpecs->sourceTable->oid);
-		return true;
 	}
-
-	/*
-	 * 1. Now COPY the TABLE DATA from the source to the destination.
-	 */
-	if (!table->excludeData)
+	else
 	{
-		if (!copydb_copy_table(specs, src, dst, tableSpecs))
+		/*
+		 * 1. Now COPY the TABLE DATA from the source to the destination.
+		 */
+		if (!table->excludeData)
+		{
+			if (!copydb_copy_table(specs, src, dst, tableSpecs))
+			{
+				/* errors have already been logged */
+				return false;
+			}
+		}
+
+		if (!copydb_mark_table_as_done(specs, tableSpecs))
 		{
 			/* errors have already been logged */
 			return false;
 		}
-	}
-
-	if (!copydb_mark_table_as_done(specs, tableSpecs))
-	{
-		/* errors have already been logged */
-		return false;
 	}
 
 	if (specs->section == DATA_SECTION_TABLE_DATA)


dimitri:
Is there a way to handle C-c so that the SQLite summary table is still updated when CREATE INDEX was successful? Other than that, I think we could either check that indexes have already been built for tables that we skip (and queue building of missing indexes), or use CREATE INDEX IF NOT EXISTS and force “cache invalidation” by always running those commands.

arajkumar:
Thanks a lot for review @dimitri .

> I think we could either check that indexes have already been built for tables that we skip (and queue building of missing indexes)

I think this is what I had done in the new iteration.

src/bin/pgcopydb/catalog.c
```diff
@@ -6998,6 +6998,58 @@ catalog_upsert_process_info(DatabaseCatalog *catalog, ProcessInfo *ps)
 }
 
 
+/*
+ * catalog_delete_s_table deletes an s_table entry for the given oid.
+ */
+bool
+catalog_truncate_process(DatabaseCatalog *catalog)
```
arajkumar, line 7005: I think this can go in a separate PR. This is to fix the original issue discussed here in https://github.com/dimitri/pgcopydb/issues/692#issue-2167452958

src/bin/pgcopydb/catalog.c
```diff
@@ -7170,6 +7222,113 @@ catalog_iter_s_table_in_copy_init(SourceTableIterator *iter)
...
+		"         i.tableoid, t.qname, t.nspname, t.relname, "
+		"         isprimary, isunique, columns, i.sql, "
+		"         c.oid as constraintoid, conname, "
+		"         condeferrable, condeferred, c.sql as condef"
+		"    from s_index i "
+		"         join s_table t on t.oid = i.tableoid "
+		"         left join s_constraint c on c.indexoid = i.oid"
+		"   where i.oid not in ("
+		"	      select indexoid from summary where"
+		"				done_time_epoch is not null	and indexoid is not null)";
```
dimitri, line 7315: It's best to use WHERE NOT EXISTS rather than NOT IN, because of SQL semantics with NULL in the list and its impact of performance characteristics of query execution. Please rewrite the WHERE clause?

src/bin/pgcopydb/indexes.c
```diff
@@ -75,6 +75,35 @@ copydb_start_index_supervisor(CopyDataSpec *specs)
 }
 
 
+/*
+ * copydb_copy_not_processed_indexes creates all the indexes that have not been
+ * processed yet.
+ */
+bool
+copydb_copy_not_processed_indexes(CopyDataSpec *specs)
```
dimitri, line 83: Name is hard to make sense of, I think. Would prefer something like “copydb_copy_remaining_indexes” maybe, or even “copydb_create_remaining_indexes”.

src/bin/pgcopydb/catalog.c
```diff
@@ -7170,6 +7222,113 @@ catalog_iter_s_table_in_copy_init(SourceTableIterator *iter)
...
 
+/*
+ * catalog_iter_s_index_not_processed iterates over the list of indexes which
+ * are not processed in our catalogs. This might be due to user interruption.
+ */
+bool
+catalog_iter_s_index_not_processed(DatabaseCatalog *catalog,
+								   void *context,
+								   SourceIndexIterFun *callback)
+{
```
dimitri, line 7233: Naming: I'd rather have “remaining_indexes” compared to “index_not_processed”. Maybe “indexes_to_build”?

dimitri (changes requested):

src/bin/pgcopydb/catalog.c
dimitri: Yeah, it makes sense to isolate that parts in its own PR, please do!

src/bin/pgcopydb/indexes.c
arajkumar: `copydb_create_remaining_indexes` is good., sorry, I'm bad at naming :) Thanks for your suggestion.

dimitri: My main trick is to avoid negation or any kind of logic in the name. Make it obvious. An old source for inspiration for me was http://www.gar1t.com/blog/solving-embarrassingly-obvious-problems-in-erlang.html ; maybe that can be helpful to others too.

src/bin/pgcopydb/table-data.c
```diff
@@ -365,6 +365,15 @@ copydb_copy_supervisor(CopyDataSpec *specs)
...
 
+	/*
+	 * If we're resuming, try creating indexes that are not processed.
+	 */
+	if (specs->resume && !copydb_create_remaining_indexes(specs))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
```
dimitri, line 376: How does that work with tables that have not been taken care of yet?

arajkumar: @dimitri copydb_create_remaining_indexes would be called after completion of table copy workers.,when the control reaches here, all the tables should have been copied already.

dimitri: I see. I expected that processing to happen on table-by-table basis, in order to benefit from the usual amount of concurrency between copy workers and index workers. Does it make sense to check for remaining indexes for each table even when skipping the table on the basis that it's been already done? We might still need that piece of code here to catch indexes remaining to be created in case that all the table were done in the previous round I suppose?

arajkumar: @dimitri You are right. We need this piece of code to handle if all the tables were done already. I'm also wondering whether we can resume the respective workers with aborted commands. Basically, populating the queue with the entries from process & summary.

arajkumar: @dimitri What if we let the control proceed through all the tables and then through all the indexes in the original execution order, regardless of their completion status, skipping creation if it has already been completed? 

The best analogy for this is how make processes its targets. It reviews all the targets and generates the output if it doesn't exist, regardless of the current existence of the target output.

arajkumar: @dimitri Pushed a new commit which allows table-data worker to continue signalling index workers even if the table data copy is done. Which would essentially retain the original flow with same concurrency pattern.

dimitri (commented):
I like the shape of the PR better now, but what happens when table are marked all done but indexes are not marked all done? I suppose we need another pass at just the indexes in that case, right?

arajkumar:
@dimitri Even if all the tables are marked as done until indexes are marked as done, [table worker will be still triggered](https://github.com/dimitri/pgcopydb/blob/01872e58104c6f308c039d4b71df30799b16bc6f/src/bin/pgcopydb/table-data.c#L45-L53) and it would skip the data copy but signals the index workers to continue with the index creation.

So, I don't think we need another pass just to process the indexes.

dimitri (approved):

dimitri:
Indeed, you're right. Thanks!

PR #703, opened 2024-03-13
Problem: Idle source doesn't move replay progress forward

arajkumar:
When the source is idle, we get only KEEPALIVE messages and pgcopydb handles it with an empty transaction like below and updates origin progress.

```
BEGIN;
SELECT pgsql_replication_origin_xact_setup(lsn, timestamp);
COMMIT;
```

However, replication origin is handled differently by the postgres backend[1] to avoid database bloat and runtime overhead[2]. This optimization leads to persist origin progress only when the txn modifies the state of the database. So, the above statement effectively ignored by the backend leading to not updating the origin progress.

Solution: Write a keepalive logical message to move replicaiton origin progress using `pg_logical_emit_message`. Since writing logical message leads to writing the content to wal, it also updates the origin progress.

[1] https://github.com/postgres/postgres/blob/dbfc44716596073b99e093a04e29e774a518f520/src/backend/replication/logical/origin.c#L1437-L1438
[2] https://www.postgresql.org/docs/current/replication-origins.html

Next:
- We can rely on pg_replication_origin_progress to populate replay_lsn without having to map between [source and target curren wal lsn
](https://github.com/dimitri/pgcopydb/blob/227db6fdb119bc43fa1d70d34031ec6239f0d3f9/src/bin/pgcopydb/ld_apply.c#L1553-L1648)

This effectively fixes the problem described in https://github.com/dimitri/pgcopydb/pull/519.


dimitri:
Nice, thanks for having a look at that problem! I'm confused about using logical replication messages on the TARGET system though... Why would we send logical replication messages there? is there the only/easiest way to bypass the Postgres optimization for the replication origin thing?

arajkumar:
> is there the only/easiest way to bypass the Postgres optimization for the replication origin thing?

@dimitri Another workaround would be creating a temp table and dropping it on the same txn. Which also persists the origin LSN similar to dummy call to pg_logical_emit_message.

```
BEGIN;
CREATE TEMP TABLE IF NOT EXISTS pgcopydb_keepalive();
DROP TABLE IF EXISTS hello;
SELECT pg_replication_origin_xact_setup(lsn, timestamp);
COMMIT;
```

dimitri:
> > is there the only/easiest way to bypass the Postgres optimization for the replication origin thing?
> 
> @dimitri Another workaround would be creating a temp table and dropping it on the same txn. Which also persists the origin LSN similar to dummy call to pg_logical_emit_message.

This is an awful idea, as CREATE TEMP TABLE in Postgres is a source of catalog bloat, the worse there is. 

Is there another way to have at it? It might be that we can live with not registering our progress on the origin, at replay we're going to replay/skip KEEPALIVE messages and then reach to the ENDPOS anyway, right? For this to work, we would need to move replay_lsn forward in the sentinel even though we skipped updating the target database origin thing. What do you think about such a design?

arajkumar:
> > > is there the only/easiest way to bypass the Postgres optimization for the replication origin thing?
> > 
> > 
> > @dimitri Another workaround would be creating a temp table and dropping it on the same txn. Which also persists the origin LSN similar to dummy call to pg_logical_emit_message.
> 
> This is an awful idea, as CREATE TEMP TABLE in Postgres is a source of catalog bloat, the worse there is.
> 
> Is there another way to have at it? It might be that we can live with not registering our progress on the origin, at replay we're going to replay/skip KEEPALIVE messages and then reach to the ENDPOS anyway, right? For this to work, we would need to move replay_lsn forward in the sentinel even though we skipped updating the target database origin thing. What do you think about such a design?

The major motivation for this PR is to avoid LSN tracking to find replay_lsn. If we can reliably move forward the origin lsn, we can get rid of LSN tracking and simply use pg_replication_origin_progress as replay_lsn. The found the current LSN tracking is slow with persistence to sqlite, for e.g. the function [lsn_tracking_write](https://github.com/dimitri/pgcopydb/blob/227db6fdb119bc43fa1d70d34031ec6239f0d3f9/src/bin/pgcopydb/ld_apply.c#L380) takes ~3mins to complete when we have 100K transaction in a single WAL segment. We can entirely eliminate if we can rely on pg_replication_origin_progress.

arajkumar:
@dimitri The following workaround seem to work,

Instead of dropping the temp table, this one would truncate it. I hope this won't bloat catalog.

```
BEGIN;
CREATE TEMP TABLE IF NOT EXISTS pgcopydb_keepalive();
TRUNCATE pgcopydb_keepalive;
SELECT pg_replication_origin_xact_setup(lsn, timestamp);
COMMIT;
```

arajkumar:
@dimitri If we can reliably move origin progress LSN, that can become the source of truth for replay_lsn right? We can completely eliminate LSN mapping?

arajkumar:
@dimitri Do you think this is worth pursuing?

dimitri:
> @dimitri Do you think this is worth pursuing?

I like the idea of only using the replication origin to track progress and replay lsn, because it should be our source of truth. Having a cached version in our SQLite catalogs might not even be necessary at all, but I'm not sure yet. That said, if we can't register empty transactions and keepalive to mark progress, then we won't be able to.

I like the spirit of the changes here, but I don't like the kind of activity/work we are doing on the target server just to be able to register an empty transaction.

dimitri:
I am told that just calling `SELECT txid_current()` would disable the optimization on the target. Can you please try that @arajkumar ?

arajkumar:
> I am told that just calling `SELECT txid_current()` would disable the optimization on the target. Can you please try that @arajkumar ?

@dimitri , it works. Thank you for finding the workaround. Updated the fix to include your suggestion.

dimitri (approved):

PR #704, opened 2024-03-14
Problem: CDC replication/apply is slow

arajkumar:
Our existing implementation is basic and it will execute a statement and waits for its result to arrive before issuing next statement. We already improved DML statements targeting same table with in a transaction by coalescing them into [multi-value insert statement](https://github.com/dimitri/pgcopydb/pull/465), but it will be beneficial only for a transaction which has multiple DML statements.

However, we have encountered few cases where the transaction had only one DML statement. This kind of txn will be slower because statements are executed sequentially from the client's perspective and network round trip latency would be added for each statement. For e.g. consider a network round trip of 10ms, executing 100 statements would simply add 1000ms(1s) to the overall response time.

Here is an example of how a single statement transaction will be executed now,

1. Execute ["BEGIN"]()
2. Execute [bunch of "SET" statements]() which are related to replication session setup
3. Prepare [DML statement]()
4. Execute [prepared statement with values]()
5. Execute procedure to update [replication origin progress]()
6. Execute [COMMIT]()
7. Map [target current insert lsn to commit lsn]() for feedback reporting (i.e. sentinel replay_lsn)

**Solution**: Use [pipeline mode](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html) from libpq. It would facilitate sending commands without waiting for it's result.

**The goal of this implementation is to reduce network latency as much as possible while apply logical messages.**

The proposed implementation would enter into pipeline mode as soon as a new libpq connection is created for the ld_apply/ld_replay process. All the statements would be executed on a pipeline by default except few for statements which needs response immediately(e.g. step 7). A dedicated connection would be used to serve (step 7), because it needs synchronous response.

The following functions are being called on the target PGSQL handle from ld_apply & ld_replay.

* pgsql_begin
* pgsql_set_gucs
* pgsql_execute
* pgsql_replication_origin_xact_setup
* pgsql_prepare
* pgsql_execute_prepared
* pgsql_current_wal_insert_lsn
* pgsql_current_wal_flush_lsn

Among all of the above function, only pgsql_current_wal_insert_lsn and pgsql_current_wal_flush_lsn returns values and other functions are write only.

The idea is to have 2 PGSQL connection handles, 1 for all write activity which can go through pipeline and another one could be used for reading.

Pipeline connection has to be synced/drained at some point to avoid accumulating results on the server & client which would end up eating lots of heap memory. The current implementation syncs based on the time interval(i.e. for every 1s). There are other methods like statement/txn count based sync, which may or may not be efficient.

The following command can be used to generate loads to understand the performance improvement made by this commit,

```
CREATE TABLE metrics (
    time TIMESTAMP NOT NULL,
    name TEXT,
    id NUMERIC,
    value FLOAT
);
```

```
-- insert_metric.sql
\set id random(1, 1000000)
\set value random(0,100)
INSERT INTO metrics (time, name, value) VALUES (NOW(), 'metric_' || :id, :value);
```
```
pgbench -n -c 40 -j 1 -t 10000 -f insert_metric.sql $SOURCE
```

- `-c` number of database connection to utlize (i.e. server side concurrency)
- `-j` number of threads to create on the client machine (i.e. client side concurrency)

synchronous_commit=off)
```
pgbench -n -c 1 -j 1 -t 10000 -f insert_metric.sql $TARGET
```
```
tps = 1177 (without initial connection time)
```

```
tps = 175
```

```
tps = 1652
```

**This commit improves the single statement txn throughput by 10x**

Ideally, we should aim to get performance number close to direct ingestion(i.e. 1200 txn/s). We are 40% performing better than the baseline in this iteration. However, we can aim more as in the real system there will be more than 1 connection will be utilized. We can't really race against multiple connection doing steady ingestion around 1000 txn/s per connection, but lets optimize the single connection throughput to the max!

This change will be a foundation to the future improvements which steers towards that.

1. Optimize step 2 - Instead of executing bunch of SET statements for every txn, run once in the beginning for the session
2. Optimize step 9 - Probably we can simply use pg_replication_origin_progress as replay_lsn?

TODO
- [x] Conditionally enable pipeline on libpq version >= 14 - Should we still need to do this?

### DIFF
--- a/src/bin/pgcopydb/ld_stream.h
+++ b/src/bin/pgcopydb/ld_stream.h
@@ -365,6 +365,12 @@ typedef struct StreamApplyContext
 	/* target connection */
 	PGSQL pgsql;
 
+	/* target connection created in pipeline mode */
+	PGSQL pgsqlPipeline;
+
+	/* track the last time we synced the pipeline */
+	uint64_t pipelineSyncTime;
+
 	/* apply needs access to the catalogs to register sentinel replay_lsn */
 	DatabaseCatalog *sourceDB;
 	uint64_t sentinelSyncTime;
@@ -648,6 +654,8 @@ bool parseWal2jsonMessage(StreamContext *privateContext,
 bool stream_apply_catchup(StreamSpecs *specs);
 
 bool stream_apply_setup(StreamSpecs *specs, StreamApplyContext *context);
+
+bool stream_apply_cleanup(StreamApplyContext *context);
 
 bool stream_apply_wait_for_sentinel(StreamSpecs *specs,
 									StreamApplyContext *context);
--- a/src/bin/pgcopydb/ld_stream.h
+++ b/src/bin/pgcopydb/ld_stream.h
@@ -363,10 +363,10 @@ typedef struct StreamApplyContext
 	CDCPaths paths;
 
 	/* target connection to find current wal_lsn for replay_lsn mapping */
-	PGSQL controlPGSQL;
+	PGSQL controlPgConn;
 
 	/* target connection created in pipeline mode responsible for apply */
-	PGSQL applyPGSQL;
+	PGSQL applyPgConn;
 
 	/* apply needs access to the catalogs to register sentinel replay_lsn */
 	DatabaseCatalog *sourceDB;
--- a/src/bin/pgcopydb/ld_stream.h
+++ b/src/bin/pgcopydb/ld_stream.h
@@ -362,14 +362,11 @@ typedef struct StreamApplyContext
 {
 	CDCPaths paths;
 
-	/* target connection */
-	PGSQL pgsql;
+	/* target connection to find current wal_lsn for replay_lsn mapping */
+	PGSQL controlPGSQL;
 
-	/* target connection created in pipeline mode */
-	PGSQL pgsqlPipeline;
-
-	/* track the last time we synced the pipeline */
-	uint64_t pipelineSyncTime;
+	/* target connection created in pipeline mode responsible for apply */
+	PGSQL applyPGSQL;
 
 	/* apply needs access to the catalogs to register sentinel replay_lsn */
 	DatabaseCatalog *sourceDB;
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -1909,11 +1909,17 @@ pgsql_pipeline_enter(PGSQL *pgsql)
 	log_trace("Enabled pipeline mode");
 	return true;
 #else
-	log_warn("Pipeline mode is not available on Postgres version %d, "
-			 "need at least version 14",
-			 PG_MAJORVERSION_NUM);
+	static bool warned = false;
 
-	return false;
+	if (!warned)
+	{
+		log_warn("Pipeline mode is not available on Postgres version %d, "
+				 "need at least version 14",
+				 PG_MAJORVERSION_NUM);
+		warned = true;
+	}
+
+	return true;
 #endif
 }
 
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -1925,7 +1925,7 @@ pgsql_enable_pipeline_mode(PGSQL *pgsql)
 
 /*
  * pgsql_sync_pipeline drains the pipeline by sending a SYNC message and
- * reading until we get a PGRES_PIPELINE_SYNC result.
+ * reads results until we get a PGRES_PIPELINE_SYNC result.
  */
 bool
 pgsql_sync_pipeline(PGSQL *pgsql)
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -30,12 +30,11 @@
 #include "signals.h"
 #include "string_utils.h"
 
-/* Pipeline mode is only available in PostgreSQL 14 and later */
-#if PG_MAJORVERSION_NUM >= 14
-#define onPipelineMode(conn) (PQpipelineStatus(conn) == PQ_PIPELINE_ON)
+#if defined(LIBPQ_HAS_PIPELINING) && LIBPQ_HAS_PIPELINING
+#define pqPipelineModeEnabled(conn) (PQpipelineStatus(conn) == PQ_PIPELINE_ON)
 #else
-#warning Pipeline mode is only available in PostgreSQL 14 and later
-#define onPipelineMode(conn) (false)
+#warning Pipeline mode is only available in libpq PostgreSQL 14 and later
+#define pqPipelineModeEnabled(conn) (false)
 #endif
 
 static char * ConnectionTypeToString(ConnectionType connectionType);
@@ -1504,7 +1503,7 @@ pgsql_execute_with_params(PGSQL *pgsql, const char *sql, int paramCount,
 		return false;
 	}
 
-	bool pipelineMode = onPipelineMode(connection);
+	bool pipelineMode = pqPipelineModeEnabled(connection);
 
 	/* parseFun is not allowed in pipeline mode */
 	if (pipelineMode && parseFun != NULL)
@@ -1871,13 +1870,13 @@ pgsql_fetch_results(PGSQL *pgsql, bool *done,
 
 
 /*
- * pgsql_pipeline_enter enables the pipeline mode in the given PGSQL
+ * pgsql_enable_pipeline_mode enables the pipeline mode in the given PGSQL
  * connection. It also sets the connection to non-blocking mode.
  */
 bool
-pgsql_pipeline_enter(PGSQL *pgsql)
+pgsql_enable_pipeline_mode(PGSQL *pgsql)
 {
-#if PG_MAJORVERSION_NUM >= 14
+#if defined(LIBPQ_HAS_PIPELINING) && LIBPQ_HAS_PIPELINING
 	PGconn *conn = pgsql_open_connection(pgsql);
 
 	if (conn == NULL)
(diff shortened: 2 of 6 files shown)


arajkumar:
I will keep this as draft until I fix all CI failures.

dimitri:
Thanks @arajkumar for working on this, that's much appreciated!

My understanding is that we need libpq 14 at build time to enable this yes. Unfortunately not all build systems out there understand that we can build with libpq 16 and then work with different Postgres versions (RPM is one of these).

Another aspect: is it possible to have a “sync” step in the pipeline mode, for the COMMIT and LSN tracking? if that was possible, we could skip using a second transaction entirely, and I think we have to do that for correctness.

arajkumar:
> My understanding is that we need libpq 14+ at build time to enable this yes. Unfortunately not all build systems out there understand that we can build with libpq 16 and then work with different Postgres versions (RPM is one of these).

@dimitri I've added this in the recent commit.

> Another aspect: is it possible to have a “sync” step in the pipeline mode, for the COMMIT and LSN tracking? if that was possible, we could skip using a second transaction entirely, and I think we have to do that for correctness.

This would make pipeline to be synced for each commit and would reduce the overall throughput of the pipeline implementation. It would essentially make pipeline mode to used only with in a txn. We are having exactly same implementation on our internal fork which is branched out around 0.13. The throughput is not so great when you have txn which has single DML statement.

Since we already do `synchronous_commit=off`, I'm not sure how much difference we would add by having a sync step for each COMMIT. I also thought of eliminating the entire LSN tracking and replace it with pg_replication_origin_progress. Do you think this is not a good idea?

arajkumar:
@dimitri 

Here more details about overhead caused by the current LSN tracking implementation. This is kinda micro benchmark where it uses only `pgcopydb apply` to ingest 122987 records with same amount tranaction.

```
$ grep BEGIN ~/dimitri/bench.sql | wc -l
122987
$ grep INSERT ~/dimitri/bench.sql | wc -l
122987

$ time pgcopydb stream apply --resume --not-consistent ~/dimitri/bench.sql
06:05:02.850 19838 INFO   Running pgcopydb version 0.15.38.gc1bc3d9 from "/home/ubuntu/dimitri/pgcopydb/src/bin/pgcopydb/pgcopydb"
06:05:02.887 19838 INFO   Using work dir "/tmp/pgcopydb"
06:05:02.936 19838 INFO   Setting up previous LSN from replication origin "pgcopydb" progress at 0/0

06:05:02.982 19838 INFO   Replaying changes from file "/home/ubuntu/dimitri/bench.sql"

06:06:29.934 19838 INFO   Before syncing with the pgcopydb sentinel
06:09:37.642 19838 INFO   After syncing with the pgcopydb sentinel

pgcopydb stream apply --resume --not-consistent ~/dimitri/bench.sql  5.11s user 6.72s system 4% cpu 4:34.82 total
```

With pipeline mode, apply took 4m 34 seconds to ingest 122987 records. i.e 448 records/sec.

If we dissect those steps further,  DML insert took (06:06:29.934 - 06:05:02.982) => 1m 27 seconds,  i.e. 1,413 recods/sec.

Pgcopydb sentinel took (06:09:37.642-06:06:29.934) => 3m 8seconds.

You can find the CPU profiling taken during the above benchmark => https://pprof.me/46714b5a2a244609af5e72e035427a5c/

The benchmark shows considerable amount of time (~80%) of the time spend in stream_apply_sync_sentinel and stream_apply_track_insert_lsn.

image


Is there an opportunity to optimize stream_apply_sync_sentinel & stream_apply_track_insert_lsn?, probably yes, I didn't really spent time on that, but should we do that? Why not make replication_origin work and simply use replication_origin_progress as replay_lsn.?

src/bin/pgcopydb/pgsql.c
```diff
@@ -30,6 +30,14 @@
...
 #include "string_utils.h"
 
+/* Pipeline mode is only available in PostgreSQL 14 and later */
+#if PG_MAJORVERSION_NUM >= 14
+#define onPipelineMode(conn) (PQpipelineStatus(conn) == PQ_PIPELINE_ON)
+#else
+#warning Pipeline mode is only available in PostgreSQL 14 and later
+#define onPipelineMode(conn) (false)
+#endif
+
```
dimitri, line 40: I don't like the name `onPipelineMode` too much. What about `pqPipelineModeEnabled` or something like that?

src/bin/pgcopydb/ld_stream.h
```diff
@@ -365,6 +365,12 @@ typedef struct StreamApplyContext
 	/* target connection */
 	PGSQL pgsql;
 
+	/* target connection created in pipeline mode */
+	PGSQL pgsqlPipeline;
+
+	/* track the last time we synced the pipeline */
+	uint64_t pipelineSyncTime;
+
```
dimitri, line 373: Again, not fond of the naming. This should be the `PGSQL applyConn`, as opposed to I suppose the `PGSQL controlConn` or something. Also pipelineSyncTime looks like it should belong within the PGSQL structure itself, right?

dimitri (changes requested):
First round of review is in, lots of naming issue, the build-time vs run-time thing with libpq14 for pipeline mode, and not so much after that. I understand why we need to stay in pipeline mode in the connection and can't have sync queries in there, but I failed to read a comment explaining why we need two connections now, what they are used for, etc. It's also a problem with the PGSQL client connection names.

src/bin/pgcopydb/pgsql.c
```diff
@@ -1843,6 +1870,213 @@ pgsql_fetch_results(PGSQL *pgsql, bool *done,
...
+	if (!warned)
+	{
+		log_warn("Pipeline mode is not available on Postgres version %d, "
+				 "need at least version 14",
+				 PG_MAJORVERSION_NUM);
+		warned = true;
+	}
+
+	return true;
+#endif
```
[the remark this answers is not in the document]
dimitri: What I mean is that if we take the opposite test, we can put the error message in the if branch and then have the normal code outside of any pre-processor directives.

``` diff
- #if defined(LIBPQ_HAS_PIPELINING) && LIBPQ_HAS_PIPELINING
+ #if !(defined(LIBPQ_HAS_PIPELINING) && LIBPQ_HAS_PIPELINING)
```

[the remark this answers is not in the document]
arajkumar: > have the normal code outside of any pre-processor directives

@dimitri the normal code calls pipeline specific functions which need to be under #ifdef again.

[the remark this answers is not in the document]
dimitri: Sorry it seems I'm very confused. It's just that I don't like having lots of code hidden within preprocessor directives, but I also asked that we could still build pgcopydb with older Postgres versions of libpq. Thinking more about it, I can't see how to make it build without doing what you did. 

I was thinking of something like the following, but of course then we just can't build with older versions.

``` c
#if ...
#error "libpq 14 or newer needed"
#endif
```

In summary, sorry about the confusion, let's keep what you did and move on to the rest of the review.

src/bin/pgcopydb/ld_stream.h
```diff
@@ -362,8 +362,11 @@ typedef struct StreamApplyContext
 {
 	CDCPaths paths;
 
-	/* target connection */
-	PGSQL pgsql;
+	/* target connection to find current wal_lsn for replay_lsn mapping */
+	PGSQL controlPGSQL;
+
+	/* target connection created in pipeline mode responsible for apply */
+	PGSQL applyPGSQL;
```
dimitri, line 369: Can we name them `controlPgConn` and `applyPgConn` maybe?

src/bin/pgcopydb/pgsql.c
```diff
@@ -1843,6 +1869,216 @@ pgsql_fetch_results(PGSQL *pgsql, bool *done,
...
+	return true;
+#else
+	static bool warned = false;
+
+	if (!warned)
+	{
+		log_warn("libpq in Postgres version %d does not support pipeline mode, "
+				 "need at least version 14",
+				 PG_MAJORVERSION_NUM);
+		warned = true;
```
dimitri, line 2074: I still think we should WARN at build-time using the `#warning` directive, and rewrite the error-message in a way that makes sense for our run-time users:

``` c
#warning "Compiling pgcopydb without pipeline mode support"
    log_warn("Skipping libpq pipeline mode optimisation because pgcopydb was built with libpq %d, "
             "pipeline mode is available since libpq 14");
```

Please also see if we can add libpq version in the `#warning` message?

dimitri (commented):
Looks like we're almost there, thanks again!

src/bin/pgcopydb/pgsql.c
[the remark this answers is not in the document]
arajkumar: @dimitri , no worries.!

src/bin/pgcopydb/pgsql.c
arajkumar: @dimitri I've already warning directive here => https://github.com/dimitri/pgcopydb/pull/704/files#diff-84ab9d926dfb0eda4b691a03c25ba22211e96c9f749fa57ff44ca24758bb5777R36. 

I will change the message as per your suggestion. thanks!

dimitri (approved):

dimitri:
> Next, I'm planning to change Dockerfile to install newer versions of postgres-client(14?) libraries to enable the pipeline mode by default.

Sounds good. I have also been wondering if switching the copy protocol calls that we do to asynchronous would raise any performance improvements... would that be an area you'd be willing to investigate?

arajkumar:
> Sounds good. I have also been wondering if switching the copy protocol calls that we do to asynchronous would raise any performance improvements... would that be an area you'd be willing to investigate?

@dimitri Do you mean copy protocol used for initial data copying?

dimitri:
I was thinking about `PQgetCopyData` but I now see that we already did that work, I just failed to remember about it because it was done to solve error handling rather than performance characteristics.

PR #708, opened 2024-03-18
Fix SQL syntax error in group_concat.

shubhamdhama:
Without this fix the migrations would fail with this error when hit this code path,
```
2024-03-13 09:56:42.303 14231 ERROR  catalog.c:7286            Failed to prepare SQLite statement: select group_concat(attname, ', ' order by attnum)        filter (where not attisgenerated)   from s_attr  where oid = $1
2024-03-13 09:56:42.303 14231 ERROR  catalog.c:7287            [SQLite] near "order": syntax error
```


dimitri (approved):

dimitri:
Thanks @shubhamdhama ! I wonder how that got past usual care and CI/tests...

Issue #694, opened 2024-03-05
Can't clone extension installed in non default schema


plvnv:
pgcopy version 
`Running pgcopydb version 0.15.30.gcb55ea1`

source db  `PostgresPro (standard) 11.22.1`
target db  `PostgresPro (standard) 13.13.1`

Can't clone extensions and objects using this extensions installed in non default schema

On source database extensions install in separate schema "extensions" 

```
\dx
                                                       List of installed extensions
        Name        | Version |   Schema   |                                          Description                                          
--------------------+---------+------------+-----------------------------------------------------------------------------------------------
 hostname           | 1.0.0   | extensions | Get the server host name
 oracle_fdw         | 1.2     | extensions | foreign data wrapper for Oracle access
 orafce             | 4.6     | extensions | Functions and operators that emulate a subset of functions and packages from the Oracle RDBMS
 pg_buffercache     | 1.3     | extensions | examine the shared buffer cache
 pg_repack          | 1.4.8   | extensions | Reorganize tables in PostgreSQL databases with minimal locks
 pg_sortstats       | 0.0.1   | extensions | An extension collecting statistics about sorts
 pg_stat_kcache     | 2.2.0   | extensions | Kernel statistics gathering
 pg_stat_statements | 1.6     | extensions | track execution statistics of all SQL statements executed
 pg_trgm            | 1.4     | extensions | text similarity measurement and index searching based on trigrams
 pgfincore          | 1.2     | public     | examine and manage the os buffer cache
 pgstattuple        | 1.5     | extensions | show tuple-level statistics
 plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language
 postgres_fdw       | 1.0     | extensions | foreign-data wrapper for remote PostgreSQL servers
 tablefunc          | 1.0     | extensions | functions that manipulate whole tables, including crosstab
 uuid-ossp          | 1.1     | extensions | generate universally unique identifiers (UUIDs)
```

clone command 
`pgcopydb clone  --follow --restart --table-jobs 4 --index-jobs 4 --plugin wal2json --wal2json-numeric-as-string`

Error log

```
2024-03-04 16:35:28.008 38106 INFO   main.c:138                Running pgcopydb version 0.15.30.gcb55ea1 from "/opt/pgpro/std-13/bin/pgcopydb"
[...]

PR #711, opened 2024-03-20
Create extensions in the schema of the original extension

kbarber:
Related to this one: https://github.com/dimitri/pgcopydb/issues/694 and this: https://github.com/dimitri/pgcopydb/pull/717

The `pgcopydb copy extensions` command doesn't create the extension in the original scheme from the source. This patch attempts to fix that by using the `with schema XXX` clause during `create extension`.

### DIFF
--- a/src/bin/pgcopydb/extensions.c
+++ b/src/bin/pgcopydb/extensions.c
@@ -233,8 +233,8 @@ copydb_copy_extensions_hook(void *ctx, SourceExtension *ext)
 		HASH_FIND(hh, context->reqs, extname, strlen(extname), req);
 
 		appendPQExpBuffer(sql,
-						  "create extension if not exists \"%s\" cascade",
-						  ext->extname);
+						  "create extension if not exists \"%s\" with schema \"%s\" cascade",
+						  ext->extname, ext->extnamespace);
 
 		if (req != NULL)
 		{
@@ -248,6 +248,7 @@ copydb_copy_extensions_hook(void *ctx, SourceExtension *ext)
 			log_error("Failed to build CREATE EXTENSION sql buffer: "
 					  "Out of Memory");
 			(void) destroyPQExpBuffer(sql);
+			return false;
 		}
 
 		log_info("Creating extension \"%s\"", ext->extname);
@@ -255,6 +256,8 @@ copydb_copy_extensions_hook(void *ctx, SourceExtension *ext)
 		if (!pgsql_execute(dst, sql->data))
 		{
 			log_error("Failed to create extension \"%s\"", ext->extname);
+			(void) destroyPQExpBuffer(sql);
+			return false;
 		}
 
 		(void) destroyPQExpBuffer(sql);
--- a/tests/extensions/copydb.sh
+++ b/tests/extensions/copydb.sh
@@ -31,6 +31,13 @@ EOF
 psql -a -1 ${PGCOPYDB_SOURCE_PGURI_SU} <<EOF
 create extension intarray cascade;
 create extension postgis cascade;
+create schema foo;
+create extension hstore with schema foo cascade;
+EOF
+
+# create schemas for extensions on the target pagila database (needs superuser)
+psql -a -1 ${PGCOPYDB_TARGET_PGURI_SU} <<EOF
+create schema foo;
 EOF
 
 #


src/bin/pgcopydb/pgcmd.c
```diff
@@ -443,6 +443,11 @@ pg_dump_db(PostgresPaths *pgPaths,
 		args[argsIndex++] = nspname;
 	}
 
+	// TODO: this was a naive way to try and filter the extensions from pg_dump, but
+	// it isn't effective, we need to find a better way to do this.
+	args[argsIndex++] = "--extension";
+	args[argsIndex++] = "pg_stat_statements";
+
```
dimitri, line 450: The short answer is that `pg_dump --extension` option is broken. It's yet to be determined if the documentation or the implement is broken, but it's not usable as it is.

dimitri (commented):
Thanks for having a look into this problem, your work is appreciated!

Is there a way to clean-up the PR and make progress? I believe the parts where we add the `extnamespace` should just work, the part that might need more attention is the creation of the needed schema? What happens if we remove all your changes in `pgcmd.c`?

kbarber:
> Thanks for having a look into this problem, your work is appreciated!
> 
> Is there a way to clean-up the PR and make progress? I believe the parts where we add the `extnamespace` should just work, the part that might need more attention is the creation of the needed schema? What happens if we remove all your changes in `pgcmd.c`?

Yeah sorry we were experimenting yesterday and I left it there in that state. We got a bit stuck so we left some comments as bread trails :-).

With just the changes the the "create extension" we get missing schema messages if they don't exist when we transfer the extension. We'll get you the error messages.

We wouldn't mind some help on the ordering flow here, what would be the solution if the above error occurs you think? We create the schemas elsewhere? What I get I'm not sure about is this really. I poked around the implicit exclusion code areas to try and include the schema where extensions exist but that resulted in other problems.

I also was wondering: What happens to schemas that have both data in them and extensions?

kbarber:
So using this DDL on the sourcedb:

```
create schema foo;
create extension hstore with schema foo cascade;
create table foo.bar (id int);
```

We get this failure for `copy extension` with this patch:

```
root@ff36ceb2b719:/# rm -rf /tmp/pgcopydb2 && \
> pgcopydb copy extensions \
>   --source postgres://**********:**********@sourcedb:5432/********** \
>   --target postgres://**********:**********@targetdb:5432/********** \
>   --dir /tmp/pgcopydb2 \
>   --skip-ext-comments \
>   --no-owner \
>   --no-acl \
>   --fail-fast
17:27:34.420 103 INFO   Running pgcopydb version 0.15 from "/usr/local/bin/pgcopydb"
17:27:34.443 103 INFO   [SOURCE] Copying database from "postgres://**********@sourcedb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
17:27:34.443 103 INFO   [TARGET] Copying database into "postgres://**********@targetdb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
17:27:34.564 103 INFO   Using work dir "/tmp/pgcopydb2"
17:27:35.864 103 INFO   Exported snapshot "00000003-03101F73-1" from the source database
17:27:38.490 103 INFO   Fetched information for 3 extensions
17:27:40.331 103 INFO   Found 0 indexes (supporting 0 constraints) in the target database
17:27:42.234 114 INFO   Creating extension "hstore"
17:27:43.246 114 ERROR  [TARGET 9216] ERROR:  schema "foo" does not exist
17:27:43.247 114 ERROR  [TARGET 9216] ERROR:  schema "foo" does not exist
17:27:43.247 114 ERROR  [TARGET 9216] SQL query: create extension if not exists "hstore" with schema "foo" cascade
17:27:43.247 114 ERROR  [TARGET 9216] SQL params:
17:27:43.247 114 ERROR  Failed to create extension "hstore"
17:27:43.254 114 INFO   Creating extension "pg_stat_statements"
17:27:44.318 114 WARN   NOTICE:  extension "pg_stat_statements" already exists, skipping
17:27:44.323 114 WARN   NOTICE:  INSTALLING EXTENSION AFTER CREATE
17:27:44.325 114 INFO   Creating extension "plpgsql"
17:27:45.504 114 WARN   NOTICE:  extension "plpgsql" already exists, skipping
17:27:45.511 114 WARN   NOTICE:  INSTALLING EXTENSION AFTER CREATE
```

As expected, can't find the schema. My question is, when should we create the schema in this kind of flow? I thought initially `pgcopydb copy schema` might do that, but its not specific to SCHEMAS lol.

Now to explain my confusion yesterday, there is another failure angle here which is when you use clone, its a different code path, but the results look very similar:

```
> pgcopydb clone \
>   --source postgres://**********:**********@sourcedb:5432/********** \
>   --target postgres://**********:**********@targetdb:5432/********** \
>   --dir /tmp/pgcopydb \
>   --skip-ext-comments \
>   --no-owner \
>   --no-acl \
>   --fail-fast
17:37:47.536 118 INFO   Running pgcopydb version 0.15 from "/usr/local/bin/pgcopydb"
17:37:47.557 118 INFO   [SOURCE] Copying database from "postgres://**********@sourcedb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
17:37:47.558 118 INFO   [TARGET] Copying database into "postgres://**********@targetdb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
17:37:47.676 118 INFO   Using work dir "/tmp/pgcopydb"
17:37:48.884 118 INFO   Exported snapshot "00000003-03103FD8-1" from the source database
17:37:49.620 129 INFO   STEP 1: fetch source database tables, indexes, and sequences
17:37:51.629 129 INFO   Fetched information for 1 tables (including 0 tables split in 0 partitions total), with an estimated total of 18 quintillion tuples and 0 B on-disk
17:37:51.758 129 INFO   Fetched information for 0 indexes (supporting 0 constraints)
17:37:51.880 129 INFO   Fetching information for 0 sequences
17:37:52.149 129 INFO   Fetched information for 3 extensions
17:37:53.974 129 INFO   Found 0 indexes (supporting 0 constraints) in the target database
17:37:53.990 129 INFO   STEP 2: dump the source database schema (pre/post data)
17:37:53.995 129 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-03103FD8-1 --section pre-data --exclude-schema foo --exclude-schema heroku_ext --file /tmp/pgcopydb/schema/pre.dump 'postgres://**********@sourcedb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
17:38:04.926 129 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-03103FD8-1 --section post-data --exclude-schema foo --exclude-schema heroku_ext --file /tmp/pgcopydb/schema/post.dump 'postgres://**********@sourcedb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
17:38:15.277 129 INFO   STEP 3: restore the pre-data section to the target database
17:38:16.284 129 INFO   ALTER DATABASE "**********" SET "search_path" TO '$user', 'public', 'heroku_ext';
17:38:16.685 129 INFO    /usr/bin/pg_restore --dbname 'postgres://**********@targetdb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --jobs 4 --no-owner --no-acl --use-list /tmp/pgcopydb/schema/pre-filtered.list /tmp/pgcopydb/schema/pre.dump
17:38:20.063 129 ERROR  pg_restore: error: could not execute query: ERROR:  schema "foo" does not exist
17:38:20.064 129 ERROR  Command was: CREATE EXTENSION IF NOT EXISTS hstore WITH SCHEMA foo;
17:38:25.261 129 WARN   pg_restore: warning:
17:38:25.261 129 ERROR  errors ignored on restore: 1
17:38:25.279 129 ERROR  Failed to run pg_restore: exit code 1
17:38:25.279 129 ERROR  Failed to prepare schema on the target database, see above for details
17:38:25.279 129 ERROR  Failed to clone source database, see above for details
17:38:25.445 118 ERROR  clone process 129 has terminated [6]
```

Which is why I was led to believe modifying the --exclude-schema and --extension stuff might help here playing around.

I think these are two related but different issues.

kbarber:
I've wrangled a test that shows the symptoms now. I had to include a return though in other places, to fail hard on extension installation: https://github.com/dimitri/pgcopydb/actions/runs/8381043186/job/22951687996?pr=711

kbarber:
I feel like we are lacking in our `copy` steps a precursor that is `schemas` perhaps, or we should do it in the extensions stage perhaps I dunno. We can always ship this fix, do the manual schema creation in our tests and lie basically, and allude to the non-feature some where, but we'd want to follow up soonish with a `copy schemas` command, and probably all the other things like `list schemas` perhaps.

Perhaps we need to lazily create schemas in the extensions transfer, but I worry about the role ownership of schemas requiring a bit of work to get right if people are using acls.

src/bin/pgcopydb/pgcmd.c
kbarber: Yep most interesting. I've learnt today about your usage of filtered lists and this kind of approach makes more sense to me now, its more systematic if somewhat more laborious, but nice that it works :-).

kbarber:
I've fixed the copy but have not solved the schema copy problem.

dimitri:
Nice, I like the shape we're getting at. At the moment I'm leaning towards adding the creation of the needed schema right within pgcopydb in the same PR as this one, rather than in a separate step. Does it makes sense for you?

What happens I believe is that we have infrastructure to filter-out both the schema and the extension objects when using `--skip-extensions`, but we are lacking the infrastructure to ensure creating the schema objects that extensions are known to depend upon when using a normal `pgcopydb clone` (or even a `pgcopydb copy extensions`) command.

Basically, it boils down to only calling `schema_list_ext_schemas` in the `copydb_fetch_filtered_oids` and filling-in the `s_namespace` in the filters database. We need the same table in our source database (see `sourceDBcreateDDLs`) and we need to fill-in that source catalog with the list of namespaces that are required for being able to create extensions. Then we should run a bunch of `CREATE SCHEMA` commands. We need to restore also comments, ownership and privileges on these schemas, so I think we might want to call in `pg_restore` for this step, with a new archive list file that only contains lines for the target schemas.

What do you think?

kbarber:
@dimitri I can look into this, what do you think about this workaround for the `pgcopydb clone` bug for the time being: https://github.com/dimitri/pgcopydb/pull/717

kbarber:
@dimitri I've added a small amount of draft code that tries to focus on copying the namspaces using pg_restore/pg_dump, took me a while to figure out the pre list filtered stuff, still learning, and I don't have the acls or comments working yet.

I guess with this kind of thing you can then do:

```
pgcopydb copy namespaces
pgcopydb copy extensions
pgcopydb clone --skip-extensions --skip-namespaces
```

If you combine it with something like: https://github.com/dimitri/pgcopydb/pull/717

I'm not sure this is exactly what you had in mind entirely, but it helped me understand how the flow works a little more today. I will keep poking.

kbarber:
@dimitri hi there, this one took me a bit to get right, but its ready for review again after getting it passing. Could you take a look when you have time?

dimitri:
Hi @kbarber, (back from vacations now), I think your idea and implementation for `pgcopydb copy schemas` is a good candidate for inclusion in pgcopydb as yet another interesting tool to have, but the PR conflates that with adding schema support in CREATE EXTENSION, and unfortunately we are not ready for that parts yet.

Can you split the `pgcopydb copy schemas` command appart in its own PR, and then we'll come back to this one and add coverage of listing the extension dependencies and making sure we install them (if needed, see your other PR) before running the CREATE EXTENSION code, in a way that allows for the SCHEMA parts to be specified at this point.

Given the steps above, we should have a fully automated fix of all the things related to ensuring extensions are created on the target system in the same schemas as on the source system.

kbarber:
@dimitri I've removed the schema copy command here, and just fixed the specific bug with `copy extensions`. The schema copy command PR is now over here: https://github.com/dimitri/pgcopydb/pull/751

dimitri (commented):
Thanks @kbarber !

As a first step that works fine, and when we have the `pgcopydb copy schemas` command it's possible to use that before copying the extensions. Now I wonder if we should automated listing the schemas that extensions depend on and dump/restore these when doing `pgcopydb copy extensions`?

kbarber:
@dimitri my argument for merging as it is: people right now might be copying extensions with this bug, not aware they are possibly breaking future queries. The bug falls into the silent-but-deadly category .... a transfer might work fine until a query that happens to use fully qualified items pointing at the extension schema is executed and thats not quite data loss, but it feels like a surprising issue we're creating that will surface later. Part of me would rather they fail and its obvious, and we learn.

Of course most users notice this during transfer, because most extension references are in DDL, its not like its always silent.

However on the other side, people might find it surprising things suddenly break after this fix if they don't create schemas upfront and it would be great to follow it up v quickly with a real feature/fix to complete the story.

If you're interested in auto loading schemas for extensions, that seems cool happy to follow your suggestion - do we do that elsewhere like for tables? I think the code we have in the other PR might help here somewhat with some changes. I'm happy to implement it like this.

PR #712, opened 2024-03-21
Fix "database is locked" due to concurrent writes between follow and clone.

shubhamdhama:
Also, for test_decoding plugin we would require the catalog information so it's better to let it initialise first and then we start the prefetch.


src/bin/pgcopydb/cli_clone_follow.c
```diff
@@ -249,6 +249,12 @@ clone_and_follow(CopyDataSpec *copySpecs)
 		exit(EXIT_CODE_INTERNAL_ERROR);
 	}
 
+	if (!copydb_fetch_schema_and_prepare_specs(copySpecs))
+	{
+		/* errors have already been logged */
+		exit(EXIT_CODE_INTERNAL_ERROR);
+	}
+
```
dimitri, line 257: I believe sub-processes also call that same function, a comment here explaining why it makes sense to call early in the main process would make sense. I suppose it's all about preparing the caches?

dimitri (changes requested):

dimitri (approved):

PR #716, opened 2024-03-21
Force sequences reset at the end of the follow process.

dimitri:


shubhamdhama (commented):
`copydb_copy_all_sequences` uses the `last_value` from the catalog. That value is stale, that is, it is picked using the snapshot. So we need to refresh those value from the source, then run the `copydb_copy_all_sequences`.

PR #717, opened 2024-03-21
Fix cloning with extensions in non default schema

kbarber:
Related to: https://github.com/dimitri/pgcopydb/pull/711 and https://github.com/dimitri/pgcopydb/issues/694

So using this DDL on the sourcedb:

```
create schema foo;
create extension hstore with schema foo cascade;
create table foo.bar (id int);
```

This patch attempts to fix the clone failure when there are extensions in different schemas due to the lack of the schema being created correctly beforehand.

For example:

```
> pgcopydb clone \
>   --source postgres://**********:**********@sourcedb:5432/********** \
>   --target postgres://**********:**********@targetdb:5432/********** \
>   --dir /tmp/pgcopydb \
>   --skip-ext-comments \
>   --no-owner \
>   --no-acl \
>   --fail-fast
17:37:47.536 118 INFO   Running pgcopydb version 0.15 from "/usr/local/bin/pgcopydb"
17:37:47.557 118 INFO   [SOURCE] Copying database from "postgres://**********@sourcedb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
17:37:47.558 118 INFO   [TARGET] Copying database into "postgres://**********@targetdb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
17:37:47.676 118 INFO   Using work dir "/tmp/pgcopydb"
17:37:48.884 118 INFO   Exported snapshot "00000003-03103FD8-1" from the source database
17:37:49.620 129 INFO   STEP 1: fetch source database tables, indexes, and sequences
17:37:51.629 129 INFO   Fetched information for 1 tables (including 0 tables split in 0 partitions total), with an estimated total of 18 quintillion tuples and 0 B on-disk
17:37:51.758 129 INFO   Fetched information for 0 indexes (supporting 0 constraints)
17:37:51.880 129 INFO   Fetching information for 0 sequences
17:37:52.149 129 INFO   Fetched information for 3 extensions
17:37:53.974 129 INFO   Found 0 indexes (supporting 0 constraints) in the target database
17:37:53.990 129 INFO   STEP 2: dump the source database schema (pre/post data)
17:37:53.995 129 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-03103FD8-1 --section pre-data --exclude-schema foo --exclude-schema heroku_ext --file /tmp/pgcopydb/schema/pre.dump 'postgres://**********@sourcedb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
17:38:04.926 129 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-03103FD8-1 --section post-data --exclude-schema foo --exclude-schema heroku_ext --file /tmp/pgcopydb/schema/post.dump 'postgres://**********@sourcedb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
17:38:15.277 129 INFO   STEP 3: restore the pre-data section to the target database
17:38:16.284 129 INFO   ALTER DATABASE "**********" SET "search_path" TO '$user', 'public', 'heroku_ext';
17:38:16.685 129 INFO    /usr/bin/pg_restore --dbname 'postgres://**********@targetdb:5432/**********?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --jobs 4 --no-owner --no-acl --use-list /tmp/pgcopydb/schema/pre-filtered.list /tmp/pgcopydb/schema/pre.dump
17:38:20.063 129 ERROR  pg_restore: error: could not execute query: ERROR:  schema "foo" does not exist
17:38:20.064 129 ERROR  Command was: CREATE EXTENSION IF NOT EXISTS hstore WITH SCHEMA foo;
17:38:25.261 129 WARN   pg_restore: warning:
17:38:25.261 129 ERROR  errors ignored on restore: 1
17:38:25.279 129 ERROR  Failed to run pg_restore: exit code 1
17:38:25.279 129 ERROR  Failed to prepare schema on the target database, see above for details
17:38:25.279 129 ERROR  Failed to clone source database, see above for details
17:38:25.445 118 ERROR  clone process 129 has terminated [6]
```


kbarber:
@dimitri I was trying to attack the problem from the `clone` perspective on this patch, it does work: but I'm doubtful. I commented out the part of the union that includes all the items in `s_namespace` but I'm not entirely certain this breaks behaviour elsewhere ... such as filtering on namespaces or something.

dimitri (commented):
Well this is passing all our testing, which include filtering on namespaces. The extension test case also passes, where we deal with extensions separately from the `pgcopydb clone` command.

The trick here is to be compatible with a target database where the extensions have already been installed in a previous step, out of the pgcopydb scope. What happens with your patch if you create the extension (and its schema) manually in the target instance before running `pgcopydb clone`?

kbarber:
> What happens with your patch if you create the extension (and its schema) manually in the target instance before running pgcopydb clone?

So if I use --drop-if-exists, it seems to work.

If I do not, I get this error:

```
16:36:53.636 21 ERROR  pg_restore: error: could not execute query: ERROR:  schema "foo" already exists
16:36:53.636 21 ERROR  Command was: CREATE SCHEMA foo;
16:37:00.019 21 WARN   pg_restore: warning: errors ignored on restore: 1
16:37:00.027 21 ERROR  Failed to run pg_restore: exit code 1
16:37:00.027 21 ERROR  Failed to prepare schema on the target database, see above for details
16:37:00.027 21 ERROR  Failed to clone source database, see above for details
16:37:00.119 10 ERROR  clone process 21 has terminated [6]
```

kbarber:
> I think your test fails because of the code you deleted that handles pre-existing SQL objects in the target database... we need to keep that feature when adding your bug fix.

So it sounds like the desire is: We need to be able to support copying extensions and having their schemas created if they don't exist, if the schema exists we shouldn't throw an error.

Is this because we imagine some users to want to precreate their schemas before a clone?

This could be done, I mean create schema if not exists is a thing, but pg_dump I'm not sure it can drop that. We could pull out the pre-schema creation too I guess and make it more programmatic like the `copy extensions` style.

Although I am wondering about the consistency here: This seems contrary to what I'm seeing for tables however, which will always fail if you try to clone again without --drop-if-exists. For example:

```
18:32:09.409 65 ERROR  string_utils.c:669        pg_restore: error: 
18:32:09.409 65 ERROR  string_utils.c:669        could not execute query: ERROR:  relation "foo" already exists
18:32:09.409 65 ERROR  string_utils.c:669        Command was: CREATE TABLE public.foo (
18:32:09.409 65 ERROR  string_utils.c:669            id integer
18:32:09.409 65 ERROR  string_utils.c:669        );
```

kbarber:
I added a `--skip-schemas` switch that reverts that filter that was originally there. Now you can run it with schemas & extensions already precreated. LMK if this is a viable approach.

kbarber:
@dimitri do you think this is enough for now for just the `pgcopy clone` case, or we need more work

src/bin/pgcopydb/catalog.c
```diff
@@ -4997,6 +4993,30 @@ catalog_prepare_filter(DatabaseCatalog *catalog,
 		}
 	}
 
+	/*
+	 * Implement --skip-schemas
+	 */
```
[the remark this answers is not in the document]
kbarber: Add a better comment

docs/include/clone.rst
```diff
@@ -23,6 +23,7 @@
      --skip-ext-comments           Skip restoring COMMENT ON EXTENSION
      --skip-collations             Skip restoring collations
      --skip-vacuum                 Skip running VACUUM ANALYZE
+     --skip-schemas                Skip creating schemas
```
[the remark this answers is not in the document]
kbarber: Improved the text here, am I missing a doc update somewhere for a longer description?

kbarber:
@dimitri how about `--skip-ext-schemas`? I've expanded the descriptions also.

kbarber:
Greetings @dimitri would you be able to take a look at this PR again when you have time? Thanks.

docs/include/clone.rst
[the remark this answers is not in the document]
dimitri: See https://pgcopydb.readthedocs.io/en/latest/ref/pgcopydb_clone.html#options

Also would be good if the options for skipping extension related stuff (comments, schemas) were grouped together in the listing.

dimitri:
> So it sounds like the desire is: We need to be able to support copying extensions and having their schemas created if they don't exist, if the schema exists we shouldn't throw an error.

Yes, please have a look at the following part of the code:

https://github.com/dimitri/pgcopydb/blob/3f07e0ba0f111abd31b0fa2bbb4c3529f491d99e/src/bin/pgcopydb/dump_restore.c#L661-L672

I would expect that we implement something similar for extensions schemas too, which means we populate the target database with the list of namespaces that extensions depend on. Do you think you could add that to your PR, basically making the `--skip-ext-schemas` fully automated now?

kbarber:
> > So it sounds like the desire is: We need to be able to support copying extensions and having their schemas created if they don't exist, if the schema exists we shouldn't throw an error.
> 
> Yes, please have a look at the following part of the code:
> 
> https://github.com/dimitri/pgcopydb/blob/3f07e0ba0f111abd31b0fa2bbb4c3529f491d99e/src/bin/pgcopydb/dump_restore.c#L661-L672
> 
> I would expect that we implement something similar for extensions schemas too, which means we populate the target database with the list of namespaces that extensions depend on. Do you think you could add that to your PR, basically making the `--skip-ext-schemas` fully automated now?

@dimitri thanks for responding. The code you pointed out seems to be doing almost the correct thing already, however: it doesn't match when the ownership of the schema between the source & target is not the same. So I was getting mismatches - if we reduce the lookup to ignore the ownership it works. I think I fixed this in the current PR, and this seems to work now without a switch.

src/bin/pgcopydb/catalog.c
```diff
@@ -5952,25 +5947,25 @@ catalog_add_s_namespace(DatabaseCatalog *catalog, SourceSchema *namespace)
...
-		log_error("BUG: catalog_lookup_s_namespace_by_rlname: db is NULL");
+		log_error("BUG: catalog_lookup_s_namespace_by_nspname: db is NULL");
 		return false;
 	}
 
 	char *sql =
 		"  select oid, nspname, restore_list_name "
 		"    from s_namespace "
-		"   where restore_list_name = $1 ";
+		"   where nspname is $1 ";
```
dimitri, line 5968: Why use the `is` operator instead of the equality operator `=`?

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -1326,6 +1326,27 @@ copydb_prepare_target_catalog(CopyDataSpec *specs)
...
+ * Convert schema restoreListName to schemaName
+ * The restoreListName is formatted as:
+ * - schema_name owner
+ * We want to extract just the schema_name part
+ */
+bool
+copydb_schema_name_from_restore_list_name(const char *restoreListName,
+										  char *schemaName)
+{
+	if (!(sscanf(restoreListName, "- %64s %*s", schemaName) == 1)) /* IGNORE-BANNED */
```
dimitri, line 1339: This is a very hopeful way to parse the restore_list_name. When a schema name includes double-quotes, these characters are escaped by being doubled, so we might have up to 128 characters in a schema name. The schema name might contain a space too. The owner name could contain a space also.

Does this archive Table of Contents line contain the schema OID? if yes, we should probably fill-in a SQLite catalog with the list of all the source schemas and then implement an OID lookup in that internal catalog. The way restore_list_name are built is not machine-parsable in the general case.

dimitri (changes requested):
Nice! Thanks for the detective work. I like that you found that we match on schema name and owner and should only match on schema name. About the implementation, see my comments. Is there a way that we can lookup the actual schema name from the schema OID, if we have it, rather than parse the restore list name?

src/bin/pgcopydb/copydb_schema.c
kbarber: Yeah, I was worrying about it too.

It completely fails to skips anything with a space, this is a toc sample I was looking at from pre-filtered.list.

```
;12; 2615 18530 SCHEMA - aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaabcdefghijklmnop fqayuutgtrkspv
;11; 2615 18529 SCHEMA - aabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghik fqayuutgtrkspv
13; 2615 18531 SCHEMA - abcdefghijklmnopqrstuv wxyz fqayuutgtrkspv
;10; 2615 18528 SCHEMA - abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghik fqayuutgtrkspv
```

There is the OID there, 18530, 18529, 18531.

I presume what you mean is: do a lookup on the source catalog for the oids from s_namespace retrieving the nspame and then use _that_ to lookup the target nspname finally by oid.

dimitri: The lookup should be done in the SQLite catalogs, and for that we need to create and fill-in a new `s_namespace` catalog table in the sourceDB. Once we have that we can lookup the SCHEMA entries by OID in the sourceDB to fetch the schema name (instead of parsing it from the archive contents), and then we can lookup the schema by name in the target database using what you did already.

So yes what you said, with the explicit caching of the whole pg_namespace catalog from the source database in our internal SQLite catalogs.

src/bin/pgcopydb/catalog.c
kbarber: Ooops fixed.

src/bin/pgcopydb/schema.h
```diff
@@ -354,10 +354,11 @@ typedef enum
 	DATA_SECTION_FILTERS,
 	DATA_SECTION_BLOBS,
 	DATA_SECTION_VACUUM,
-	DATA_SECTION_ALL
+	DATA_SECTION_ALL,
+	DATA_SECTION_NAMESPACES
 } CopyDataSection;
 
-#define DATA_SECTION_COUNT (DATA_SECTION_ALL + 1)
+#define DATA_SECTION_COUNT (DATA_SECTION_NAMESPACES + 1)
```
dimitri, line 361: Please keep DATA_SECTION_ALL as the last entry in the list, and then DATA_SECTION_COUNT definition does not need to change either. I think I would insert DATA_SECTION_NAMESPACES just before DATA_SECTION_TABLE_DATA in the enum.

dimitri (changes requested):
Wow. Almost there. I added a nitpick change that I think is necessary, and it looks like the last one needed here in this PR!

kbarber:
> Wow. Almost there. I added a nitpick change that I think is necessary, and it looks like the last one needed here in this PR!

@dimitri Yeah I was wondering about the ordering myself.

Okay I've fixed that, rebased, squashed and the tests are passing again.

dimitri:
> @dimitri okay the code is now retrieving the namespaces for the source catalog, and registering the section as it expects so that commands can be repeated safely. It uses that information to lookup the nspname from source catalog using the oid from the archive list, then uses that name on the target catalog to check if it already exists or not.

Last round of review now, I think we're only missing registering the cache-invalidation for the namespace list at this place in the code, where I would add the DATA_SECTION_NAMESPACES just before DATA_SECTION_TABLE_DATA:

https://github.com/dimitri/pgcopydb/blob/3f07e0ba0f111abd31b0fa2bbb4c3529f491d99e/src/bin/pgcopydb/copydb_schema.c#L233-L241

> Looks like its passing, I modified the test to include schemas with spaces to help with regression.

Thanks for that, that's good thinking!

dimitri:
Thanks for all the work @kbarber ; we end-up in a much different place than we started at, with a long way to get there, and the result is much better than what we started with. Now, I believe the other needed PR is going to be much simpler with just adding the SCHEMA in the CREATE EXTENSION command, I'm not sure anything else is needed really.

PR #722, opened 2024-03-26
Support pg 11 and 12

RekGRpth:
Support pg 11 and 12

Postgres 11 and 12 has not PG_MAJORVERSION_NUM macro,
so this patch replaces it by PG_MAJORVERSION

### DIFF
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -1913,9 +1913,8 @@ pgsql_enable_pipeline_mode(PGSQL *pgsql)
 	if (!warned)
 	{
 		log_warn("Skipping libpq pipeline mode optimisation because pgcopydb "
-				 "was built with libpq %d, pipeline mode is available since "
-				 "libpq 14",
-				 PG_MAJORVERSION);
+				 "was built with libpq " PG_MAJORVERSION
+				 ", pipeline mode is available since libpq 14");
 
 		warned = true;
 	}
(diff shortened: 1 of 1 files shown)


src/bin/pgcopydb/pgsql.c
```diff
@@ -1913,9 +1913,8 @@ pgsql_enable_pipeline_mode(PGSQL *pgsql)
 	if (!warned)
 	{
 		log_warn("Skipping libpq pipeline mode optimisation because pgcopydb "
-				 "was built with libpq %d, pipeline mode is available since "
-				 "libpq 14",
-				 PG_MAJORVERSION_NUM);
+				 "was built with libpq " PG_MAJORVERSION
```
arajkumar, line 1916: I'm wondering what is the purpose of this change? PG_MAJORVERSION_NUM is an integer and we were using %d format specifier to print, I don't see any flaw with that.

RekGRpth: support postgres 11 and 12

arajkumar: `PG_MAJORVERSION_NUM` have been used in other places too. I'm not sure why the problem is not pronounced there?

https://github.com/search?q=repo%3Adimitri%2Fpgcopydb%20PG_MAJORVERSION_NUM&type=code

It would be great if you can write a description of why you need this change along with the problem.

Thank you.

arajkumar: Ok, after greping into PG source, I see that `PG_MAJORVERSION_NUM` doesn't exists on PG11 and it has only `PG_MAJORVERSION`. Please update PR title and commit message to reflect the intent of this change.

dimitri (approved):

src/bin/pgcopydb/pgsql.c
RekGRpth: > `PG_MAJORVERSION_NUM` have been used in other places too. I'm not sure why the problem is not pronounced there?

I don't understand, why problem is not appeared in other places, but I prefer change them to universal check like
```c
#if PG_VERSION_NUM >= 150000
```

PR #725, opened 2024-03-27
Problem: Extension configurations are copied partially

arajkumar:
I'm keeping it as draft, I'm still not sure whether we should copy all extension config tables under single txn or open a txn per extension.


dimitri:
Closed by PR #990 (shipped in v0.18): all config tables for a given extension are now copied inside a single transaction on the target, settling the question raised here. The summary table tracks done state so a crashed or interrupted run skips already-copied extensions on resume instead of re-running COPY and producing duplicate rows.

PR #726, opened 2024-03-28
Check configuration on target before applying db properties.

VaibhaveS:
Certain configurations on source might not exist in the target server. Before applying the configuration we need to check if it exists.

**before**
```
2024-03-27 18:43:21 64 ERROR pgsql.c:1978 [TARGET 29176] ERROR: unrecognized configuration parameter "sql_inheritance"
2024-03-27 18:43:21 64 ERROR pgsql.c:1989 [TARGET 29176] SQL query: ALTER DATABASE postgres SET "sql_inheritance" TO 'off';
```
**after**
```
12:16:53.589 16592 SQL    pgsql.c:1504              [TARGET 12718] select exists(select name from pg_settings WHERE name = $1);
12:16:53.589 16592 SQL    pgsql.c:1517              [TARGET 12718] 'sql_inheritance'
12:16:53.810 16592 WARN   dump_restore.c:363        Skipping database property sql_inheritance=off which does not exists on the target database
```

### DIFF
--- a/src/bin/pgcopydb/copydb_schema.c
+++ b/src/bin/pgcopydb/copydb_schema.c
@@ -180,7 +180,7 @@ copydb_fetch_schema_and_prepare_specs(CopyDataSpec *specs)
 
 /*
  * copydb_fetch_source_catalog_setup initializes our local catalog cache and
- * checks th setup and cache state.
+ * checks the setup and cache state.
  */
 static bool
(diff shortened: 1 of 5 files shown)


dimitri (approved):
We might want another pass at the topic to check all parameters in a single SQL query, but maybe we don't need that optimization at all.

PR #727, opened 2024-03-28
Problem: replay_lsn is not moving until the first message received

arajkumar:
When we start with an idle source, pgcopydb doesn't send generate keep alive message until it receives message from the source.

pgsql_stream_logical calls [flushFunction](https://github.com/dimitri/pgcopydb/blob/e1dcb60b8f007179536c85a34ab5ce902d932497/src/bin/pgcopydb/pgsql.c#L4294-L4306), however it has been ignored by the implementation because there is [no jsonFile is created yet](https://github.com/dimitri/pgcopydb/blob/e1dcb60b8f007179536c85a34ab5ce902d932497/src/bin/pgcopydb/ld_stream.c#L1406-L1410).

Solution: Don't check for the existence of jsonFile, which leads to generation of keep alive messages causing replay_lsn to move forward.

Related to https://github.com/dimitri/pgcopydb/pull/703.

### DIFF
--- a/src/bin/pgcopydb/ld_stream.c
+++ b/src/bin/pgcopydb/ld_stream.c
@@ -1403,12 +1403,6 @@ streamFlush(LogicalStreamContext *context)
 {
 	StreamContext *privateContext = (StreamContext *) context->private;
 
-	/* when there is currently no file opened, just skip the flush operation */
-	if (privateContext->jsonFile == NULL)
-	{
-		return true;
-	}
-
 	log_debug("streamFlush: %X/%X %X/%X",
 			  LSN_FORMAT_ARGS(context->tracking->written_lsn),
 			  LSN_FORMAT_ARGS(context->cur_record_lsn));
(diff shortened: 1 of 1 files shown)


src/bin/pgcopydb/ld_stream.c
```diff
@@ -1403,12 +1403,6 @@ streamFlush(LogicalStreamContext *context)
 {
 	StreamContext *privateContext = (StreamContext *) context->private;
 
-	/* when there is currently no file opened, just skip the flush operation */
-	if (privateContext->jsonFile == NULL)
-	{
-		return true;
-	}
-
```
dimitri, line 1411: I think this only works because the later call to `streamKeepalive()` is in-turn going to call `streamRotateFile()` which ensures that a JSON file is now opened. Please add a comment either before calling `streamKeepalive()` or before calling `fileno` explaining this detail.

dimitri (changes requested):

dimitri (approved):

Issue #666, opened 2024-02-05
pg_dump: error: server version: 14.10.0; pg_dump version: 13.13 (Debian 13.13-0+deb11u1)


andres-chavez-bi:
Hello, I'm testing the docker image and also the package for the tool, from postgres red hat repository, and I get the same error while executing from a RHEL 8.6 (either from the image or the package, the error is the same).

Here's the sequence: 
I'm trying to run the v0.15 on a Red Hat machine, as per instructions, binaries are included in the postgres repository. I've installed the repo and specifically this package: 

```
# dnf search --all --quiet pgcopydb
=============================================================== Name & Description & URL Matched: pgcopydb ================================================================
pgcopydb.x86_64 : Automate pg_dump | pg_restore between two running Postgres servers
pgcopydb_12.x86_64 : Automate pg_dump | pg_restore between two running Postgres servers
pgcopydb_13.x86_64 : Automate pg_dump | pg_restore between two running Postgres servers
pgcopydb_14.x86_64 : Automate pg_dump | pg_restore between two running Postgres servers
pgcopydb_15.x86_64 : Automate pg_dump | pg_restore between two running Postgres servers

```

```
dnf list installed pgcopydb*
Updating Subscription Management repositories.
Installed Packages
pgcopydb.x86_64                                                                              0.13-1PGDG.rhel8                                                                            @pgdg16
pgcopydb_12.x86_64                                                                           0.10-1.rhel8                                                                                @pgdg12
pgcopydb_13.x86_64                                                                           0.10-1.rhel8                                                                                @pgdg13
pgcopydb_14.x86_64                                                                           0.10-1.rhel8                                                                                @pgdg14
pgcopydb_15.x86_64                                                                           0.10-1.rhel8                                                                                @pgdg15
```

![image](https://github.com/dimitri/pgcopydb/assets/134699002/6ac50108-0393-4dbd-b31b-4531f9db7466)

```
# pgcopydb --version
10:07:14 1 INFO   Running pgcopydb version 0.14 from "/usr/local/bin/pgcopydb"
pgcopydb version 0.14
compiled with PostgreSQL 13.11 (Debian 13.11-0+deb11u1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
compatible with Postgres 11, 12, 13, 14, 15, and 16
```

First thing I see is that the version of the tool, does not match either the repo installed (in the step before) and the version used:

[...]

PR #735, opened 2024-04-04
Dockerfile: make pg_dump more interoperable

toxeus:
Before, `pg_dump` would refuse to talk to Postgres versions higher than version 14 because it allows only equal or lower major versions. Therefore, we have bumped the Postgres version to 16 to allow `pg_dump` to connect to all available Postgres versions.

@DimCitus [wrote](https://github.com/dimitri/pgcopydb/issues/666#issuecomment-1980917660):
> We could improve the main Dockerfile to use that trick and publish docker images with Postgres 16 by default I suppose.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ FROM --platform=${TARGETPLATFORM} debian:11-slim as build
 ARG TARGETPLATFORM
 ARG TARGETOS
 ARG TARGETARCH
-ARG PGVERSION=14
+ARG PGVERSION=16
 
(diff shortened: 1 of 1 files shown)


dimitri (dismissed):

dimitri:
Please review build errors and fix the following:

``` 
 #19 [build 8/8] RUN make -s clean && make -s -j$(nproc) install
#19 33.98 /usr/bin/ld: cannot find -lzstd
```

toxeus:
@dimitri done

dimitri (approved):

PR #737, opened 2024-04-09
Implement support for filtering materialized view and data (refresh).

arajkumar:
Currently, there is no way to filter materialized view or materialized view data(refresh).

pg_dump supports it with following option,

- `pg_dump --exclude-table` filters materialized view altogether.
- `pg_dump --exclude-table-data` filters REFRESH MATERIALIZED VIEW.

Solution: Incorporate materialized view filtering into existing filtering framework. [exclude-table], [include-table] can be used filter the materialized view and [exclude-table-data] can be used to filter materialized view data(refresh).


arajkumar:
Tests are failing with the following message.,

> The action 'Run a test' has timed out after 5 minutes.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -495,6 +495,56 @@ pg_dump_db(PostgresPaths *pgPaths,
...
+			}
+
+			sformat(excludeTableData[i], PG_NAMEDATALEN_FQ, "%s.%s",
+					table->nspname,
+					table->relname);
+
+			/* check that we still have room for --exclude-schema args */
+			if (PG_CMD_MAX_ARG < (argsIndex + 2))
+			{
+				log_error("Failed to call pg_dump, too many exclude-schema entries: "
```
dimitri, line 536: Fix name to exclude-table-data.

dimitri (changes requested):
Thanks, almost clicked merge on first review ;-)

dimitri:
I'm now wondering about using `--exclude-table-data`. Sorry didn't think about it before.

Would it make more sense to filter REFRESH commands in the Archive Table of Contents instead? Can we reduce the option for only the materialized views otherwise? We have users with so many table (millions) that we want to avoid cluttering the command line if possible.

arajkumar:
@dimitri 

> I'm now wondering about using --exclude-table-data. Sorry didn't think about it before.

Yeah, I thought of it, but I felt the approach used in this PR is quite easy to implement than filtering Archive Table of Contents. 

> Can we reduce the option for only the materialized views otherwise?

Could you please explain more? 

Do you mean to introduce a dedicated filtering section only for materialized views?
e.g.
```.ini
[exclude-materialized-view-refresh]
schema1.view1
schema2.view2
```

or

Do you mean to introduce a flag which automatically skips all matview refresh?
e.g.
```
pgcopydb clone --skip-materialized-view-refresh
```

> We have users with so many table (millions) that we want to avoid cluttering the command line if possible.

I'm wondering this would be an issue only when the user wants to exclude copying table data for lots(millions) of tables right? How common is that case?

dimitri:
> > I'm now wondering about using --exclude-table-data. Sorry didn't think about it before.
> 
> Yeah, I thought of it, but I felt the approach used in this PR is quite easy to implement than filtering Archive Table of Contents.

I see. For filtering at pg_restore level using the Archive Table of Contents we need a filter in `copydb_write_restore_list` for ARCHIVE_TAG_REFRESH_MATERIALIZED_VIEW entries. We need to see if these entries contain the OID of the MATVIEW only, or also the nspname and relname, and we need to match that name to the filtering setup.

So in terms of complexity it's just another lookup in that loop. In terms of code to write, we might need a lot of infrastructure, maybe even a new SQLite catalog table (s_matviews) with the corresponding Postgres catalogs query  and SQLite lookup function.

> > Can we reduce the option for only the materialized views otherwise?
> 
> Could you please explain more?

What I mean is that if the `[exclude-table-data]` filter contains many tables and a few matview, is there a way to only output the pg_dump command line arguments for the matviews in there, skipping the plain tables?

There is a Unix limit on the length of a command line and it's not very large (use `getconf ARG_MAX` on Unix systems to discover that limit in bytes).

arajkumar:
> We need to see if these entries contain the OID of the MATVIEW only, or also the nspname and relname, and we need to match that name to the filtering setup.

@dimitri Archive TOC contains both oid and nspname + relname.

```
6712; 0 93394 MATERIALIZED VIEW DATA public metrics_count tsdbadmin
```

Do you think we could simply filter them out by comparing nspname + relname with the [exclude-table-data] list and ignore them instead of going through the complex filtering setup?

dimitri:
> > We need to see if these entries contain the OID of the MATVIEW only, or also the nspname and relname, and we need to match that name to the filtering setup.
> 
> @dimitri Archive TOC contains both oid and nspname + relname.
> 
> ```
> 6712; 0 93394 MATERIALIZED VIEW DATA public metrics_count tsdbadmin
> ```
> 
> Do you think we could simply filter them out by comparing nspname + relname with the [exclude-table-data] list and ignore them instead of going through the complex filtering setup?

Well apparently, if I read correctly, the OID is zero, which means we can't lookup by OID here. We have to lookup by restore_list_name, which unfortunately isn't machine friendly. That's because characters such as `\n` are replaced with spaces, and spaces are not escaped, no quoting is used, and the separator between the three fields is also a space.

With that in mind, I think the best approach here is to fill-in a new s_matviews catalog in the sourceDB, where we store the list of materialized views from the source database, including the computed `restore_list_name` field. We can then add that to the filterDB filter table and would not need to change the lookup bits I believe.

What do you think?

arajkumar:
> ```
> 6712; 0 93394 MATERIALIZED VIEW DATA public metrics_count tsdbadmin
> ```

@dimitri IIUC, catalog id is 0 and oid is 93394. The below is a result of querying pg_class using the oid.

```
select * from pg_class where oid=93394;

-[ RECORD 1 ]-------+--------------
oid                 | 93394
relname             | metrics_count
relnamespace        | 2200
reltype             | 93396
reloftype           | 0
relowner            | 16418
relam               | 2
relfilenode         | 93418
reltablespace       | 0
relpages            | 0
reltuples           | -1
relallvisible       | 0
reltoastrelid       | 0
relhasindex         | f
relisshared         | f
relpersistence      | p
relkind             | m
relnatts            | 1
relchecks           | 0
relhasrules         | t
relhastriggers      | f
relhassubclass      | f
relrowsecurity      | f
relforcerowsecurity | f
relispopulated      | t
relreplident        | d
relispartition      | f
relrewrite          | 0
relfrozenxid        | 44467302
relminmxid          | 1586679
relacl              |
reloptions          |
relpartbound        |
```

dimitri:
> @dimitri IIUC, catalog id is 0 and oid is 93394. The below is a result of querying pg_class using the oid.

Ah nice, okay, we can indeed do a lookup by OID then. I know you wanted to reduce the amount of code to write, but I still think that the robust approach is to implement that filtering the same way as we implement the table filtering, with `s_matviews` and then just an OID lookup in the `filter` table in `filterDB`.

arajkumar:
@dimitri Modified the implementation to use filter catalog to filter matview and refresh.

docs/ref/pgcopydb_config.rst
```diff
@@ -78,6 +78,8 @@ configuration then the sections ``exclude-schema`` and ``exclude-table`` are
 disallowed. We would not know how to handle tables that exist on the source
 database and are not part of any filter.
 
+NOTE: Materialized views are also considered as tables during the filtering.
+
```
dimitri, line 82: I think this only makes sense if the existing `[exclude-table]` section allows filtering-out materialized views. Is that the case? (can't remember now and being lazy).

src/bin/pgcopydb/catalog.c
```diff
@@ -4819,6 +4894,16 @@ catalog_prepare_filter(DatabaseCatalog *catalog,
 
 		"  union all "
 
+		/* we need to filter-out matviews and matview_refresh separately */
+		"	 select oid, restore_list_name, "
+		"		case exclude_data "
+		"          when 0 then 'matview' "
+		"		   when 1 then 'matview_refresh' "
+		"       end "
+		"	   from s_matview"
```
dimitri, line 4903: That feels strange, more on that later.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -711,7 +711,30 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
+		 * dump, but the refresh is filtered out using [exclude-table-data].
+		 * In this case, we want to skip the refresh.
+		 */
+		if (!skip &&
+			(item->desc == ARCHIVE_TAG_REFRESH_MATERIALIZED_VIEW ||
+			 item->desc == ARCHIVE_TAG_MATERIALIZED_VIEW))
+		{
+			bool isRefresh = item->desc == ARCHIVE_TAG_REFRESH_MATERIALIZED_VIEW;
+
+			if (copydb_matview_objectid_is_filtered_out(specs, oid, isRefresh))
```
dimitri, line 725: I was expecting a special processing for `[exclude-table-data]` filtering when the tag is about REFRESH, but the rest of the processing to just be as usual, covered in the filter table and the `copydb_objectid_is_filtered_out()` function call.

docs/ref/pgcopydb_config.rst
arajkumar: @dimitri , originally it wasn't supported by pgcopydb, but this PR adds the capability to filter matview using [exclude-table].

src/bin/pgcopydb/dump_restore.c
arajkumar: The reason for having special handling for ARCHIVE_TAG_MATERIALIZED_VIEW is that we also need to exclude ARCHIVE_TAG_REFRESH_MATERIALIZED_VIEW when the materialized view itself is filtered out.

dimitri: Yes that's what I meant. In that case when the matview itself is filtered out, certainly the existing function and filter table are covering all we need, right?

arajkumar: The existing function `copydb_objectid_is_filtered_out` filters based on oid and never it considers the fact whether it is a matview itself or just matview refresh.

dimitri: Exactly. So first do a special check for REFRESH, and then let the normal code behave as usual for the CREATE MATERIALIZED VIEW, skipping that only when the OID is found in our filter table. Which means when skipping only the REFRESH parts, we do NOT add the OID of the matview to the filter table.

It seems to me that would be easier to understand and maintain.

arajkumar: @dimitri IIUC, you are suggesting not to insert the OID of the materialized view into the filter table if it's data is excluded. Instead, refer to s_matview table to filter the materialized view refresh?

dimitri: Indeed. Keep the current behavior, just add CREATE MAT VIEWS to the filtering when needed, and only add a specific new behavior for REFRESH, which is exclude-data, and new, and needs special care.

src/bin/pgcopydb/schema.c
```diff
@@ -925,6 +936,254 @@ schema_prepare_pgcopydb_table_size(PGSQL *pgsql,
...
+/*
+ * schema_list_matviews grabs the list of materialized view refreshes from the
+ * given source Postgres instance and allocates a SourceTable array with
+ * the result of the query.
+ */
+bool
+schema_list_matviews(PGSQL *pgsql,
+					 SourceFilters *filters,
+					 DatabaseCatalog *catalog)
+{
```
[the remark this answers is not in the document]
arajkumar: @dimitri , I think adding matview as part of s_table might cause complexities with table data copying, we have to exclude matviews. Let me do more study on this suggestion.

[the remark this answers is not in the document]
arajkumar: I think one of the most confusing part would be using source.db to filter matview refresh to support [exclude-table-data].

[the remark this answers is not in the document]
dimitri: I see your point, it's a good one. There is another way to approach it: can we add relkind = 'm' to the source table queries, and then register these entries in the `s_matviews` catalog?

At the following place, divert the call to add to catalog depending on the relkind?

https://github.com/dimitri/pgcopydb/blob/19795fca058a34292a9f15a8cd5e360d41bf46c0/src/bin/pgcopydb/schema.c#L4681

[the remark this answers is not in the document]
arajkumar: Yeah, I was adding relkind to differentiate tables from matview. But, didn't consider to put it into s_matview. You mean to create a s_matview on source catalog right?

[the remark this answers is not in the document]
arajkumar: The latest change reflects you last suggestion.  i.e. utilising the table query to populate s_matview.

src/bin/pgcopydb/schema.c
```diff
@@ -4769,9 +4777,28 @@ getTableArray(void *ctx, PGresult *result)
 
 		if (context->catalog != NULL && context->catalog->db != NULL)
 		{
-			if (!catalog_add_s_table(context->catalog, table))
+			if (table->relkind == 'r' || table->relkind == 'p')
 			{
```
dimitri, line 4781: Nitpicking: can we use a `swich(table->relkind)` statement? I find those easier to follow when maintaining the code.

dimitri (changes requested):
LGTM. Added a small comment to refactor if into switch, other than that it feels ready to merge. Will mark as such, please send a last round of update with the stylistic change if you agree with it!

src/bin/pgcopydb/schema.c
arajkumar: Fixed it. Thank you.

arajkumar:
> LGTM. Added a small comment to refactor if into switch, other than that it feels ready to merge. Will mark as such, please send a last round of update with the stylistic change if you agree with it!

@dimitri I really liked the way this PR shaped up :) Thanks a ton for your guidance.

dimitri (approved):

dimitri:
Uh oh, because of merge ordering you now have a conflict to merge in schema.c

src/bin/pgcopydb/catalog.c
```diff
@@ -243,6 +252,15 @@ static char *filterDBcreateDDLs[] = {
...
 
+	"create table s_matview("
+	"  oid integer primary key, "
+	"  qname text, nspname text, relname text, restore_list_name text, "
+	"  exclude_data boolean"
+	")",
+
+	"create unique index s_mv_rlname on s_matview(restore_list_name)",
+	"create unique index s_mv_qname on s_matview(nspname, relname)",
+
```
dimitri, line 263: I am not seeing where this is filled, do we still use it in the PR?

arajkumar: It would be populated when table complement filter operation runs (i.e. exclude-table or include-only-table)., i.e. the same function which populates this table on source.db also populates for filter.db during complement operation.

PR #751, opened 2024-04-18
Copy schemas/namespaces as a separate sub-command

kbarber:
This adds a new sub-command for copying schemas. Split from: https://github.com/dimitri/pgcopydb/pull/711


kbarber:
This is a little old and may not be useful any more. Closing it.

Issue #749, opened 2024-04-18
ERROR:  out of memory when calling schema_list_extensions()


kbarber:
When calling a `pgcopydb clone` I occasionally see the following error occur. It always happens to the same set of DBs:

```
05:33:10.264 792 ERROR  pgsql.c:2314              [SOURCE 24800] ERROR:  out of memory
05:33:10.264 792 ERROR  pgsql.c:2314              [SOURCE 24800] DETAIL:  Failed on request of size 6442450944 in memory context "ExecutorState".
05:33:11.240 792 ERROR  pgsql.c:2314              [SOURCE 24800] ERROR:  out of memory
05:33:11.241 792 ERROR  pgsql.c:2325              [SOURCE 24800] SQL query: with recursive fk_constraints as (      select conrelid, confrelid from pg_constraint where contype = 'f' and conrelid <> confrelid  ), raw_ordered_fk_constraints as (      select             distinct c.confrelid as relid,             0 as depth        from fk_constraints ...
```

Which is the SQL here inside `schema_list_extensions()`: https://github.com/dimitri/pgcopydb/blob/main/src/bin/pgcopydb/schema.c#L425-L484

I notice it on databases with FK constraints mostly, at first I thought it was due to a large number of constraints (I first saw this on a DB with 1000's of them), but I've reproduced the bug now with only 3 tables & these 3 FK constraints:

```
create table one (id int, two_id int);
create table two (id int, one_id int, three_id int);
create table three (id int);
alter table one add constraint one_pk primary key (id);
alter table two add constraint two_pk primary key (id);
alter table three add constraint three_pk primary key (id);
alter table one add constraint one_two_fk foreign key (two_id) references two(id);
alter table two add constraint two_one_fk foreign key (one_id) references one(id);
alter table two add constraint two_three_fk foreign key (three_id) references three(id);
```

I think there is an infinite recursion causing this, at least its very expensive for just 3 FK constraints.

kbarber:
I've put up a draft PR for perusal here: https://github.com/dimitri/pgcopydb/pull/752

PR #702, opened 2024-03-12
Respect FK constraints for extension config tables

JamesGuthrie:
Extension configuration tables can have foreign key dependencies between them. If the configuration tables are restored in the incorrect order, it can cause foreign key constraint violations.

This change ensures that the configuration tables are ordered according to the dependency order in `pg_constraint`.

I don't expect that this change will be merged as-is, but I've tested it for our use-case and it seems to work.

Fixes #701

Issue #701, opened 2024-03-12
pgcopydb doesn't respect foreign key dependencies for extension configuration tables


JamesGuthrie:
Extension configuration tables can have foreign key dependencies between them. If the configuration tables are restored in the incorrect order, it can cause foreign key constraint violations.

pgcopydb (implicitly?) [assumes](https://github.com/dimitri/pgcopydb/blob/ff6b47389fd68447ad9e2039a418b97fce9e3186/src/bin/pgcopydb/schema.c#L426-L450) that the ordering of the oids in `pg_extension.extconfig` is sufficient, but it is not (see pg_dump's [implementation](https://github.com/postgres/postgres/blob/539e328b1c3a4047c7ff525524edbcca2bac6e10/src/bin/pg_dump/pg_dump.c#L17858-L17880)).

## Steps to reproduce

Note: this requires the extension modifying its own catalog to get the config tables in a weird order, so these steps to reproduce must be followed exactly.

```
# setup source db
docker run -d -e POSTGRES_HOST_AUTH_METHOD=trust -p 5432:5432 timescale/timescaledb-ha:pg15

# setup target db
docker run -d -e POSTGRES_HOST_AUTH_METHOD=trust -p 5433:5432 timescale/timescaledb-ha:pg15

# load specific version (these specific versions modify the catalog in a way that illustrates this issue)
psql postgres://postgres@localhost:5432 -c "DROP EXTENSION timescaledb;"
psql postgres://postgres@localhost:5432 -c "CREATE EXTENSION timescaledb VERSION '2.11.2';"
psql postgres://postgres@localhost:5432 -c "ALTER EXTENSION timescaledb UPDATE TO '2.12.0';"

# correctly set up target
psql postgres://postgres@localhost:5432 -c "DROP EXTENSION timescaledb;"
psql postgres://postgres@localhost:5432 -c "CREATE EXTENSION timescaledb VERSION '2.12.0';"

# ensure that there's data to cause a FK violation
psql postgres://postgres@localhost:5432 -c "CREATE TABLE metrics(time timestamptz, value bigint); SELECT create_hypertable('metrics', 'time');"

# migrate
docker run --network=host -ti dimitri/pgcopydb:v0.15 pgcopydb copy extensions --source postgres://postgres@localhost:5432 --target postgres://postgres@localhost:5433 --verbose
```

Output
```
12:40:35.014 1 INFO   Running pgcopydb version 0.15 from "/usr/local/bin/pgcopydb"
[...]

PR #752, opened 2024-04-19
Fix infinite recursion for schema_list_extensions() catalog query

kbarber:
Fixes: https://github.com/dimitri/pgcopydb/issues/749

This PR: https://github.com/dimitri/pgcopydb/pull/702 introduced important ordering for extension installation that took into account relationships between extension configuration tables.

The new query executed by `schema_list_extensions()` seems to have an infinite loop. You can reproduce this by running a `pgcopydb clone` with the follow tables:

```
create table one (id int, two_id int);
create table two (id int, one_id int, three_id int);
create table three (id int);
alter table one add constraint one_pk primary key (id);
alter table two add constraint two_pk primary key (id);
alter table three add constraint three_pk primary key (id);
alter table one add constraint one_two_fk foreign key (two_id) references two(id);
alter table two add constraint two_one_fk foreign key (one_id) references one(id);
alter table two add constraint two_three_fk foreign key (three_id) references three(id);
```

This patch attempts to remove the loop by adding cycle detection as documented as a pattern here: https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-CYCLE it also adds the above tables & FK creations to extensions/copydb.sh to test for regression in the future.


kbarber:
@JamesGuthrie  if you have any strong feelings about this fix, or a better idea please let me know :-).

I was thinking of adding the tests that James described here: https://github.com/dimitri/pgcopydb/issues/701 to avoid regression, looks like it needs the TimescaleDB module.

tests/extensions/copydb.sh
```diff
@@ -33,6 +33,20 @@ create extension intarray cascade;
...
+psql -a ${PGCOPYDB_SOURCE_PGURI} <<EOF
+create table one (id int, two_id int);
+create table two (id int, one_id int, three_id int);
+create table three (id int);
+alter table one add constraint one_pk primary key (id);
+alter table two add constraint two_pk primary key (id);
+alter table three add constraint three_pk primary key (id);
+alter table one add constraint one_two_fk foreign key (two_id) references two(id);
+alter table two add constraint two_one_fk foreign key (one_id) references one(id);
+alter table two add constraint two_three_fk foreign key (three_id) references three(id);
```
dimitri, line 47: Can we inline the pkey and fkey definitions into the CREATE TABLE statements? I think it's easier to read. It needs be, let's just add another SQL file to that test unit (such as `ddl.sql` as found in other units) to have more flexibility in the SQL syntax/layout.

dimitri (approved):
Thanks for the fix! I think we could maybe improve on the spelling of the test case, so while I could merge as-is, if you think my proposal is an improvement would you be up for another revision before merge?

tests/extensions/ddl.sql
```diff
@@ -0,0 +1,15 @@
...
+---
+--- This file implements DDL changes in the test database.
+
+begin;
+-- Create new tables with foreign key relationships to test
+-- the query for `schema_list_extensions()`
+create table one (id int primary key, two_id int);
+create table two (id int primary key, one_id int, three_id int);
+create table three (id int primary key);
+alter table one add constraint one_two_fk foreign key (two_id) references two(id);
```
kbarber, line 12: I left the FK's in their own lines, ordering makes this harder to do all in one because of the loop.

kbarber:
@dimitri 'twas a fun problem, let me know if you need anything else on this one.

tests/extensions/ddl.sql
dimitri: I see, makes sense!

dimitri (approved):

PR #769, opened 2024-05-08
Freeze only if we TRUNCATE

### LINKED ISSUE #765: cannot perform COPY FREEZE because the table was not created or truncated in the current subtransaction
Try a migration where truncate permission was missing for the migration user. 
Migration failed with error:cannot perform COPY FREEZE because the table was not created or truncated in the current sub transaction.

The bug must have been introduced in this commit - https://github.com/dimitri/pgcopydb/commit/1acbff9bd73ff426b849ac5f64ddcefb012add87

LOGS:

> 

```
MessageString
2024-04-25 05:54:23.313 79 SQL    pgsql.c:495               Connecting to [target] "xxxxx"
2024-04-25 05:54:23.314 70 SQL    pgsql.c:1542              [TARGET 32332] select has_table_privilege($1, $2);;
2024-04-25 05:54:23.314 70 SQL    pgsql.c:1555              [TARGET 32332] 'public.test_truncate_table', 'TRUNCATE'
2024-04-25 05:54:23.322 79 SQL    pgsql.c:1542              [TARGET 32335] BEGIN;
2024-04-25 05:54:23.322 79 SQL    pgsql.c:1542              [TARGET 32335] COMMIT;
2024-04-25 05:54:23.322 79 SQL    pgsql.c:404               Disconnecting from [target] "xxxxx"
2024-04-25 05:54:23.344 70 SQL    pgsql.c:1542              [TARGET 32332] BEGIN;
2024-04-25 05:54:23.345 70 NOTICE pgsql.c:2444              COPY public.test_truncate_table
2024-04-25 05:54:23.345 70 SQL    pgsql.c:2824              copy (SELECT id FROM ONLY public.test_truncate_table) to stdout;
2024-04-25 05:54:23.347 70 SQL    pgsql.c:2824              copy public.test_truncate_table(id) from stdin with (freeze);
2024-04-25 05:54:23.347 70 ERROR  pgsql.c:2874              [TARGET 32332] [55000] ERROR:  cannot perform COPY FREEZE because the table was not created or truncated in the current subtransaction
2024-04-25 05:54:23.347 70 ERROR  pgsql.c:2889              [TARGET 32332] Context: Failed to copy data to target
```

VaibhaveS:
Fixes #765


src/bin/pgcopydb/table-data.c
```diff
@@ -1161,6 +1161,9 @@ copydb_table_create_lockfile(CopyDataSpec *specs,
 		}
 
 		args->truncate = granted;
+
+		/* cannot perform COPY FREEZE if the table was not created or truncated in the current subtransaction */
+		args->freeze &= granted;
```
arajkumar, line 1166: Can we do this from [pgsql.c](https://github.com/dimitri/pgcopydb/blob/4f41dd430c8160876b89c26a99d81fc068d54c99/src/bin/pgcopydb/pgsql.c#L2737)? That would be a common place and applies to all usage of pg_copy?

hanefi: Although we can do this from pgsql.c, I believe this is a better solution. The function that builds the arguments should decide on the fields, and pgsql.c should execute them.

The comment block above suggests that we should decide whether we can benefit from `with (freeze)` in this file.

FWIW, the following patch seems to solve the issue as well, but I can not say that I like it more than this PR here.

```diff
diff --git a/src/bin/pgcopydb/pgsql.c b/src/bin/pgcopydb/pgsql.c
index 09c362c..c45229c 100644
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -3106,7 +3106,7 @@ pg_copy_send_query(PGSQL *pgsql, CopyArgs *args, ExecStatusType status)
             appendPQExpBuffer(sql, "copy %s from stdin", args->dstQname);
         }

-        if (args->freeze)
+        if (args->freeze && args->truncate)
         {
             appendPQExpBuffer(sql, " with (freeze)");
         }
```

hanefi (approved):
I tested and saw that the current patch solves the issue.

dimitri:
I think the solution needs to be implemented at the place where we manage the transaction. The FREEZE option is only available if we did TRUNCATE *in the same transaction*, which the caller does NOT control here. See:

https://github.com/dimitri/pgcopydb/blob/812a108976b43738ac81dd4fce060851d55076f8/src/bin/pgcopydb/pgsql.c#L2732-L2744

I think we should edit `args->freeze` from within the `if (args->truncate)` block here, because we just issued a BEGIN at that place. This requires some proper commenting, of course.

dimitri (approved):

PR #770, opened 2024-05-08
Add --disable-by-ctid

VaibhaveS:
- Added an option to disable split by CTID, since the partitioning based on CTID is slow in certain systems


docs/include/clone.rst
```diff
@@ -37,4 +37,5 @@
      --create-slot                 Create the replication slot
      --origin                      Use this Postgres replication origin node name
      --endpos                      Stop replaying changes when reaching this LSN
+     --disable-split-by-ctid	   	 Disable splitting tables by ctid
```
hanefi, line 40: Please keep the indentation consistent.

```suggestion
     --endpos                      Stop replaying changes when reaching this LSN
     --disable-split-by-ctid       Disable splitting tables by ctid
```

src/bin/pgcopydb/cli_copy.c
```diff
@@ -34,25 +34,26 @@ static CommandLine copy_db_command =
...
-		"  --no-comments         Do not output commands to restore comments\n"
-		"  --no-tablespaces      Do not output commands to select tablespaces\n"
-		"  --skip-large-objects  Skip copying large objects (blobs)\n"
-		"  --filters <filename>  Use the filters defined in <filename>\n"
-		"  --fail-fast           Abort early in case of error\n"
-		"  --restart             Allow restarting when temp files exist already\n"
-		"  --resume              Allow resuming operations after a failure\n"
-		"  --not-consistent      Allow taking a new snapshot on the source database\n"
-		"  --snapshot            Use snapshot obtained with pg_export_snapshot\n",
+		"  --source              		Postgres URI to the source database\n"
```
hanefi, line 37: Please do not change the spaces into tabs, and keep the format consistent as it was before.

hanefi: Reminder: You will need to update the documentation after fixing this.

hanefi (changes requested):
IMO, you should also update the logic in `schema_list_partitions` to disable splits by ctid if the flag or env is set.

docs/include/clone.rst
VaibhaveS: I ran the linter, I wonder how it did not fix this

hanefi: EditorConfig is not a great linter. It is better than having none, I suppose.

VaibhaveS:
> IMO, you should also update the logic in `schema_list_partitions` to disable splits by ctid if the flag or env is set.

I don't think this is required, because we will return from here..

```
	else if (!streq(source->amname, "heap"))
	{
		log_info("Table %s is %s large "
				 "which is larger than --split-tables-larger-than %s, "
				 "does not have a unique column of type integer, "
				 "and uses table access method \"%s\": "
				 "same table concurrency is not enabled",
				 source->qname,
				 source->bytesPretty,
				 specs->splitTablesLargerThan.bytesPretty,
				 source->amname);

		return true;
	}
```

src/bin/pgcopydb/cli_copy.c
VaibhaveS: I had to change it because the new argument was exceeding the length

dimitri:
Hi @VaibhaveS can you please resolve the merge conflicts with the current main? The addition of the new option `--skip-db-properties` seems to conflict with your PR. Also, please consider reviewing that other PR to inform some style issues that you have here. For instance, please add the option in the array before the trailing options (debug, verbose, help, etc).

VaibhaveS:
The branch name is incorrect. Please see https://github.com/dimitri/pgcopydb/pull/785

PR #775, opened 2024-05-13
Introduce new option for estimating table sizes

hanefi:
This PR introduces a new option --estimate-table-sizes. When set, pgcopydb will use the relpages value in system catalog to estimate the size of each table. This is done by multiplying the number of pages by
the default page size PostgreSQL uses (8KB).

Optionally, the user can set the environment variable
 `PGCOPYDB_ESTIMATE_TABLE_SIZES` to a boolean value to enable the option.

In passing, I removed `--cache`, `--disable-cache` parameters and removed references to `pgcopydb_table_size` table that is no longer used.

The only missing piece is to add `vacuumdb --analyze-only --jobs=<indexJobs>` before accessing relpages values in PG catalogs.


src/bin/pgcopydb/schema.c
```diff
@@ -848,13 +848,171 @@ struct FilteringQueries listSourceTableSizeSQL[] = {
...
- * on the given connection (typically, the source database). The creation is
- * skipped if the table already exists.
+ * For code simplicity the index array is also the SourceFilterType enum value.
+ */
+struct FilteringQueries listSourceTableSizeEstimateSQL[] = {
+	{
+		SOURCE_FILTER_TYPE_NONE,
+
+		"  select c.oid, c.relpages * 8192 as bytes, "
+		"         pg_size_pretty(c.relpages * 8192) "
```
dimitri, line 858: We already have c.relpages and c.reltuples::bigint in the `listSourceTablesSQL`, I think we should re-use them rather than fetch them again from the source database. Maybe we can use a SQLite query and still populate the `s_table_size` table, which seems like a good idea to simplify our work later. If it's not possible to have SQLite compute `pg_size_pretty()` then we could compute that on the fly when reading from the SQLite catalogs and the `bytes_pretty` column is NULL when the `bytes` column is populated.

Now that we have our own SQLite copy vendored-in, I wonder how difficult it is to enhance it with our own SQL functions, that said. We might want to implement `pg_size_pretty` as a SQLite function. If that proves easy enough, later we might even implement an LSN type in SQLite.

dimitri (commented):
First round of review, I think we're re-fetching information we already have locally and should avoid that round-trip.

src/bin/pgcopydb/schema.c
hanefi: I updated `getTableArray` function to also record the table size estimates for each table. This function was responsible for looping through the result of `listSourceTablesSQL` that is mentioned on your comment.

I did not need to implement `pg_size_pretty` as I could reuse `pretty_print_bytes` function that has a similar purpose.

src/bin/pgcopydb/schema.c
```diff
@@ -4739,6 +4666,24 @@ getTableArray(void *ctx, PGresult *result)
 				parsedOk = false;
 				break;
 			}
+
+			if (context->estimateTableSizes)
+			{
+				uint64_t bytes = table->relpages * 8192;
```
dimitri, line 4672: Nitpicking: let's use a #define'd constant here and some comments too.

src/bin/pgcopydb/schema.c
```diff
@@ -4739,6 +4666,24 @@ getTableArray(void *ctx, PGresult *result)
...
+
+				SourceTableSize sourceTableSize = {
+					.oid = table->oid,
+					.bytes = bytes
+				};
+				pretty_print_bytes(sourceTableSize.bytesPretty, PG_NAMEDATALEN, bytes);
```
dimitri, line 4678: Nitpicking: use `sizeof(sourceTableSize.bytesPretty)` here?

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -542,7 +542,8 @@ copydb_fetch_source_schema(CopyDataSpec *specs, PGSQL *src)
 		!sourceDB->sections[DATA_SECTION_TABLE_DATA].fetched)
 	{
 		/* copydb_fetch_filtered_oids() needs the table size table around */
-		if (!schema_prepare_pgcopydb_table_size(src,
+		if (!specs->estimateTableSizes &&
+			!schema_prepare_pgcopydb_table_size(src,
```
dimitri, line 546: I think it's time to move that logic within the `copydb_prepare_table_specs()` function. What do you think?

src/bin/pgcopydb/schema.c
dimitri: Ah, that's a nice and simple approach to it, cheers.

src/bin/pgcopydb/defaults.h
```diff
@@ -81,6 +82,7 @@
 #define POSTGRES_PING_RETRY_BASE_SLEEP_TIME 5         /* milliseconds */
 
 #define POSTGRES_PORT 5432
+#define POSTGRES_BLOCK_SIZE 8192
```
dimitri, line 85: Searching for the Postgres source code definition of that value, it seems that it is dynamic enough (compile-time?) to warrant actually asking the database for the value. It now seems best to have that computation done in the SQL query we run on the source database, adding a new column with the result of `relpages * current_setting('block_size')::bigint` or something.

dimitri (changes requested):
It seems to me like we should be prepared to handle a different block size, even though it's a compile time option in Postgres.

src/bin/pgcopydb/defaults.h
hanefi: This value is set at PG configure. Allowed values are (in kB): 1,2,4,8,16,32.

```
./configure --with-blocksize=BLOCKSIZE
```

In the worst case, we end up overestimating by a factor of 8, or underestimate by a factor of 4.

Let me update the SQL commands to fetch the current value.

hanefi:
Github is being weird, showing a different set of commits than I expected. I will open a new one.

PR #779, opened 2024-05-15
Implement --skip-db-properties.

### LINKED ISSUE #730: Skip ALTER DATABASE commands
I'm experimenting with using pgcopydb to migrate from Heroku to AWS Aurora Postgres.

When I attempt to clone the DB I'm getting an error:

```
00:52:24.398 12833 INFO   ALTER DATABASE "my_database" SET "temp_tablespaces" TO 'ephemeral';
00:52:24.399 12833 ERROR  [TARGET 52359] ERROR:  permission denied to set parameter "temp_tablespaces"
```

Seemingly because this parameter can't be set in AWS Aurora. Is there anyway to skip the copying of database settings. Ideally I'd just copy the schema (tables, indexes, foreign keys, etc...) and not any of the database settings from my source database?

It seems like I can manually `pg_dump`/`pg_restore` the `pre-data` and `post-data` and then use `pgcopydb copy data` to copy just the data/indexes, but I'd love to use the tool end-to-end if possible.

Thanks for your work on this very helpful tool!

dimitri:
In some cases the source database properties can't be installed on the target using ALTER DATABASE SET command, sometimes because the source hosting provider has added its own properties, sometimes because the target hosting provider has disabled or revoked privileges to edit the property.

For those cases, implement --skip-db-properties.

Fixes #730


src/bin/pgcopydb/cli_common.c
```diff
@@ -825,9 +854,11 @@ cli_copy_db_getopts(int argc, char **argv)
 		exit(EXIT_CODE_BAD_ARGS);
 	}
 
+	const char *optstring =
+		"S:T:D:J:I:b:L:cAPOXj:xBeMlUdyF:F:Q:irRCN:fp:ws:o:tE:Vvdzqh";
```
hanefi, line 858: I do not see `g` character on this line. You appear to add the characters `d` and `y` in this PR.

If you try the short version of the new argument, you get `clone: invalid option -- 'g'`.

PR #780, opened 2024-05-15
Add missing args for skip-db-properties option

hanefi:
On #779 we forgot to add short option -g for skip-db-properties.

See https://github.com/dimitri/pgcopydb/pull/779/files#r1601520982


dimitri (changes requested):
Now we have `d` twice in the optstring, can we clean that up by removing the first one? It should be `-d` for `--debug`.

dimitri (approved):

PR #785, opened 2024-05-21
Skip disable by ctid

VaibhaveS:
In some cases the CTID scan is slow due to the underlying system. For those cases, implement `--skip-ctid-split`.


dimitri (changes requested):
I need to think about the name of the option. I think `--disable-split-by-ctid` makes more sense, even though I see why you're adding another `--skip` feature flag here. I mean, your approach also makes sense. Naming is hard.

I'm surprised we're lacking unit tests coverage. Could you add something in the tests/unit suite?

src/bin/pgcopydb/cli_clone_follow.c
```diff
@@ -47,6 +47,7 @@
 	"  --skip-collations             Skip restoring collations\n" \
 	"  --skip-vacuum                 Skip running VACUUM ANALYZE\n" \
 	"  --skip-db-properties          Skip copying ALTER DATABASE SET properties\n" \
+	"  --skip-ctid-split             Skip spliting tables by ctid\n" \
```
dimitri, line 50: I think `--skip-split-bt-ctid` would be a improvement for the naming, what do you think?

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -773,6 +773,20 @@ copydb_prepare_table_specs_hook(void *ctx, SourceTable *source)
 	if (IS_EMPTY_STRING_BUFFER(source->partKey) &&
 		streq(source->amname, "heap"))
 	{
+		if (specs->skipCtidSplit)
+		{
+			log_info("Table %s is %s large "
+					 "which is larger than --split-tables-larger-than %s, "
+					 "does not have a unique column of type integer, "
+					 "and CTID split is disabled."
```
dimitri, line 781: Missing space after end `.`.

src/bin/pgcopydb/cli_list.c
```diff
@@ -1278,8 +1287,22 @@ cli_list_table_parts(int argc, char **argv)
...
-	if (IS_EMPTY_STRING_BUFFER(table->partKey))
+	if (IS_EMPTY_STRING_BUFFER(table->partKey) &&
+		streq(table->amname, "heap"))
 	{
+		if (copySpecs.skipCtidSplit)
+		{
+			log_info("Table %s is %s large "
+					 "which is larger than --split-tables-larger-than %s, "
+					 "does not have a unique column of type integer, "
+					 "and CTID split is disabled."
```
dimitri, line 1298: Missing space after end `.`.

src/bin/pgcopydb/cli_list.c
```diff
@@ -275,7 +277,7 @@ cli_list_db_getopts(int argc, char **argv)
 		exit(EXIT_CODE_BAD_ARGS);
 	}
 
-	while ((c = getopt_long(argc, argv, "S:D:s:t:F:xPL:fcCyarJRIN:Vdzvqh",
+	while ((c = getopt_long(argc, argv, "S:D:s:t:F:xPL:fcCyarJRIN:Vdzvqhk",
```
dimitri, line 280: Please put the option short letter at the right place in the string (same place as in the long options array).

src/bin/pgcopydb/cli_common.c
```diff
@@ -855,7 +884,7 @@ cli_copy_db_getopts(int argc, char **argv)
 	}
 
 	const char *optstring =
-		"S:T:D:J:I:b:L:cAPOXj:xBeMlUgyF:F:Q:irRCN:fp:ws:o:tE:Vvdzqh";
+		"S:T:D:J:I:b:L:cAPOXj:xBeMlUgyF:F:Q:irRCN:fp:ws:o:tE:Vvdzqhk";
```
dimitri, line 887: Please put the option short letter at the right place in the string (same place as in the long options array).

dimitri (changes requested):
Looks good to me, apart from some nitpicking details.

dimitri (approved):

PR #789, opened 2024-05-24
Allow exclude-index filtering when used with other filters.

arajkumar:
[exclude-index] works only when there is no other filter present.

Solution: Fix index listing query to consider exclude-index filters on all possible scenarios.


dimitri (dismissed):

dimitri:
The style checker is unhappy, can you see about that?

dimitri (approved):

PR #793, opened 2024-05-29
Introduce new options for estimating table sizes

hanefi:
This commit introduces a new option `--estimate-table-sizes`. When set, pgcopydb will use the relpages value in system catalog to estimate the size of each table. This is done by multiplying the number of pages by the page size. PostgreSQL uses a default of 8KB for page sizes, but can be changed in build time.

Optionally, the user can set the environment variable `PGCOPYDB_ESTIMATE_TABLE_SIZES` to a boolean value to enable the option.

If this option is used, we run `vacuumdb --analyze-only` on the source to update the relpages values before calculating the estimates.

In passing, I fixed several things including the following:
- Fix comment references to pg_autoctl
- Remove some unused functions
- Remove reference to non-existing vacuumdb command
- Remove `--cache` and `--drop-cache` references that are no longer used
- Fix typos in vacuum logs and comments
- Remove references to `pgcopydb_table_size` table

Closes #775


docs/ref/pgcopydb_clone.rst
```diff
@@ -729,6 +729,16 @@ PGCOPYDB_SPLIT_TABLES_LARGER_THAN
    When ``--split-tables-larger-than`` is ommitted from the command line,
    then this environment variable is used.
 
+PGCOPYDB_ESTIMATE_TABLE_SIZES
```
dimitri, line 732: Actual documentation coverage for the option is missing, see the option list at https://pgcopydb.readthedocs.io/en/latest/ref/pgcopydb_clone.html#options. In particular, we need to document the `vacuumdb` call that is implemented when using that option, as that's not transparent to the user at all.

src/bin/pgcopydb/schema.c
```diff
@@ -4919,6 +4885,8 @@ parseCurrentSourceTable(PGresult *result, int rowNumber, SourceTable *table)
 	int fnrestorelistname = PQfnumber(result, "format");
 	int fnpartkey = PQfnumber(result, "partkey");
 	int fnattrs = PQfnumber(result, "attributes");
+	int fnbytesestimate = PQfnumber(result, "bytesestimate");
+	int fnbytesestimatepp = PQfnumber(result, "pg_size_pretty");
```
dimitri, line 4889: Nitpicking: could we place these in the SQL query column list order?

dimitri (changes requested):
Minor change needed, and documentation coverage of the new option.

dimitri (approved):

PR #801, opened 2024-06-04
Fix issues in the docs.

rimbi:
Fix issues in the docs:

- Updates the docs and help message with regards to sentinel table as it was moved from source databse to sqlite.
- Fixes minor grammatical issues in the docs.


README.md
```diff
@@ -107,22 +107,22 @@ $ pgcopydb help
...
     transform  Transform changes from the source database into SQL commands
     apply      Apply changes from the source database into the target database
 
   pgcopydb stream sentinel
-    create  Create the sentinel table on the source database
-    drop    Drop the sentinel table on the source database
-    get     Get the sentinel table values on the source database
-  + set     Maintain a sentinel table on the source database
+    create  Create the sentinel table
+    drop    Drop the sentinel table
```
dimitri, line 117: We don't have these two commands anymore, I believe this PR should also remove the entries from this list.

docs/update-help-messages.sh
```diff
@@ -114,10 +114,10 @@ function parse_help_output() {
 
         # Parse subcommands that are followed by a section header. For example,
         # there are the subcommands under pgcopydb strem sentinel section:
-        #     create  Create the sentinel table on the source database
-        #     drop    Drop the sentinel table on the source database
-        #     get     Get the sentinel table values on the source database
-        #   + set     Maintain a sentinel table on the source database
+        #     create  Create the sentinel table
+        #     drop    Drop the sentinel table
```
dimitri, line 118: I'm confused as of why we still have these.

dimitri (changes requested):
LGTM, thanks!

There is only one catch with the `pgcopydb sentinel create|drop` CLIs that I believe do not exist anymore, right?

README.md
rimbi: You are right, fixed. 👍

docs/update-help-messages.sh
rimbi: You are right, fixed. 👍

dimitri (approved):

Issue #799, opened 2024-06-04
Implement an option to limit maximum number of table splits


hanefi:
When same table concurrency is enabled, a user may want to limit the maximum number of splits.

We can consider adding `--split-count-max-limit` that provides an upper limit to the number of splits.

PR #803, opened 2024-06-04
Add support for limiting same table concurrency

hanefi:
This commit adds a new option --split-count-max-limit to limit the number of jobs that can be run concurrently for the same table. This option is useful when the number of tables is large and the system is running out of resources.

TODO:

- [x] Add a proper test. Possibly a unit test
- [x] Add a new env var for the new option
- [x] Update docs

Fixes: #799


hanefi:
[like]  Hanefi Onaldi reacted to your message:
________________________________

src/bin/pgcopydb/cli_common.c
```diff
@@ -1010,6 +1034,20 @@ cli_copy_db_getopts(int argc, char **argv)
 				break;
 			}
 
+			case 'u':
+			{
+				if (!stringToInt(optarg, &options.splitMaxParts) ||
+					options.splitMaxParts < 1 ||
+					options.splitMaxParts > 128)
```
dimitri, line 1041: Why is the maximum number of partitions 128? Sounds reasonable to me, but I think it warrants a comment in the code and also documentation coverage.

src/bin/pgcopydb/cli_list.c
```diff
@@ -221,6 +222,29 @@ cli_list_getenv(ListDBOptions *options)
 		++errors;
 	}
 
+	/* when --split-max-parts has not been used, check PGCOPYDB_SPLIT_MAX_PARTS */
```
dimitri, line 225: Is there a way to reuse `cli_copydb_getenv()` from here?

src/bin/pgcopydb/schema.c
```diff
@@ -3230,14 +3230,16 @@ schema_list_pg_depend(PGSQL *pgsql,
...
+ * our parameters: table size, --split-tables-larger-than, and
+ * --split-max-parts.
  */
 bool
 schema_list_partitions(PGSQL *pgsql,
 					   DatabaseCatalog *catalog,
 					   SourceTable *table,
-					   uint64_t partSize)
+					   uint64_t partSize,
+					   int partCountMaxLimit)
```
dimitri, line 3242: Can you rename this parameter also?

dimitri (changes requested):
Mostly nitpicking now.

src/bin/pgcopydb/cli_common.c
hanefi: Now that you mention it, it does not make sense to set an upper bound at all. Maybe I should just remove the upper bound.

src/bin/pgcopydb/cli_list.c
hanefi: `cli_copydb_getenv()` reads environment variables and fills in the `CopyDBOptions` values.

`cli_list_getenv()` does a similar thing but it fills in `ListDBOptions` instead.

We can consider creating a helper function that is called by both of these functions and fill in the values for the relevant structure. I can not think of a clean way to implement this as we do not have generic functions that enables writing reusable code that adapts to different data types.

We can consider doing this in a follow-up refactor PR.

dimitri: Good point, forgot about that. Thanks for the clear reminder.

tests/pagila/copydb.sh
```diff
@@ -61,6 +61,7 @@ pgcopydb clone --notice \
          --skip-ext-comments \
          --skip-db-properties \
          --estimate-table-sizes \
+         --split-max-parts=5 \
```
dimitri, line 64: Could we use the environment variable in the docker-compose file for this one? that would allow testing that parts of the code.

dimitri (dismissed):
Please consider one last change before merge.

tests/pagila/copydb.sh
hanefi: I removed this line and added environment variable in docker-compose. I verified that we still limit the splits to 5 on my local test environment.

dimitri (approved):

Issue #760, opened 2024-04-26
Migration failing with deadlock happening due to circular fkey dependencies at constraint creation time


sarvkuma:
Logs from migration

```
2024-04-24 10:50:20.974 80 ERROR string_utils.c:669 DETAIL: Process 14145 waits for ShareRowExclusiveLock on relation 130381996 of database 130381954; blocked by process 14151.
2024-04-24 10:50:20.974 80 ERROR string_utils.c:669 Process 14151 waits for ShareRowExclusiveLock on relation 130382053 of database 130381954; blocked by process 14145.
--
2024-04-24 10:50:21.067 80 ERROR string_utils.c:669 DETAIL: Process 14148 waits for ShareRowExclusiveLock on relation 130382044 of database 130381954; blocked by process 14166.
2024-04-24 10:50:21.067 80 ERROR string_utils.c:669 Process 14166 waits for ShareRowExclusiveLock on relation 130382053 of database 130381954; blocked by process 14167.
2024-04-24 10:50:21.067 80 ERROR string_utils.c:669 Process 14167 waits for ShareRowExclusiveLock on relation 130382044 of database 130381954; blocked by process 14148.
```

Each Process details with statements 

**Process 14145**:
```
ALTER TABLE ONLY public.booking_reservation_template
    ADD CONSTRAINT fk_rlf_booking_template FOREIGN KEY (relief_booking_id) 
    REFERENCES public.booking(id) ON UPDATE CASCADE ON DELETE CASCADE;
```

**Process 14151**: 
```
ALTER TABLE ONLY public.booking
    ADD CONSTRAINT fk_booking_template_rlf FOREIGN KEY (reservation_template_id) 
    REFERENCES public.booking_reservation_template(id) ON UPDATE CASCADE ON DELETE CASCADE;
```

**Process 14148:** 
```
ALTER TABLE ONLY public.long_term_vehicle
    ADD CONSTRAINT long_term_vehicle_booking_id_fkey FOREIGN KEY (booking_id) 
    REFERENCES public.booking_prelease(id) ON DELETE CASCADE;
```

**Process 14166:** 
```
ALTER TABLE ONLY public.booking_prelease
    ADD CONSTRAINT fk_booking_template_prl FOREIGN KEY (reservation_template_id) 
    REFERENCES public.booking_reservation_template(id) ON UPDATE CASCADE ON DELETE CASCADE;
```

**Process 14167**: 
```
[...]

PR #804, opened 2024-06-05
Use a single dump file to give more dependency information to pg_restore.

rimbi:
See #760


src/bin/pgcopydb/copydb.c
```diff
@@ -423,18 +423,15 @@ copydb_prepare_dump_paths(CopyFilePaths *cfPaths, DumpPaths *dumpPaths)
 	sformat(dumpPaths->extnspFilename, MAXPGPATH, "%s/%s",
 			cfPaths->schemadir, "extnamespaces.dump");
 
-	sformat(dumpPaths->preFilename, MAXPGPATH, "%s/%s",
-			cfPaths->schemadir, "pre.dump");
+	sformat(dumpPaths->dumpFilename, MAXPGPATH, "%s/%s",
+			cfPaths->schemadir, "pre_post.dump");
```
dimitri, line 427: The actual name from pg_dump and pg_restore documentations is “schema-only”. Let's use `schema.dump`.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -86,6 +87,48 @@ copydb_objectid_has_been_processed_already(CopyDataSpec *specs,
...
+getSectionOptions(PostgresDumpSection section, char *sections[3])
+{
+	switch (section)
+	{
+		case PG_DUMP_SECTION_SCHEMA | PG_DUMP_SECTION_ALL:
+		{
+			sections[0] = "pre-data";
+			sections[1] = "post-data";
+			break;
+		}
```
dimitri, line 108: I do not think that's needed. The pg_dump documented way to do that is to use `--schema-only`, I believe.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -110,46 +153,24 @@ copydb_dump_source_schema(CopyDataSpec *specs,
...
+			log_info("Skipping pg_dump --section=pre-data, "
 					 "as \"%s\" already exists",
-					 specs->dumpPaths.postFilename);
+					 specs->dumpPaths.dumpFilename);
 		}
 		else if (!pg_dump_db(&(specs->pgPaths),
 							 &(specs->connStrings),
 							 snapshot,
-							 "post-data",
+							 sections,
```
dimitri, line 170: For pg_dump, it's going to be `--schema-only` always, so I believe this could be a static information in the `pg_dump_db` implementation?

src/bin/pgcopydb/pgcmd.c
```diff
@@ -358,13 +358,13 @@ typedef struct DumpExtensionNamespaceContext
...
 /*
- * Call pg_dump and get the given section of the dump into the target file.
+ * Call pg_dump and get the given section(s) of the dump into the target file.
  */
 bool
 pg_dump_db(PostgresPaths *pgPaths,
 		   ConnStrings *connStrings,
 		   const char *snapshot,
-		   const char *section,
+		   char *sections[],
```
dimitri, line 367: See other comments, I believe no argument is needed here.

src/bin/pgcopydb/pgcmd.c
```diff
@@ -402,8 +402,11 @@ pg_dump_db(PostgresPaths *pgPaths,
 		args[argsIndex++] = (char *) snapshot;
 	}
 
-	args[argsIndex++] = "--section";
-	args[argsIndex++] = (char *) section;
+	for (int i = 0; sections[i] != NULL; i++)
+	{
+		args[argsIndex++] = "--section";
+		args[argsIndex++] = sections[i];
+	}
```
dimitri, line 409: Just have the following, I suppose that's all we need really?

``` c
    args[argsIndex++] = "--schema-only";
```

dimitri (changes requested):
I expected to see `pg_dump --schema-only` and then `pg_restore --section=pre-data` and a variant with `pg_restore --section=post-data`. In particular, I am not seeing the change where pg_restore works on a single section at a time?

rimbi:
> I expected to see `pg_dump --schema-only` and then `pg_restore --section=pre-data` and a variant with `pg_restore --section=post-data`. In particular, I am not seeing the change where pg_restore works on a single section at a time?

If you search for `strlcpy` you can see the parts related to `pg_restore`. Meanwhile, I will make the changes to use `--schema-only`. Btw, do you know the difference between using `--schema-only` and `pre-data + post-data`? The documentation does not clarify it but just says `they are similar but not identical`.

dimitri:
> Meanwhile, I will make the changes to use `--schema-only`. Btw, do you know the difference between using `--schema-only` and `pre-data + post-data`? The documentation does not clarify it but just says `they are similar but not identical`.

I don't know the specifics. I see that in your testing, using the pre+post variant passes our test suite, and using --schema-only fails. Did you see why? I mean I'm good with either one, it would be better to know how the choice was made if we want to be able to maintain it. Also will need comments and maybe documentation coverage.

src/bin/pgcopydb/pgcmd.h
```diff
@@ -191,6 +191,7 @@ typedef struct RestoreOptions
 	bool noACL;
 	bool noTableSpaces;
 	int jobs;
+	char section[BUFSIZE];
```
dimitri, line 194: Style issue: either have an enum datatype for internal option management, or just add a `const char *section` argument to `pg_restore_db`.

dimitri (commented):
What happens to the command lines `pgcopydb dump pre-data` and `pgcopydb dump post-data`? I expected to see them removed.

src/bin/pgcopydb/pgcmd.h
rimbi: Ok. Can you point me to an example in the sources, please?

rimbi:
> > Meanwhile, I will make the changes to use `--schema-only`. Btw, do you know the difference between using `--schema-only` and `pre-data + post-data`? The documentation does not clarify it but just says `they are similar but not identical`.
> 
> I don't know the specifics. I see that in your testing, using the pre+post variant passes our test suite, and using --schema-only fails. Did you see why? I mean I'm good with either one, it would be better to know how the choice was made if we want to be able to maintain it. Also will need comments and maybe documentation coverage.

I am still not sure what the behavior differs but I see the following difference in the logs:
image
with the left side being the failed one. Apparently, `pg_restore --section=pre-data` after `pg_dump --schema-only` does not restore large object metadata.

src/bin/pgcopydb/pgcmd.h
dimitri: The structure `RestoreOptions` itself is a good example, along with code such as the following. As you can see we do not store the string "--no-owner" in the structure, we store the C boolean information that the option noOwner has been selected. This is expanded to the command line argument, as a string, as needed, when preparing to command line.

``` c
	if (options.noOwner)
	{
		args[argsIndex++] = "--no-owner";
	}
```

rimbi: gotcha! 👍

rimbi:
> > Apparently, `pg_restore --section=pre-data` after `pg_dump --schema-only` does not restore large object metadata.
> 
> That would make sense, kind of. Could you run the pg_dump locally with the two sets of options, and then diff the SQL output? or maybe diff the pg_restore --list output?
> 
> If we confirm that the diff is the LO metadata, then we can add a comment to that effect in our source code.

here is the diff:
image

src/bin/pgcopydb/pgcmd.h
rimbi: done

src/bin/pgcopydb/pgcmd.c
```diff
@@ -402,8 +402,8 @@ pg_dump_db(PostgresPaths *pgPaths,
 		args[argsIndex++] = (char *) snapshot;
 	}
 
-	args[argsIndex++] = "--section";
-	args[argsIndex++] = (char *) section;
+	args[argsIndex++] = "--section=pre-data";
+	args[argsIndex++] = "--section=post-data";
```
dimitri, line 406: Please add a comment here covering why we want both the sections in the same file (pg_restore dependency tracking) and why we prefer using these two options rather than `--schema-only`: we want to retain the BLOB metadata in the dump, and this is included in --pre-data, but not in --schema-only.

src/bin/pgcopydb/pgcmd.h
```diff
@@ -183,6 +183,41 @@ typedef struct ArchiveContentArray
...
 
 
+/*
+ * Section options to use with pg_dump and pg_restore.
+ */
+typedef enum
+{
+	PRE_DATA = 0,
+	POST_DATA,
+} SectionOption;
```
dimitri, line 193: Please add a prefix such as PG_RESTORE_SECTION_ to these constants. C lacks namespaces :/

src/bin/pgcopydb/pgcmd.h
```diff
@@ -183,6 +183,41 @@ typedef struct ArchiveContentArray
...
+ * Convert SectionOption to string.
+ */
+static inline const char *
+sectionOptionToString(SectionOption section)
+{
+	switch (section)
+	{
+		case PRE_DATA:
+		{
+			return "--section=pre-data";
```
dimitri, line 205: I think I would just return "pre-data" here, and use two entries in the command line argument array:

``` c
    args[index++] = "--section";
    args[index++] = (char *) sectionOptionToString(section);
```

src/bin/pgcopydb/pgcmd.h
dimitri: Or better yet, can you just re-use PG_DUMP_SECTION_PRE_DATA etc maybe?

src/bin/pgcopydb/dump_restore.c
```diff
@@ -574,10 +545,11 @@ copydb_target_finalize_schema(CopyDataSpec *specs)
 		return false;
 	}
 
+	specs->restoreOptions.section = POST_DATA;
```
dimitri, line 548: I still wonder what's the advantage of that approach compared to adding a section argument to the function call to `pg_restore_db`?

dimitri (changes requested):
More changes are needed, and please remove the `pgcopydb dump pre-data` and `pgcopydb dump post-data` commands both in the code and the docs.

src/bin/pgcopydb/dump_restore.c
rimbi: My point of view is that it is in the end one of the restore options and thus should be part of this structure.

src/bin/pgcopydb/pgcmd.h
rimbi: makes sense 👍

src/bin/pgcopydb/pgcmd.h
rimbi: makes sense :thumbsup:

src/bin/pgcopydb/dump_restore.c
dimitri: That's a good point, sure. The way we edit the RestoreOption just before the call, and only there, still makes me wonder about it. But yeah, fine.

src/bin/pgcopydb/pgcmd.c
rimbi: done! :thumbsup:

src/bin/pgcopydb/pgcmd.h
```diff
@@ -182,32 +182,33 @@ typedef struct ArchiveContentArray
...
 		}
 
-		case POST_DATA:
+		case PG_DUMP_SECTION_POST_DATA:
 		{
-			return "--section=post-data";
+			return "post-data";
 		}
 
 		default:
```
dimitri, line 214: We need to implement “defensive programming” here. List the other cases from the enum and error out with a specific error message like “BUG: postgresDumpSectionToString called with unexpected section %d”. Then the default catch-all can have a proper error message such as “unknown pg_dump section %d“.

dimitri (changes requested):

src/bin/pgcopydb/pgcmd.h
rimbi: done!

rimbi:
> What happens to the command lines `pgcopydb dump pre-data` and `pgcopydb dump post-data`? I expected to see them removed.

They have just been removed. The docs have been updated too. 👍

docs/ref/pgcopydb_restore.rst
```diff
@@ -290,8 +290,8 @@ First, using ``pgcopydb restore schema``
    09:54:37 20401 INFO  Found a stale pidfile at "/tmp/target//pgcopydb.pid"
    09:54:37 20401 WARN  Removing the stale pid file "/tmp/target//pgcopydb.pid"
    09:54:37 20401 INFO  Using pg_restore for Postgres "12.9" at "/Applications/Postgres.app/Contents/Versions/12/bin/pg_restore"
-   09:54:37 20401 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54314 dbname=demo' --clean --if-exists /tmp/target//schema/pre.dump
-   09:54:38 20401 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54314 dbname=demo' --clean --if-exists --use-list /tmp/target//schema/post.list /tmp/target//schema/post.dump
+   09:54:37 20401 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54314 dbname=demo' --clean --if-exists /tmp/target//schema/schema.dump
+   09:54:38 20401 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54314 dbname=demo' --clean --if-exists --use-list /tmp/target//schema/post.list /tmp/target//schema/schema.dump
```
dimitri, line 294: I think this is missing the `--section pre-data` and `--section post-data` command line options. Feel free to reproduce these outputs from scratch on your system.

src/bin/pgcopydb/pgcmd.h
```diff
@@ -182,6 +183,56 @@ typedef struct ArchiveContentArray
...
+		{
+			return "post-data";
+		}
+
+		case PG_DUMP_SECTION_DATA:
+		{
+			return "data";
+		}
+
+		case PG_DUMP_SECTION_SCHEMA | PG_DUMP_SECTION_ROLES | PG_DUMP_SECTION_ALL:
```
dimitri, line 220: That's not the typical C spelling... this needs to look like other switch statements in the code base:

``` c
    switch (value)
    {
        case 1:
        case 2:
        {
            ...
        }
    }
```

src/bin/pgcopydb/pgcmd.c
```diff
@@ -897,6 +903,13 @@ pg_restore_db(PostgresPaths *pgPaths,
 	args[argsIndex++] = "--dbname";
 	args[argsIndex++] = (char *) connStrings->safeTargetPGURI.pguri;
 
+	const char *sectionOption = postgresDumpSectionToString(options.section);
+	if (sectionOption != NULL)
+	{
+		args[argsIndex++] = "--section";
+		args[argsIndex++] = (char *) sectionOption;
+	}
```
dimitri, line 911: That needs an else branch that just returns false, right?

dimitri (changes requested):
Almost there!

src/bin/pgcopydb/pgcmd.h
rimbi: ah this is Rust syntax, good catch! 😄

docs/ref/pgcopydb_restore.rst
rimbi: That makes sense. Alternatively, we can remove these log messages from the docs completely. What do you think?

README.md
```diff
@@ -145,18 +143,17 @@ an overall summary that looks like the following:
...
+18:26:36 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54311 dbname=plop' /tmp/pgcopydb/schema/schema.dump
 18:26:36 77615 INFO  STEP 3: copy data from source to target in sub-processes
 18:26:36 77615 INFO  STEP 4: create indexes and constraints in parallel
 18:26:36 77615 INFO  STEP 5: vacuum analyze each table
 18:26:36 77615 INFO  Listing ordinary tables in "port=54311 host=localhost dbname=pgloader"
 18:26:36 77615 INFO  Fetched information for 56 tables
 ...
 18:26:37 77615 INFO  STEP 6: restore the post-data section to the target database
-18:26:37 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54311 dbname=plop' --use-list /tmp/pgcopydb/schema/post.list /tmp/pgcopydb/schema/post.dump
+18:26:37 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54311 dbname=plop' --use-list /tmp/pgcopydb/schema/post.list /tmp/pgcopydb/schema/schema.dump
```
dimitri, line 156: It's missing `--section=post-data`, right?

docs/ref/pgcopydb_clone.rst
```diff
@@ -919,10 +919,9 @@ Examples
...
-   14:49:01 24 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-00000022-1 --section post-data --file /tmp/pgcopydb/schema/post.dump 'postgres://pagila@source/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
+   14:49:01 24 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-00000022-1 --section pre-data --section post-data --file /tmp/pgcopydb/schema/schema.dump 'postgres://pagila@source/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
    14:49:02 24 INFO   STEP 3: restore the pre-data section to the target database
-   14:49:02 24 INFO    /usr/bin/pg_restore --dbname 'postgres://pagila@target/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --single-transaction --use-list /tmp/pgcopydb/schema/pre-filtered.list /tmp/pgcopydb/schema/pre.dump
+   14:49:02 24 INFO    /usr/bin/pg_restore --dbname 'postgres://pagila@target/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --single-transaction --use-list /tmp/pgcopydb/schema/pre-filtered.list /tmp/pgcopydb/schema/schema.dump
```
dimitri, line 924: I think it's missing `--section=pre-data`

docs/ref/pgcopydb_clone.rst
```diff
@@ -932,7 +931,7 @@ Examples
    14:49:02 52 INFO   Reset sequences values on the target database
    14:49:02 51 INFO   Added 0 large objects to the queue
    14:49:04 24 INFO   STEP 10: restore the post-data section to the target database
-   14:49:04 24 INFO    /usr/bin/pg_restore --dbname 'postgres://pagila@target/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --single-transaction --use-list /tmp/pgcopydb/schema/post-filtered.list /tmp/pgcopydb/schema/post.dump
+   14:49:04 24 INFO    /usr/bin/pg_restore --dbname 'postgres://pagila@target/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --single-transaction --use-list /tmp/pgcopydb/schema/post-filtered.list /tmp/pgcopydb/schema/schema.dump
```
dimitri, line 934: `--section=post-data`

docs/ref/pgcopydb_copy.rst
```diff
@@ -473,8 +473,7 @@ Now, first dump the schema:
    14:28:50 22 INFO   Dumping database into directory "/tmp/pgcopydb"
    14:28:50 22 INFO   Using pg_dump for Postgres "16.1" at "/usr/bin/pg_dumpall"
    14:28:50 22 INFO   Exported snapshot "00000003-00000022-1" from the source database
-   14:28:50 22 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-00000022-1 --section pre-data --file /tmp/pgcopydb/schema/pre.dump 'postgres://pagila@source/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
-   14:28:51 22 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-00000022-1 --section post-data --file /tmp/pgcopydb/schema/post.dump 'postgres://pagila@source/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
+   14:28:50 22 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-00000022-1 --section pre-data --section post-data --file /tmp/pgcopydb/schema/schema.dump 'postgres://pagila@source/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
 
 Now restore the pre-data schema on the target database, cleaning up the
 already existing objects if any, which allows running this test scenario
```
dimitri, line 479: Did you check that this still works? I suppose `make tests/pagila-multi-steps` is as good a test as any. Again, it might be worth replacing these output samples entirely with the new commands.

docs/ref/pgcopydb_restore.rst
dimitri: As a user I like being able to make sense of what the software is doing by reading the logs, and it's even better if I didn't need to setup anything to be able to read through logs.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -635,20 +607,20 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
 	 * constraints and indexes that we have already built concurrently in the
 	 * previous step, so we want to filter those out.
 	 *
 	 * Here's how to filter out some objects with pg_restore:
 	 *
-	 *   1. pg_restore -f post.list --list post.dump
+	 *   1. pg_restore -f post.list --list schema.dump
 	 *   2. edit post.list to comment out lines and save as filtered.list
-	 *   3. pg_restore --use-list filtered.list post.dump
+	 *   3. pg_restore --use-list filtered.list schema.dump
```
dimitri, line 623: This also is missing `--section=post-data`

src/bin/pgcopydb/pgcmd.c
dimitri: That sounds like a silent error to me. We only return NULL when we detected a bug of some sorts... (calling the function with an unexpected value). I do not think it makes sense to act as if we didn't notice the problem.

docs/ref/pgcopydb_restore.rst
rimbi: Because it has a considerable maintenance cost, I am not a fan of it. But I will update it as it already looks out-of-date.

src/bin/pgcopydb/dump_restore.c
rimbi: good catch!

README.md
rimbi: replaced all the logs

README.md
```diff
@@ -145,18 +143,17 @@ an overall summary that looks like the following:
...
 18:26:35 77615 INFO  [TARGET] Copying database into "port=54311 dbname=plop"
 18:26:35 77615 INFO  STEP 1: dump the source database schema (pre/post data)
-18:26:35 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_dump -Fc --section pre-data --file /tmp/pgcopydb/schema/pre.dump 'port=54311 host=localhost dbname=pgloader'
-18:26:35 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_dump -Fc --section post-data --file /tmp/pgcopydb/schema/post.dump 'port=54311 host=localhost dbname=pgloader'
+18:26:35 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_dump -Fc --section pre-data --section post-data --file /tmp/pgcopydb/schema/schema.dump 'port=54311 host=localhost dbname=pgloader'
 18:26:36 77615 INFO  STEP 2: restore the pre-data section to the target database
-18:26:36 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54311 dbname=plop' /tmp/pgcopydb/schema/pre.dump
+18:26:36 77615 INFO   /Applications/Postgres.app/Contents/Versions/12/bin/pg_restore --dbname 'port=54311 dbname=plop' /tmp/pgcopydb/schema/schema.dump
```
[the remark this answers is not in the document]
rimbi: replaced all the logs

docs/ref/pgcopydb_clone.rst
rimbi: replaced all the logs

docs/ref/pgcopydb_copy.rst
rimbi: It seems to work. On the other hand, when I follow the commands on this page as follows:
```bash
pgcopydb dump schema
PGCOPYDB_DROP_IF_EXISTS=on pgcopydb restore pre-data --no-owner --resume --not-consistent
pgcopydb copy table-data --resume --not-consistent
pgcopydb copy indexes --resume --not-consistent
```
the last command (indexes) fails on both this and main branch at firs run as follows:

```bash
09:12:04.390 50533 INFO   Running pgcopydb version 0.15.74.gc74047a from "/home/cem/scratch/pgcopydb/src/bin/pgcopydb/pgcopydb"
09:12:04.392 50533 INFO   [SOURCE] Copying database from "postgres://cem@127.0.0.1:5434/testdb1?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
09:12:04.392 50533 INFO   [TARGET] Copying database into "postgres://cem@127.0.0.1:5435/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
09:12:04.422 50533 INFO   Using work dir "/tmp/pgcopydb"
09:12:04.424 50533 INFO   Re-using catalog caches
09:12:04.424 50533 INFO   Creating 1 indexes in the target database using 4 processes
09:12:04.436 50542 INFO   STEP 6: starting 4 CREATE INDEX processes
09:12:04.437 50542 INFO   STEP 7: constraints are built by the CREATE INDEX processes
09:12:04.478 50543 INFO   Found 1 indexes on target database for table public.table1
09:12:04.489 50543 ERROR  Failed to send a message to (null) queue (0) with type 1: Invalid argument
09:12:04.490 50543 ERROR  Failed to queue VACUUM ANALYZE public.table1 [16386]
09:12:04.490 50543 ERROR  Failed to create index with oid 16390, see above for details
09:12:04.493 50543 ERROR  CREATE INDEX worker 50543 encountered 1 errors, see above for details
09:12:04.538 50542 ERROR  Sub-process 50543 exited with code 12
09:12:04.539 50542 ERROR  Some INDEX worker process(es) have exited with error, see above for details
09:12:04.539 50542 ERROR  Failed to create indexes, see above for details
09:12:04.625 50533 ERROR  Sub-process 50542 exited with code 12
09:12:04.626 50533 ERROR  Some sub-processes have exited with error status, see above for details
```
but the second run succeeds as follows:
```bash
09:36:55.946 51658 INFO   Running pgcopydb version 0.15.74.gc74047a from "/home/cem/scratch/pgcopydb/src/bin/pgcopydb/pgcopydb"
09:36:55.951 51658 INFO   [SOURCE] Copying database from "postgres://cem@127.0.0.1:5434/testdb1?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
09:36:55.951 51658 INFO   [TARGET] Copying database into "postgres://cem@127.0.0.1:5435/testdb2?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
09:36:56.026 51658 INFO   Using work dir "/tmp/pgcopydb"
09:36:56.031 51658 INFO   Re-using catalog caches
09:36:56.031 51658 INFO   Creating 1 indexes in the target database using 4 processes
09:36:56.044 51667 INFO   STEP 6: starting 4 CREATE INDEX processes
09:36:56.045 51667 INFO   STEP 7: constraints are built by the CREATE INDEX processes
```
What is your take on it? Is it expected? Are the commands valid or invalid?

README.md
```diff
@@ -140,38 +140,53 @@ the table COPY, cumulative timing for the CREATE INDEX commands), and then
...
-                               Finalize Schema       target        29ms             1
- ---------------------------------------------   ----------  ----------  ------------
-                     Total Wall Clock Duration         both       2s639         8 + 2
- ---------------------------------------------   ----------  ----------  ------------
+07:54:31.646 40543 INFO   [SOURCE] Copying database from "postgres://user@source.host.dev/dbname?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
+07:54:31.646 40543 INFO   [TARGET] Copying database into "postgres://role@target.host.dev/dbname?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60"
+07:54:31.711 40543 INFO   Using work dir "/tmp/pgcopydb"
+07:54:31.719 40543 INFO   Exported snapshot "00000003-000000D1-1" from the source database
+07:54:31.722 40554 INFO   STEP 1: fetch source database tables, indexes, and sequences
+07:54:32.046 40554 INFO   Fetched information for 1 tables (including 0 tables split in 0 partitions total), with an estimated total of 18 quintillion tuples and 8192 B on-disk
```
dimitri, line 148: That reads very strange, 18 quitillion tuples on 8kB. Can you copy/paste the pgbench example, or maybe use the pagila example? (it would be fair to use the output of `make tests/pagila` here).

docs/ref/pgcopydb_clone.rst
```diff
@@ -903,81 +903,54 @@ Examples
...
+   08:13:14.448 42904 INFO    /usr/bin/pg_dump -Fc --snapshot 00000003-000000EB-1 --section=pre-data --section=post-data --file /tmp/pgcopydb/schema/schema.dump 'postgres://pagila:0wn3d@source/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
+   08:13:14.513 42904 INFO   STEP 3: restore the pre-data section to the target database
+   08:13:14.524 42904 INFO    /usr/bin/pg_restore --dbname 'postgres://pagila:0wn3d@target/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --section pre-data --jobs 2 --use-list /tmp/pgcopydb/schema/pre-filtered.list /tmp/pgcopydb/schema/schema.dump
+   08:13:14.608 42919 INFO   STEP 4: starting 8 table-data COPY processes
+   08:13:14.678 42921 INFO   STEP 8: starting 8 VACUUM processes
+   08:13:14.678 42904 INFO   Skipping large objects: none found.
+   08:13:14.693 42920 INFO   STEP 6: starting 2 CREATE INDEX processes
+   08:13:14.693 42920 INFO   STEP 7: constraints are built by the CREATE INDEX processes
+   2024-06-10 08:13:14.695 +03 [42955] LOG:  unexpected EOF on client connection with an open transaction
```
dimitri, line 925: That looks like you're using debug level Postgres settings that pushes server logs as notifications on the client system, and I would rather we do not have that in our example here. That's just a distraction for the docs.

Also, if the server side log is due to pgcopydb operations, we will need to fix that in another PR.

docs/ref/pgcopydb_copy.rst
dimitri: Commands look valid to me. So that's another bug to report and fix.

docs/ref/pgcopydb_restore.rst
dimitri: We can work separately on maintenance cost. There is no substitute to having good docs. That's just a lot of work, and very useful, and we are the ones to make it happen.

src/bin/pgcopydb/pgcmd.c
rimbi: what should be the behavior then here when the value is NULL?

dimitri: Error message and failure of the whole process. NULL here means we know we have a bug, we don't know the section that was asked to be taken care of.

docs/ref/pgcopydb_copy.rst
rimbi: Opened a new issue #813

README.md
rimbi: I don't see it now. I will update this and when we encounter this case once more we can open an issue for it.

docs/ref/pgcopydb_clone.rst
rimbi: Removed.

src/bin/pgcopydb/pgcmd.c
rimbi: I have introduced another enum to represent the restore options.

dimitri (approved):

dimitri:
Fixes #760

Issue #800, opened 2024-06-04
Skip individual ANALYZE commands when estimating table sizes and splitting by CTID


hanefi:
We introduced estimating table sizes recently in #793. For this feature, we ran ANALYZE for all tables before the migration to update the number of pages for each relation.

If we split some tables by CTID, we run ANALYZE commands one more time, and this is not really needed.

PR #805, opened 2024-06-06
Remove redundant ANALYZE queries

hanefi:
If the user enabled the --estimate-table-sizes option, we should not run ANALYZE queries on the tables that will be split by ctid, as we already ran vacuumdb --analyze-only before.

Sample output from pagila:
```
12:02:38.720 46 INFO   Running vacuumdb --analyze-only on source database before calculating table size estimates
12:02:38.720 46 INFO    /usr/bin/vacuumdb --analyze-only --jobs 4 --dbname 'postgres://pagila@source/pagila?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60'
12:02:38.955 46 INFO   vacuumdb: vacuuming database "pagila"

12:02:39.318 46 INFO   Table public.film_actor is 240 kB large which is larger than --split-tables-larger-than 200 kB, and does not have a unique column of type integer: splitting by CTID
12:02:39.318 46 NOTICE Skipping running ANALYZE on table public.film_actor as we are estimating table sizes, and we already ran vacuumdb --analyze-only
```

Fixes: #800

### DIFF
--- a/src/bin/pgcopydb/copydb_schema.c
+++ b/src/bin/pgcopydb/copydb_schema.c
@@ -830,17 +830,25 @@ copydb_prepare_table_specs_hook(void *ctx, SourceTable *source)
 		 * Make sure we have proper statistics (relpages) about the table
 		 * before compute the CTID ranges for the concurrent table scans.
 		 */
-		char sql[BUFSIZE] = { 0 };
-
-		sformat(sql, sizeof(sql), "ANALYZE %s", source->qname);
-
-		log_notice("%s", sql);
-
-		if (!pgsql_execute(context->pgsql, sql))
+		if (specs->estimateTableSizes)
 		{
-			log_error("Failed to refresh table %s statistics",
+			log_debug("Skipping running ANALYZE on table %s for CTID split",
 					  source->qname);
-			return false;
+		}
+		else
+		{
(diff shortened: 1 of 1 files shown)


src/bin/pgcopydb/copydb_schema.c
```diff
@@ -827,17 +827,27 @@ copydb_prepare_table_specs_hook(void *ctx, SourceTable *source)
 		 * Make sure we have proper statistics (relpages) about the table
 		 * before compute the CTID ranges for the concurrent table scans.
 		 */
-		char sql[BUFSIZE] = { 0 };
+		if (specs->estimateTableSizes)
+		{
+			log_notice("Skipping running ANALYZE on table %s as we are "
```
hanefi, line 832: This notice message may contain too much information. Maybe I should remove references to table size estimates from the message.

I will not change this until I hear what Dimitri thinks about it.

dimitri: I believe it could be a very simple and quick DEBUG message. I'm not sure it adds anything that the end-user should consider. While it's important to show every SQL query that is send, specifically ANALYZE which can have a non-trivial impact to production, here we're just skipping a query, no need to tell much about it.

dimitri (approved):

PR #807, opened 2024-06-06
Refactor environment variable parsing logic

hanefi:
We had a repetitive pattern in the code where we checked if an environment variable was set, and if it was, we parsed it and stored it in a target variable.

This patch introduces a new function `get_env_using_parser` that takes all the necesary information needed to parse the environment variable in a struct, and does the parsing. This way, we can avoid repeating the same code over and over again.

In passing, I also refactored our logic to set defaults before reading environment variables, so that we can avoid repeating the same default values in multiple places.

Relevant discussion: https://github.com/dimitri/pgcopydb/pull/803#discussion_r1629197581


src/bin/pgcopydb/cli_common.c
```diff
@@ -228,92 +228,33 @@ cli_copydb_getenv(CopyDBOptions *options)
...
-		}
-		else
-		{
-			/* errors have already been logged */
-			++errors;
-		}
-	}
+	EnvParser parsers[] = {
+		{ PGCOPYDB_TABLE_JOBS, ENV_TYPE_INT, &(options->tableJobs), 1, 128 },
+		{ PGCOPYDB_INDEX_JOBS, ENV_TYPE_INT, &(options->indexJobs), 1, 128 },
```
dimitri, line 234: How are you going to share that parts at two different places? I would host the target value within the structure, like we do in `parseSingleValueResult()` and `SingleValueResultContext`, and then have the caller code dispatch the parsed value at the right place. 

That way this array can be defined only once, and be the authoritative list of supported environment variables.

dimitri (changes requested):
I suggest that you also add the next refactoring step in that same PR, because of the following comments.

hanefi:
In the test suite I saw a failure that indicate I failed to get the character array sizes correctly when parsing environment variables that store strings.

Excerpt from https://github.com/dimitri/pgcopydb/actions/runs/9403866395/job/25901415654?pr=807 :

```
+ pgcopydb dump schema --snapshot 00000003-00000011-1
2024-06-06 15:44:08.649 21 INFO   main.c:136                Running pgcopydb version 0.16.16.g17cb197 from "/usr/local/bin/pgcopydb"
2024-06-06 15:44:08.650 21 ERROR  env_utils.c:110           Failed to copy value stored in PGCOPYDB_SNAPSHOT environment setting, which is 19 long. pgcopydb only supports 7 bytes for this environment setting
2024-06-06 15:44:08.650 21 FATAL  cli_dump.c:132            Failed to read default values from the environment
```

We should be able store 1024-1 characters, not 7. This may be a bug in my current code.

src/bin/pgcopydb/cli_common.c
hanefi: I was planning to have a copy of this array with different target pointer values. Your suggestion is definitely an improvement on my current implementation.

dimitri:
> We should be able store 1024-1 characters, not 7. This may be a bug in my current code.

Well that looks like a classic mistake in C. Hint: 7 = 8 - 1, and that's your pointer size: `sizeof(void *) == 1`. So if I had to guess, you're taking the `sizeof()` of a `char *` instead of the `sizeof()` of an array, or a `char[]`. Which can be denoted the same way in C, hence lots of confusion.

src/bin/pgcopydb/cli_common.c
hanefi: Now that I spent some more time on this task, I realize only 2 items in this list makes sense for list commands: `splitMaxParts` and `estimateTableSizes`. It made more sense for me to create a new array with 2 items in `cli_list.c` instead of improving the structure to be reusable in two places.

dimitri: Makes sense, I was under the impression that more was needed from this place.

src/bin/pgcopydb/env_utils.c
```diff
@@ -215,3 +217,83 @@ get_env_pgdata_or_exit(char *pgdata)
...
+
+
+/*
+ * get_env_using_parser checks if the environment variable exists and if it does
+ * it parses it and stores it in the target. If the environment variable is not
+ * set, the target is not modified. If the environment variable is set but the
+ * parsing fails, an error is logged and the errors counter is incremented.
+ */
+void
+get_env_using_parser(EnvParser *parser, int *errors)
```
dimitri, line 229: I think this should take an Array of EnvParser items and walk through it, maybe using the usual way we approach arrays with an EnvParserArray structure definition that has both the array and the count.

src/bin/pgcopydb/env_utils.c
```diff
@@ -215,3 +217,83 @@ get_env_pgdata_or_exit(char *pgdata)
...
+{
+	if (env_exists(parser->envname))
+	{
```
dimitri, line 232: If you implement that the other way around (early exit), then you can have one less level of nesting, which will ease reading and maintenance.

src/bin/pgcopydb/env_utils.h
```diff
@@ -8,6 +8,22 @@
...
+
+typedef struct EnvParser
+{
+	char *envname;
+	enum EnvType type;
+	void *target;
+	int minValue;
+	int maxValue;
+} EnvParser;
+
```
dimitri, line 26: What about the following?

``` c
typedef struct EnvParserArray
{
    int count;
    EnvParser *array;
} EnvParserArray;
```

hanefi:
> > We should be able store 1024-1 characters, not 7. This may be a bug in my current code.
> 
> Well that looks like a classic mistake in C. Hint: 7 = 8 - 1, and that's your pointer size: `sizeof(void *) == 8`. So if I had to guess, you're taking the `sizeof()` of a `char *` instead of the `sizeof()` of an array, or a `char[]`. Which can be denoted the same way in C, hence lots of confusion.

I also realized that this is the case. I added a new member to the `EnvParser` struct to store sizes of character arrays.

src/bin/pgcopydb/cli_common.c
```diff
@@ -228,160 +235,43 @@ cli_copydb_getenv(CopyDBOptions *options)
...
-				++errors;
-			}
-		}
-		else
-		{
-			/* errors have already been logged */
-			++errors;
-		}
-	}
+	get_env_using_parsers(&parserArray, &errors);
```
dimitri, line 267: Now you can return a bool and use the standard error handling if form?

src/bin/pgcopydb/env_utils.c
hanefi: Instead of changing this function, I introduced `get_env_using_parsers` later in the file. This new function iterates over the array elements and call `get_env_using_parser`.

This way I did not increase the indentation by one in this function. There is enough levels already here.

src/bin/pgcopydb/env_utils.c
```diff
@@ -8,8 +8,11 @@
 
 #include "defaults.h"
 #include "env_utils.h"
+#include "parsing_utils.h"
+#include "string_utils.h"
 #include "log.h"
 
+void get_env_using_parser(EnvParser *parser, int *errors);
```
dimitri, line 15: Return a bool like every where else?

src/bin/pgcopydb/env_utils.h
```diff
@@ -17,4 +40,6 @@ bool get_env_dup(const char *name, char **result);
 bool get_env_dup_with_fallback(const char *name, char **result, const char *fallback);
 bool get_env_pgdata(char *pgdata);
 void get_env_pgdata_or_exit(char *pgdata);
+void get_env_using_parsers(EnvParserArray *parsers, int *errors);
```
dimitri, line 43: Return a bool?

src/bin/pgcopydb/env_utils.c
```diff
@@ -215,3 +218,95 @@ get_env_pgdata_or_exit(char *pgdata)
...
+ * it parses it and stores it in the target. If the environment variable is not
+ * set, the target is not modified. If the environment variable is set but the
+ * parsing fails, an error is logged and the errors counter is incremented.
+ */
+void
+get_env_using_parser(EnvParser *parser, int *errors)
+{
+	if (!env_exists(parser->envname))
+	{
+		return;
```
dimitri, line 234:
``` c
return true;
```

src/bin/pgcopydb/env_utils.h
```diff
@@ -8,6 +8,29 @@
...
+	ENV_TYPE_BOOL
+};
+
+typedef struct EnvParser
+{
+	char *envname;      /* name of the environment variable */
+	enum EnvType type;
+	void *target;
+	int targetSize;     /* for strings, 0 for other types */
+	int minValue;       /* for integers, 0 if there is no lower bound */
```
dimitri, line 24: We might have a use for zero as the lower bound, I would use -1 here. What do you think?

dimitri (changes requested):

src/bin/pgcopydb/env_utils.h
hanefi: I used to be able to leave out the last members in struct assignments as I knew they will be filled with zeros. Now that the default is -1, I explicitly set the values of all the members.

Although the earlier method looked better to me, I can see the benefit of showing it explicitly that the allowed values of the options are not bounded.

dimitri: You could also introduce a boolean such as "bounded", default is false and then no magic values are needed anymore.

src/bin/pgcopydb/env_utils.c
```diff
@@ -215,3 +218,104 @@ get_env_pgdata_or_exit(char *pgdata)
...
+			else if (!stringToInt(envValue, (int *) parser->target) ||
+					 (parser->lowerBounded &&
+					  *((int *) parser->target) < parser->minValue) ||
+					 (parser->upperBounded &&
+					  *((int *) parser->target) > parser->maxValue))
+			{
+				log_fatal("Failed to parse \"%s\": \"%s\"", parser->envname,
+						  envValue);
+				return false;
+			}
```
dimitri, line 277: I think we need a specific error message for values out of bound, so that the user has a chance to fix it.

src/bin/pgcopydb/env_utils.h
hanefi: I introduced `lowerBounded` and `upperBounded` struct members.

src/bin/pgcopydb/cli_common.c
```diff
@@ -212,104 +212,64 @@ cli_copydb_getenv(CopyDBOptions *options)
...
+	/* Fill in the defaults before reading environment variables */
+	options->tableJobs = DEFAULT_TABLE_JOBS;
+	options->indexJobs = DEFAULT_INDEX_JOBS;
+	options->restoreOptions.jobs = DEFAULT_RESTORE_JOBS;
+	options->lObjectJobs = DEFAULT_LARGE_OBJECTS_JOBS;
+	options->splitTablesLargerThan.bytes = DEFAULT_SPLIT_TABLES_LARGER_THAN;
+
+	EnvParser parsers[] = {
+		{ PGCOPYDB_TABLE_JOBS, ENV_TYPE_INT,
+		  &(options->tableJobs), 0, 1, 1, 1, 128 },
```
dimitri, line 224: I would use `true` instead of 1 for the boolean value for the “bounded” slot of the structure here.

dimitri (changes requested):

src/bin/pgcopydb/env_utils.c
hanefi: Good suggestion.

I needed a way to manipulate character arrays, and I decided to use `PQExpBuffer` to generate the error message. If you can think of a better way to implement this, please share it with me.

src/bin/pgcopydb/env_utils.c
```diff
@@ -215,3 +219,132 @@ get_env_pgdata_or_exit(char *pgdata)
...
+				appendPQExpBuffer(errorMessage,
+								  "Failed to parse \"%s\": \"%s\", "
+								  "expected an integer",
+								  parser->envname,
+								  envValue);
+
+				if (parser->lowerBounded)
+				{
+					appendPQExpBuffer(errorMessage,
+									  " greater than %d",
```
dimitri, line 286: or equal?

src/bin/pgcopydb/env_utils.c
```diff
@@ -215,3 +219,132 @@ get_env_pgdata_or_exit(char *pgdata)
...
+
+				if (parser->lowerBounded && parser->upperBounded)
+				{
+					appendPQExpBuffer(errorMessage, " and");
+				}
+
+				if (parser->upperBounded)
+				{
+					appendPQExpBuffer(errorMessage,
+									  " less than %d",
```
dimitri, line 298: Or equal?

src/bin/pgcopydb/env_utils.c
hanefi: Would you prefer

`greater than or equal to << parser->minValue >>`

or

`greater than << parser->minValue - 1 >>` ?

In my opinion, it is better to keep the number of characters small. Hence I added `+/- 1` instead of adding `or equal` phrases.

hanefi: Let me share an example. I created 2 sample error messages. I like the first one more, but the other one is also ok.

```
Failed to parse "PGCOPYDB_RESTORE_JOBS": "1234", expected an integer greater than 0 and smaller than 129
```

```
Failed to parse "PGCOPYDB_RESTORE_JOBS": "1234", expected an integer greater than or equal to 1 and smaller than or equal to 128
```

dimitri: What about using math operators `<=` and `>=` ?

hanefi: `Failed to parse "PGCOPYDB_RESTORE_JOBS": "1234", expected an integer >= 1 and <= 128`

It is definitely shorter now. I do not prefer having a binary operator mixed in a natural language. Especially when the left side is missing an expression.

Nevertheless, it is no big deal as it conveys the message clearly either way.

hanefi: Another idea:

`Failed to parse "PGCOPYDB_RESTORE_JOBS": "1234", expected an integer i that satisfies 1 <= i <= 128`

That feels even weirder. Never mind.

PostgreSQL usually shows the min and max values in the detail or hint of the error message. I don't think  this is an option for us. I'd rather have the information on the same line as our users fail to read the consecutive lines of log messages quite frequently.

```c
 ereport(ERROR,
		(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
			errmsg("value %s out of bounds for option \"%s\"",
				value, option->gen->name),
			errdetail("Valid values are between \"%d\" and \"%d\".",
					optint->min, optint->max)));
```

dimitri: Mixing with operators helps quick-reading through the logs, even if it somewhat looks weird. I think I like the `1 <= jobs <= 128` best for the quick-reading aspect to it, even though it's the worst option in terms of being a “proper” English phrase.

dimitri (approved):

PR #814, opened 2024-06-11
Fix intialisation of the start time, fixing timing reports on subsequent runs

### LINKED ISSUE #697: End report after a db clone shows higher numbers than expected
So while transferring a database on the main branch, I noticed the report is showing elevated numbers. Here is the estimation at the beginning, which matches our estimate of our ~150 GB database:

```
2024-03-05 13:28:35.885 75 INFO   copydb_schema.c:773       Table public.pgbench_accounts is 125 GB large, 2687 COPY processes will be used, partitioning on aid.
2024-03-05 13:28:35.886 75 INFO   copydb_schema.c:665       Fetched information for 4 tables (including 1 tables split in 2687 partitions total), with an estimated total of 1000 million tuples and 125 GB on-disk
```

But as you can see below, the report shows highly elevated numbers in the TB.

For example, the transmitted bytes in the first table shows 248 TB, and the column for transfer shows 328 TB.

Also, the copy duration for pgbench_accounts is showing 2939 days, which might contribute to the math issue here.

This didn't show these numbers until 0.15 and later, this test was done in this recent main: `Running pgcopydb version 0.15.32.g67df98b from "/usr/local/bin/pgcopydb"`.

```
2024-03-05 17:32:00.891 75 INFO   pgcmd.c:998                /usr/bin/pg_restore --dbname 'postgres://target@target_host:5432/d8v0iusebghvok?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --single-transaction --no-owner --no-acl --exclude-schema _heroku --use-list /tmp/pgcopydb/schema/post-filtered.list /tmp/pgcopydb/schema/post.dump
2024-03-05 17:32:00.968 75 INFO   cli_clone_follow.c:617    All step are now done,  4h03m elapsed
2024-03-05 17:32:00.969 75 INFO   summary.c:3152            Printing summary for 4 tables and 3 indexes
2024-03-05 17:32:02.017 75 NOTICE summary.c:2835            Storing migration summary in JSON file "/tmp/pgcopydb/summary.json"
    OID | Schema |             Name | Parts | copy duration | transmitted bytes | indexes | create index duration 
  ------+--------+------------------+-------+---------------+-------------------+---------+----------------------
  16530 | public | pgbench_accounts |  2687 |      2939d16h |            248 TB |       1 |                46m17s
  16527 | public |  pgbench_tellers |     1 |         4s375 |           1540 kB |       1 |                   14s
  16533 | public | pgbench_branches |     1 |         3s193 |             96 kB |       1 |                 2s309
  16524 | public |  pgbench_history |     1 |         3s127 |               0 B |       0 |                   0ms
                                                 Step   Connection    Duration    Transfer   Concurrency
   --------------------------------------------------   ----------  ----------  ----------  ------------
     Catalog Queries (table ordering, filtering, etc)       source       1m05s                         1
                                          Dump Schema       source       611ms                         1
                                       Prepare Schema       target       120ms                         1
        COPY, INDEX, CONSTRAINTS, VACUUM (wall clock)         both       4h03m                        17
                                    COPY (cumulative)         both       1d02h      328 TB             8
                            CREATE INDEX (cumulative)       target      46m31s                         1
                             CONSTRAINTS (cumulative)       target       2s001                         1
                                  VACUUM (cumulative)       target         0ms                         8
                                      Reset Sequences         both        29ms                         1
                           Large Objects (cumulative)         both                                     4
                                      Finalize Schema         both        98ms                         1
   --------------------------------------------------   ----------  ----------  ----------  ------------
                            Total Wall Clock Duration         both       4h03m                        22
```

Is anyone else seeing this issue perhaps?

rimbi:
This fixes the incorrect duration on the second run issue. ~I am not able to reproduce the incorrect volume bug.~ The transmitted bytes using `PqgetCopyData` seems to be less than the table size obtained by the function `pg_relation_size`. This is the reason for the difference in the final summary tables. But I could not understand the ratianle behind that yet.

Make sure that the `startTime` is set on each catalog start timing.

I could not find any place where the previous if check is still valid. When there is a need to perform this check it can be done at the place where `catalog_start_timing` is invoked.


dimitri:
> This fixes the incorrect duration on the second run issue. ~I am not able to reproduce the incorrect volume bug.~ 

Nice thanks.

> The transmitted bytes using `PqgetCopyData` seems to be less than the table size obtained by the function `pg_relation_size`. This is the reason for the difference in the final summary tables. But I could not understand the ratianle behind that yet.

The `pg_relation_size()` includes the file format overhead and the bloat on-disk, whereas the transmitted bytes is not counting the protocol overhead, nor the disk layout overhead, and the bloat is not sent over the network at all,

> Make sure that the `startTime` is set on each catalog start timing.
> 
> I could not find any place where the previous if check is still valid. When there is a need to perform this check it can be done at the place where `catalog_start_timing` is invoked.

I think we used to have another layer of API that we would use at some places, but that might have gotten refactored.

Please also see about the other side of the problem as reported in #697 : I believe we still have a problem in the way the total size is computed and reported when using several partitions.

dimitri (approved):

PR #816, opened 2024-06-11
Standalone sequences do not get reset when filtering

kbarber:
Standalone sequences not explicitly tied to a table do not get reset.

This is because they do not have relationships to tables or attributes compared to default and identity sequences so the filtering mostly wasn't doing the right thing here.

This patch adds parts to the SQL that allow for filtering just using the namespace from the sequence table only in that case.

Fixes: https://github.com/dimitri/pgcopydb/issues/817


tests/unit/copydb.sh
```diff
@@ -28,7 +28,10 @@ EOF
 
 # pgcopydb fork uses the environment variables
 export PGCOPYDB_SPLIT_TABLES_LARGER_THAN="2MB"
-pgcopydb fork --skip-collations --fail-fast --debug
+printf "[exclude-schema]
```
kbarber, line 31: Any ideas on how to test this with & without filters?

kbarber: Adding it to the filtering area

dimitri (approved):

dimitri:
Hi @kbarber ; thanks for the contribution! I want to stress the unit testing coverage you added, that was key for me to efficiently approve your PR. Cheers!

kbarber:
> Hi @kbarber ; thanks for the contribution! I want to stress the unit testing coverage you added, that was key for me to efficiently approve your PR. Cheers!

No problemo my friend.

PR #818, opened 2024-06-12
Fix #811: Error:Failed to create pg_restore list file : out of memory

### LINKED ISSUE #811: Error:Failed to create pg_restore list file : out of memory
Migration failing with OOM error while reading whole list file in-memory .
```
2024-06-06 21:13:59.896 69 ERROR  dump_restore.c:781        Failed to create pg_restore list file: out of memory
2024-06-06 21:13:59.896 69 ERROR  dump_restore.c:212        Failed to prepare the pg_restore --use-list catalogs, see above for details
2024-06-06 21:13:59.896 69 ERROR  cli_clone_follow.c:600    Failed to prepare schema on the target database, see above for details
2024-06-06 21:13:59.896 69 ERROR  cli_clone_follow.c:506    Failed to clone source database, see above for details
2024-06-06 21:14:00.062 51 ERROR  cli_clone_follow.c:768    clone process 69 has terminated [6]
```

Code path - https://github.com/dimitri/pgcopydb/blob/022f2a37f2317fc19c5376e7c9b5fb110e2c5286/src/bin/pgcopydb/dump_restore.c#L799
I also found similar error 

Discussion with Dimitri:
The error message actually comes from the pgcopydb source code, so that's a problem with memory handling within pgcopydb itself. At the moment the strategy that has been implemented consists of reading the whole list file in-memory, and then parsing it line by line (walking through an array), preparing a new version of it in memory again, only to write that memory to disk at the end.
 
I think it would be fairly easy to replace that mechanism with line-by-line reading of the file using [fgets](https://man7.org/linux/man-pages/man3/fgets.3p.html) instead of fread(), and then writing the line with fwrite() rather than accumulating in-memory.

rimbi:
Instead of loading the contents of the list file before parsing and transforming, this change set uses the stream parsing method by introducing `FileIterator` and `ArchiveIterator` concepts.


src/bin/pgcopydb/archive_iterator.h
```diff
@@ -0,0 +1,20 @@
...
+
+#ifndef ARCHIVE_ITER_H
+#define ARCHIVE_ITER_H
+
+#include <stdbool.h>
+
+typedef struct ArchiveIterator ArchiveIterator;
+typedef struct ArchiveContentItem ArchiveContentItem;
+
+ArchiveIterator * archive_iterator_from(const char *filename);
```
dimitri, line 14: Have a look around the iterator constructs in catalog.c, and consider having an API that looks more like that.

src/bin/pgcopydb/cli_restore.c
```diff
@@ -565,19 +567,26 @@ cli_restore_schema_parse_list(int argc, char **argv)
...
-			/* errors have already been logged */
+			log_error("Failed to open archive iterator for \"%s\"", filename);
 			exit(EXIT_CODE_INTERNAL_ERROR);
 		}
 
-		log_notice("Read %d archive items in \"%s\"", contents.count, filename);
+		ArchiveContentItem *item = NULL;
 
-		for (int i = 0; i < contents.count; i++)
+		for (;;)
```
dimitri, line 579: Please use a while (iter_next() !=
NULL), the most obvious form.

src/bin/pgcopydb/cli_restore.c
```diff
@@ -565,19 +567,26 @@ cli_restore_schema_parse_list(int argc, char **argv)
...
 		{
-			ArchiveContentItem *item = &(contents.array[i]);
+			if (!archive_iterator_next(iterator, &item))
```
dimitri, line 581: There is no call to the destroy api, can we remove the destroy api?

src/bin/pgcopydb/dump_restore.c
```diff
@@ -622,31 +624,64 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
 	{
 		/* errors have already been logged */
 		return false;
 	}
 
 	/* edit our post.list file now */
-	PQExpBuffer listContents = createPQExpBuffer();
-
-	if (listContents == NULL)
+	PQExpBuffer line = createPQExpBuffer();
```
dimitri, line 636: A single pg_restore archive TOC item has a maximum known length. Consider switching to local static allocation instead; or dynamic allocation of a known size.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -622,31 +624,64 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
+	}
+
+	ArchiveIterator *iterator = archive_iterator_from(listOutFilename);
+	if (iterator == NULL)
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	if (!unlink_file(listFilename))
```
dimitri, line 651: Quick review, but I'm confused by this unlink call here.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -622,31 +624,64 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
+	{
+		/* errors have already been logged */
 		return false;
 	}
 
-	/* for each object in the list, comment when we already processed it */
-	for (int i = 0; i < contents.count; i++)
+
+	ArchiveContentItem *item = NULL;
+	for (;;)
```
dimitri, line 659: Same comment, use a while loop here. Make it obvious.

src/bin/pgcopydb/file_iterator.c
```diff
@@ -0,0 +1,118 @@
...
+file_iterator_next(FileIterator *iterator, char **line)
+{
+	*line = NULL;
+	if (iterator->line)
+	{
+		free(iterator->line);
+		iterator->line = NULL;
+	}
+	size_t len = 0;
+	if (getline(&(iterator->line), &len, iterator->file) != -1)
```
dimitri, line 66: Note to myself: compare getline() and fgets()

src/bin/pgcopydb/file_iterator.c
```diff
@@ -0,0 +1,118 @@
...
+
+	return true;
+}
+
+
+/*
+ * Get the line number
+ */
+size_t
+file_iterator_get_line_number(FileIterator *iterator)
```
dimitri, line 87: Useless and arguably harmful, please remove accessor functions. Design patterns are tricks to overcome bad OOP designs in the language. We do not want nor need that here, in C. 

Reference in a Peter Norvig presentation on the topic to be added later.

src/bin/pgcopydb/file_iterator.c
```diff
@@ -0,0 +1,118 @@
...
+{
+	return iterator->line_num;
+}
+
+
+/*
+ * Get the file name
+ */
+const char *
+file_iterator_get_file_name(FileIterator *iterator)
```
dimitri, line 97: Same.

src/bin/pgcopydb/file_iterator.c
```diff
@@ -0,0 +1,118 @@
...
+{
+	return iterator->filename;
+}
+
+
+/*
+ * Destroy the iterator and free resources
+ */
+void
+file_iterator_destroy(FileIterator *iterator)
```
dimitri, line 107: The resources should be managed by the caller, in a way that makes resource management obvious. Then remove this api.

src/bin/pgcopydb/file_iterator.c
```diff
@@ -0,0 +1,118 @@
...
+{
+	if (iterator->file)
+	{
+		fclose(iterator->file);
+	}
+	if (iterator->line)
+	{
+		free(iterator->line);
```
dimitri, line 115: We use a Garbage Collector, we don't call free() anywhere anymore.

src/bin/pgcopydb/file_iterator.c
```diff
@@ -0,0 +1,118 @@
...
+
+/*
+ * Destroy the iterator and free resources
+ */
+void
+file_iterator_destroy(FileIterator *iterator)
+{
+	if (iterator->file)
+	{
+		fclose(iterator->file);
```
dimitri, line 111: Using a function pointer based API like in the iterators defined in catalog.c and summary.c etc, you could make the resource management obvious and error prone. Please consider that option.

dimitri (changes requested):
I expected a very simple loop around reading one line at a time, rather than a whole new iterator API with a single call site. I believe this PR introduces a lot of complexity for no benefits. 

If you still feel strongly about using an abstraction layer for a single use-case, which the Postgres development team and practices opposes to, then please get inspiration from existing constructs in the code base.

rimbi:
> I expected a very simple loop around reading one line at a time, rather than a whole new iterator API with a single call site. I believe this PR introduces a lot of complexity for no benefits.
> 
> If you still feel strongly about using an abstraction layer for a single use-case, which the Postgres development team and practices opposes to, then please get inspiration from existing constructs in the code base.

Sorry but I have not been able to see a simple solution which just uses a single while loop. Therefore ended up with this solution. Apparently, one of us is missing something.

src/bin/pgcopydb/file_iterator.c
dimitri: See https://norvig.com/design-patterns/

src/bin/pgcopydb/file_iterator.c
dimitri: So `getline()` implements its own memory allocation, which looks nice, but I wonder if that works well with libgc, the Garbage Collector that we use. It might be easier for us to use `fgets()`.

src/bin/pgcopydb/file_iterator.c
dimitri: It might be that a better presentation about what I have in mind is found at https://www.infoq.com/presentations/Simple-Made-Easy/

rimbi:
@dimitri I think the latest changes fix the issues we talked about and ready for a merge. If you see any important issues and have the time and energy to fix them, feel free to do so. Otherwise, I can continue when I get back.

src/bin/pgcopydb/archive_iterator.c
```diff
@@ -0,0 +1,532 @@
...
+typedef struct ArchiveContext
+{
+	ArchiveIterCallback *callback;
+	void *context;
+	size_t line_number;
+	const char *filename;
+} ArchiveContext;
+
+static bool
+file_iter_callback(void *context, const char *line)
```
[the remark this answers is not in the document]
dimitri: Also the function name is wrong. We could have many different callbacks all being called by a file iterator. The information needed in the name is what the callback is doing, not from where it's called. For instance, this one could be named `parse_archive_list_entry_hook`.

src/bin/pgcopydb/cli_restore.c
```diff
@@ -553,6 +555,21 @@ cli_restore_roles(int argc, char **argv)
 }
 
 
+static bool
+print_archive_item(void *context, ArchiveContentItem *item)
```
dimitri, line 559: Missing function-level comment. Missing `_hook` suffix in the function name.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -567,6 +570,167 @@ copydb_target_finalize_schema(CopyDataSpec *specs)
 }
 
 
+typedef struct ArchiveContext
+{
+	CopyDataSpec *specs;
+	PQExpBuffer line;
+	FILE *listFile;
+} ArchiveContext;
```
dimitri, line 578: This is the same name as in `archive_iterator.c` but with a different definition. While the C compiler has no problem dealing with that, I do. Please rename.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -567,6 +570,167 @@ copydb_target_finalize_schema(CopyDataSpec *specs)
...
+
+static bool
+copydb_write_restore_list_callback(void *ctx, ArchiveContentItem *item)
```
dimitri, line 581: Missing function-level comments. Also, consider using `_hook` suffix instead of `_callback` like every other place in the rest of the code. Same concept, same name, helps maintenance (cognitive overhead).

src/bin/pgcopydb/dump_restore.c
```diff
@@ -567,6 +570,167 @@ copydb_target_finalize_schema(CopyDataSpec *specs)
...
+				   item->catalogOid,
+				   item->objectOid,
+				   item->restoreListName);
+	}
+
+	/*
+	 * Since printfPQExpBuffer generates a NULL-terminated string,
+	 * we have the ability to use the same buffer repeatedly.
+	 */
+	printfPQExpBuffer(line, "%s%d; %u %u %s %s\n",
```
dimitri, line 715: It seems to me we should still call `PQExpBufferBroken(line)` each time we call `printPQExpBuffer` and report an OOM error in case the buffer is broken. If not, then we should add a comment explaining why that's not necessary.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -622,169 +786,48 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
 		log_error(ALLOCATION_FAILED_ERROR);
-		destroyPQExpBuffer(listContents);
+		destroyPQExpBuffer(line);
 		return false;
 	}
 
-	/* for each object in the list, comment when we already processed it */
-	for (int i = 0; i < contents.count; i++)
+	FILE *listFile = fopen_with_umask(listFilename, "wb", FOPEN_FLAGS_W, 0644);
+	if (!listFilename)
```
dimitri, line 807: This needs to be written `if (listFile == NULL)`, and as an immediate benefit would fix the bug where you compare the `const char *listFilename` instead of the `FILE *listFile` after opening.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -622,169 +786,48 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
-						  contents.array[i].objectOid,
-						  contents.array[i].description,
-						  contents.array[i].restoreListName);
+		log_error("Failed to open list file \"%s\": %m", listFilename);
+		return false;
 	}
+	ArchiveContext context = { 0 };
+	context.specs = specs;
+	context.line = line;
+	context.listFile = listFile;
```
dimitri, line 815: Please embed the structure slots definitions within the static definition, as it's all pointers anyways, like we do in all the rest of the code.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -622,169 +786,48 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
-	if (PQExpBufferBroken(listContents))
+	bool res = archive_iter(listOutFilename, &context,
+							copydb_write_restore_list_callback);
+	if (!res)
 	{
-		log_error("Failed to create pg_restore list file: out of memory");
-		destroyPQExpBuffer(listContents);
-		return false;
+		log_error("Failed to write filtered pg_restore list file at \"%s\"",
+				  listFilename);
```
dimitri, line 822: I believe we should append “see above for details” to this error message.

src/bin/pgcopydb/archive_iterator.c
```diff
@@ -0,0 +1,532 @@
...
+}
+
+
+bool
+archive_iter(const char *filename,
+			 void *context,
+			 ArchiveIterCallback *callback)
+{
+	ArchiveContext archiveContext = { callback, context, 0, filename };
+	return file_iter(filename, &archiveContext, file_iter_callback);
```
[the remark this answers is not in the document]
dimitri: After reading through the file_iterator implementation: it needs a lot of work in terms of API and code organization in order for it to look like all the other iterators. There is only one call-site for it, which is here in the archive iterator.

My advice is to give the responsibility of reading the file line-by-line to the archive iterator itself, removing the need for a way to compose iterators on-top of one another. If someday we have 3 iterators in the code that all loop around `fgets` then we will see about how to refactor the code and how to compose two iterators properly. But this is a problem for another day.

dimitri (changes requested):
This PR is taking a good shape, thanks for all the work!

Lots of nitpicking still, and a design issue with how to compose iterators that I think would be best to avoid entirely at the moment, as it won't have much benefits to the current PR anyways.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -804,21 +810,20 @@ copydb_write_restore_list(CopyDataSpec *specs, PostgresDumpSection section)
...
+	ArchiveContext context = { .specs = specs, .line = line, .listFile = listFile };
 
 	bool res = archive_iter(listOutFilename, &context,
-							copydb_write_restore_list_callback);
+							copydb_write_restore_list_hook);
 	if (!res)
 	{
-		log_error("Failed to write filtered pg_restore list file at \"%s\"",
+		log_error("Failed to write filtered pg_restore list file at \"%s\""
+				  ", see above for details",
```
dimitri, line 826: We have 210 occurrences of "see above for details" with the coma on the previous line, why change the code style in this specific part?

dimitri (changes requested):
We need to find a better/simpler way to compose the file iterator and the archive iterator, that's not it yet.

dimitri (changes requested):
I spend some time thinking about (the previous iteration of) this PR. I believe I now understand what feels wrong to me about the way your code is organized. 

In all the already existing iterators, the context belongs to the callback function. The iterator job is to handle resource allocation (here, open/close file and line memory allocation), possibly at the iterator level (re-use same memory area for each line read) meaning in the `_init` function. When it's time to call the user-provided callback, call it with the `void *` context that was given to the main function, and with the object/item provided this round, here an instance of the structure `ArchiveContentItem`.

Your code mixes the responsibilities and uses the context from within the iterator code itself. Even in the previous version, where you still had a file_iterator concept, cleaning-up responsibilities would have allowed you to build an archive_iterator on top of a file_iterator, by having a single context for the single user-provided callback, and handling the parsing of the archive TOC entry from within the Archive TOC iterator itself.

``` c
typedef struct ArchiveTOCIterator
{
    FileIterator *file_iter;
    ArchiveContentItem *tocEntry;
} ArchiveTOCIterator;

bool
iterate_archive_toc_init(ArchiveTOCIterator *iter)
{
    FileIterator *file_iter = calloc();

    if (!file_iterator_init(file_iter, iter->filename))
    {
        return false;
    }

    iter->tocEntry = calloc();

    return true;
}

bool
iterate_archive_toc_next(ArchiveTOCIterator *iter)
{
    // grab next line
    if (!file_iterator_next(iter->file_iter))
    {
        return false;
    }

    // now parse the Archive TOC entry
    if (!parse_archive_list_entry(iter->tocEntry, iter->file_iter->line))
    {
        return false;
    }

    return true;
}

bool
iterate_archive_toc_finishArchiveTOCIterator *iter)
{
    return file_iterator_finish(iter->file_iter);
}

typedef bool (ArchiveTOCIteratorFun)(void *context, ArchiveContentItem *item);

bool
iterate_archive_toc(const char *filename, void *context, ArchiveTOCIteratorFun *callback)
{
    ArchiveTOCIterator *toc_iter = calloc();

    if (!iterate_archive_toc_init(toc_iter, filename))
    {
        return false;
    }

    for (;;)
    {
        if (!iterate_archive_toc_next(toc_iter))
        {
            return false;
        }

        ArchiveContentItem *entry = iter->tocEntry;

        if (entry == NULL)
        {
            // finish
            return true;
        }

        if (!callback)(context, entry)
        {
            return false;
        }
    }

    // finish

    return true;
}
```

What you tried to solve, using the file_iterator callback to call the archive_iterator callback, is more complex: the archive TOC callback would not be called from the `iterate_archive_toc_next()` function, but from the callback given to the file_iterator, in a way that the context of that lower-level callback then contains the upper-level callback. That's too confusing for my abilities to debug the code.

src/bin/pgcopydb/archive_iterator.c
```diff
@@ -0,0 +1,654 @@
...
+			break;
+		}
+
+		/* now call the provided callback */
+		if (!parse_archive_list_entry_and_callback(&iterContext, line, callback))
+		{
+			log_error("Failed to iterate over list of colls, "
+					  "see above for details");
+			return false;
+		}
```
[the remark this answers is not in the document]
dimitri: Now that I have done the exercise of sketching the code I wanted to see, I understand better what you did, and how the two levels of context and callback works in what you did. I still feel like that's too complex, even though the composition of the two levels of callbacks is mostly kept within the outer callback function. It's not entirely cleanly organized, or I am still missing how it works, which means I will never be able to debug it. Let's make it obvious.

rimbi:
@dimitri I am lost in these conversations and have difficulty understanding what you want. Would you prefer to finish it yourself? That way it might take much less time to get what you want.

PR #824, opened 2024-06-25
[WIP] Refactoring of the iterators

rimbi:
The purpose of this MR is to remove or minimise duplication in the implementation of the iterators by using an iterator ADT.

This current version is only a beginning and should be considered as a PoC.

### DIFF
--- a/src/bin/pgcopydb/catalog.c
+++ b/src/bin/pgcopydb/catalog.c
@@ -10,6 +10,7 @@
 #include <time.h>
 #include <unistd.h>
 
+#include "iterator.h"
 #include "sqlite3.h"
 
 #include "catalog.h"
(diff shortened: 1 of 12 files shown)


dimitri (commented):
First review, one small thing and a major problem.

Small thing: I don't like the `for_each` function name, because C does not allow to provide control operators in the code, and it looks too generic. A better name would be `iterate` maybe.

Major problem: if I understand correctly, with this approach, we lose the ability for the compiler to check that we're using the expected data-structure with the callback function, because of the `void *` casting on the current iteration's object. That is why I have written the code in a copy/paste mode: I failed to find any other way to do it in C that gives us the compiler checks around data types for the iterator object and the callback function, ensuring that there is no mix-and-match.

rimbi:
`for_each` is a common method/function name used to consume iterators. For example: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

I don't think that using void pointer casts is a big problem. Aside from being widely used in other projects, it is already used in many places in the current sources.

But anyway, I am closing the PR.

dimitri:
I admit that there is a level of personal preference here. 

My preference is for the callback function to have a “proper” signature with known data types in terms of iterator API, and allowing `void *` only for the context parts. That's because the context is never used in the iterator, which only bounces it back to the callback, so typically the context structure definition / instance happens nearby at the callback code place, and the iterator code can be maintained separately in its own file and does not have to know what is the context. Iterator code never casts the `void *` pointer back to anything else.

Issue #819, opened 2024-06-17
Migration failed to create a replication slot due to concurrent transactions. Failed after 5 hours


pranavidandu:
Migration failed to create snapshot due to concurrent transactions and failed after 5 hours. Can we have a timeout in these scenarios and inform the customer about this error?

```

2024-06-13 16:13:20.388 80 INFO   main.c:136                Running pgcopydb version 0.16 from "/usr/lib/postgresql/15/bin/pgcopydb"
2024-06-13 16:13:20.392 80 INFO   cli_common.c:1491         [SOURCE] Copying database from "<source-conn-string>"
2024-06-13 16:13:20.392 80 INFO   cli_common.c:1492         [TARGET] Copying database into "<target-conn-string>"
2024-06-13 16:13:20.407 80 INFO   copydb.c:105              Using work dir "xxxxxx"
2024-06-13 16:13:20.555 80 SQL    pgsql.c:502               Connecting to [source] "<source-conn-string>"
2024-06-13 21:37:45.518 80 ERROR  pgsql.c:4041              Failed to send CREATE_REPLICATION_SLOT command
```

Source server logs:
```
2024-06-13 16:13:20 UTC-666b1aa0.7a700-LOG: logical decoding found initial starting point at 6F4/32BC3A90
--
2024-06-13 16:13:20 UTC-666b1aa0.7a700-DETAIL: Waiting for transactions (approximately 1) older than 16601567 to end.
```

Pg stat sessions logs:
```
PreciseTimeStamp | database_name | application_name | state | wait_event_type | wait_event
-- | -- | -- | -- | -- | --
2024-06-13 16:17:00.5181900 | <dbname> | pgcopydb[79] /usr/lib/postgresql/15/bin/pgcopydb | active | Lock | transactionid
.
.
.
2024-06-13 21:33:38.3123537 | <dbname> | pgcopydb[79] /usr/lib/postgresql/15/bin/pgcopydb | active | Lock | transactionid

```

Repro link - https://issues.redhat.com/browse/DBZ-862

hanefi:
Do you suggest we set `lock_timeout` only for the `CREATE_REPLICATION_SLOT` command sent over the replication connection? What do you think is a meaningful value?

Can you share the next lines of output from pgcopydb. I believe the next lines of the logs after `Failed to send CREATE_REPLICATION_SLOT command:` (notice the colon at the end) should contain more information on the error.

marikkan-microsoft:
[...]

PR #826, opened 2024-06-26
Set lock_timeout for replication connections

hanefi:
pgcopydb fails to create a replication slot on systems that have long running transactions. This is because the CREATE_REPLICATION_SLOT command waits for the lock to be released. This patch adds a new connection parameter lock_timeout to the connection string for replication connections. This parameter is set to 60 seconds by default. This will reduce the waiting time for the CREATE_REPLICATION_SLOT command to fail.

Fixes: #819

### DIFF
--- a/src/bin/pgcopydb/copydb.c
+++ b/src/bin/pgcopydb/copydb.c
@@ -80,6 +80,28 @@ KeyVal connStringDefaults = {
 	}
 };
 
+/*
+ * We add an extra parameter to the connection string for replication to reduce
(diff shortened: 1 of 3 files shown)


hanefi:
This attempt was not really correct.

`lock_timeout` is not a connection setting listed in [docs](https://www.postgresql.org/docs/current/runtime-config-connection.html#RUNTIME-CONFIG-CONNECTION-SETTINGS). Instead it is a GUC that needs to be set.

We can not use SET or similar commands on a replication connection. We need another solution.

PR #828, opened 2024-06-28
Add support for multi-deletes.

VaibhaveS:
Continuation of #465

### DIFF
--- a/src/bin/pgcopydb/ld_transform.c
+++ b/src/bin/pgcopydb/ld_transform.c
@@ -1372,51 +1372,116 @@ coalesceLogicalTransactionStatement(LogicalTransaction *txn,
 {
 	LogicalTransactionStatement *last = txn->last;
 
-	LogicalMessageValuesArray *lastValuesArray =
-		&(last->stmt.insert.new.array->values);
+	int action = last->action;
 
-	LogicalMessageValuesArray *newValuesArray =
-		&(new->stmt.insert.new.array->values);
-
-	int capacity = lastValuesArray->capacity;
-	LogicalMessageValues *array = lastValuesArray->array;
-
-	/*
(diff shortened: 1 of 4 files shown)


dimitri (commented):
First review: thanks for the work!

  1. can we share more code between the INSERT and the DELETE paths? it seems like we are duplicating code parts that are “infrastructure” (memory resizing, array elements handling, not “functional” bits)
  2. I'm not sure about using the WHERE ... OR ... approach for the DELETE statement with multiple target rows ; I think I would prefer a JOIN using VALUES

VaibhaveS:
> * can we share more code between the INSERT and the DELETE paths? it seems like we are duplicating code parts that are “infrastructure” (memory resizing, array elements handling, not “functional” bits)

Yep.

> * I'm not sure about using the WHERE ... OR ... approach for the DELETE statement with multiple target rows ; I think I would prefer a JOIN using VALUES

Can you share how the query would look like? I tried constructing one, but it was making things more complicated.

hanefi:
I could not think of any reasonable way to write a JOIN using VALUES. I stumbled upon [a PostgreSQL docs page](https://www.postgresql.org/docs/current/sql-values.html) containing a note:

> VALUES lists with very large numbers of rows should be avoided, as you might encounter out-of-memory failures or poor performance. VALUES appearing within INSERT is a special case (because the desired column types are known from the INSERT's target table, and need not be inferred by scanning the VALUES list), so it can handle larger lists than are practical in other contexts.

I wonder if it actually makes sense to add support for multi deletes. Can we run a simple benchmark to understand whether or not this adds value?

VaibhaveS:
Let me get back with the perf results, meanwhile @dimitri let us know your thoughts.

PR #831, opened 2024-07-01
Fix out-of-memory issue in creating our pg_restore list file.

### LINKED ISSUE #811: Error:Failed to create pg_restore list file : out of memory

dimitri:
Instead of creating the whole contents of the file in-memory, write it to disk one line at a time, keeping only that line in-memory during the whole operation.

Also arrange to code to only read the pg_restore --list output one line at a time, no context is needed to be able to parse this file format (pg_dump archive Table Of Contents).

Fixes #811 
Replaces and closes #818

### DIFF
--- a/src/bin/pgcopydb/file_utils.h
+++ b/src/bin/pgcopydb/file_utils.h
@@ -54,6 +54,26 @@ bool write_file(char *data, long fileSize, const char *filePath);
 bool append_to_file(char *data, long fileSize, const char *filePath);
 bool read_file(const char *filePath, char **contents, long *fileSize);
 bool read_file_if_exists(const char *filePath, char **contents, long *fileSize);
+
+/* iterate over a file one line at a time */
+typedef bool (FileIterLinesFun)(void *context, char *line);
+
+bool file_iter_lines(const char *filename, size_t bufsize,
+					 void *context,
+					 FileIterLinesFun *callback);
+
+typedef struct FileLinesIterator
+{
+	const char *filename;
+	FILE *stream;
+	size_t bufsize;
+	char *line;                 /* malloc'ed area */
+} FileLinesIterator;
+
+bool file_iter_lines_init(FileLinesIterator *iter);
+bool file_iter_lines_next(FileLinesIterator *iter);
+bool file_iter_lines_finish(FileLinesIterator *iter);
+
 bool move_file(char *sourcePath, char *destinationPath);
 bool duplicate_file(char *sourcePath, char *destinationPath);
 bool create_symbolic_link(char *sourcePath, char *targetPath);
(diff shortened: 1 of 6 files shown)


src/bin/pgcopydb/file_utils.h
```diff
@@ -54,6 +54,26 @@ bool write_file(char *data, long fileSize, const char *filePath);
 bool append_to_file(char *data, long fileSize, const char *filePath);
 bool read_file(const char *filePath, char **contents, long *fileSize);
 bool read_file_if_exists(const char *filePath, char **contents, long *fileSize);
+
+/* iterate over a file one line at a time */
+typedef bool (FileIterLinesFun)(void *context, char *line);
+
+bool file_iter_lines(const char *filename, size_t bufsize,
+					 void *context,
+					 FileIterLinesFun *callback);
```
hanefi, line 63: Why do we have a function declaration and definition if it is not used anywhere in the codebase?

dimitri: The idea is that we have other places where we read a file line-by-line that could benefit from this iterator and one-line-at-a-time in memory approach. Also to implement the full iterator protocol, even though the composition with the archive iterator needs to be “open-coded” at the moment, using the lower-level API directly.

PR #835, opened 2024-07-03
Remove redundant catalog db attach while using restore parse-list 

arajkumar:
Before this commit, pgcopydb restore parse-list failed with following error.
```
11:44:52.406 233882 ERROR  Failed to attach '/tmp/pgcopydb/schema/source.db' as source
11:44:52.406 233882 ERROR  database source is already in use
```

cli_restore_prepare_specs already prepases catalogs.

### DIFF
--- a/src/bin/pgcopydb/cli_restore.c
+++ b/src/bin/pgcopydb/cli_restore.c
@@ -592,27 +592,6 @@ cli_restore_schema_parse_list(int argc, char **argv)
 
 	(void) cli_restore_prepare_specs(&copySpecs);
 
-	SourceFilters *filters = &(copySpecs.filters);
-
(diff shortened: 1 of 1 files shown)


dimitri:
> cli_restore_prepare_specs already prepases catalogs.

Does this one also needs to prepare the snapshot, like in the code you're removing, or can we skip that step (and why)?

arajkumar:
@dimitri copydb_fetch_schema_and_prepare_specs already takes care of [creating a new snapshot](https://github.com/dimitri/pgcopydb/blob/cce0894ead328b9d4e97243b6a730864855a32d9/src/bin/pgcopydb/copydb_schema.c#L84-L94) in case a valid one is not supplied. Hence, preparing a snapshot is redundant.

dimitri (approved):

PR #838, opened 2024-07-08
Use a fixed version for checkout action

hanefi:
This PR fixes our stylechecker tests using the following changes:

- Updated stylechecker image to a new version with updated packages
- Set a fixed version of checkout action
 to not get affected by later releases
- Set safe directory for git repository to fix a new set of errors I got during my tests

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -60,9 +60,14 @@ jobs:
 
   style_checker:
     runs-on: ubuntu-latest
-    container: citus/stylechecker:latest
+    container: citus/stylechecker:no-py
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -60,9 +60,14 @@ jobs:
 
   style_checker:
     runs-on: ubuntu-latest
-    container: citus/stylechecker:latest
+    container: citus/stylechecker:no-py
```
hanefi, line 63: I wanted to rebuild the docker image, but failed as some of the python package versions in the container no longer is available in Python Package Index. I ended up removing all of those broken dependencies in https://github.com/citusdata/the-process/pull/156 to create a new docker image that has working `citus_indent` tool.

dimitri (approved):

PR #839, opened 2024-07-09
Add byte-level progress to pgcopydb list progress

### LINKED ISSUE #388: Add byte-level progress to pgcopydb list progress
It would be useful if `pgcopydb list progress` would give more detailed information such as the number of bytes copied and maybe information from `pg_stat_progress_create_index`.

rimbi:
Fixes: #388

Allows the user to see the progress of the copy operation by executing the `pgcopydb list progress` command. The user will see that the duration and the bytes transferred will be increasing.

### DIFF
--- a/src/bin/pgcopydb/copydb.h
+++ b/src/bin/pgcopydb/copydb.h
@@ -448,7 +448,6 @@ bool copydb_table_create_lockfile(CopyDataSpec *specs,
 								  bool *isDone);
 
 bool copydb_save_copy_progress(CopyDataSpec *specs,
-							   CopyTableDataSpec *tableSpecs,
 							   uint64_t count,
 							   uint64_t bytes,
 							   instr_time lastSavingTime);
--- a/src/bin/pgcopydb/copydb.h
+++ b/src/bin/pgcopydb/copydb.h
@@ -436,7 +436,9 @@ bool copydb_process_table_data_worker(CopyDataSpec *specs);
 bool copydb_process_table_data_with_workers(CopyDataSpec *specs);
 
 bool copydb_copy_table(CopyDataSpec *specs, PGSQL *src, PGSQL *dst,
-					   CopyTableDataSpec *tableSpecs);
+					   CopyTableDataSpec *tableSpecs,
+					   CopyProgressCallback onCopyProgress,
+					   void *onCopyProgressContext);
 
 
 bool copydb_table_create_lockfile(CopyDataSpec *specs,
@@ -444,7 +446,7 @@ bool copydb_table_create_lockfile(CopyDataSpec *specs,
 								  PGSQL *dst,
 								  bool *isDone);
 
-bool copydb_mark_table_as_done(CopyDataSpec *specs,
+bool copydb_save_copy_progress(CopyDataSpec *specs,
 							   CopyTableDataSpec *tableSpecs);
 
 bool copydb_table_parts_are_all_done(CopyDataSpec *specs,
--- a/src/bin/pgcopydb/copydb.h
+++ b/src/bin/pgcopydb/copydb.h
@@ -533,6 +533,9 @@ bool summary_add_table(DatabaseCatalog *catalog,
 bool summary_finish_table(DatabaseCatalog *catalog,
 						  CopyTableDataSpec *tableSpecs);
 
+bool summary_update_table(DatabaseCatalog *catalog,
+						  CopyTableDataSpec *tableSpecs);
+
 bool summary_delete_table(DatabaseCatalog *catalog,
 						  CopyTableDataSpec *tableSpecs);
 
--- a/src/bin/pgcopydb/copydb.h
+++ b/src/bin/pgcopydb/copydb.h
@@ -10,6 +10,7 @@
 #include "copydb_paths.h"
 #include "filtering.h"
 #include "lock_utils.h"
+#include "portability/instr_time.h"
 #include "queue_utils.h"
 #include "pgcmd.h"
 #include "pgsql.h"
@@ -437,8 +438,8 @@ bool copydb_process_table_data_with_workers(CopyDataSpec *specs);
 
 bool copydb_copy_table(CopyDataSpec *specs, PGSQL *src, PGSQL *dst,
 					   CopyTableDataSpec *tableSpecs,
-					   CopyProgressCallback onCopyProgress,
-					   void *onCopyProgressContext);
+					   void *onCopyProgressContext,
+					   CopyProgressCallback onCopyProgress);
 
 
 bool copydb_table_create_lockfile(CopyDataSpec *specs,
@@ -447,7 +448,10 @@ bool copydb_table_create_lockfile(CopyDataSpec *specs,
 								  bool *isDone);
 
 bool copydb_save_copy_progress(CopyDataSpec *specs,
-							   CopyTableDataSpec *tableSpecs);
+							   CopyTableDataSpec *tableSpecs,
+							   uint64_t count,
+							   uint64_t bytes,
+							   instr_time lastSavingTime);
 
 bool copydb_table_parts_are_all_done(CopyDataSpec *specs,
 									 CopyTableDataSpec *tableSpecs,
--- a/src/bin/pgcopydb/extensions.c
+++ b/src/bin/pgcopydb/extensions.c
@@ -7,6 +7,7 @@
 #include <getopt.h>
 #include <inttypes.h>
 #include <sys/wait.h>
+#include <time.h>
 #include <unistd.h>
 
 #include "catalog.h"
@@ -173,10 +174,12 @@ copydb_copy_ext_table(PGSQL *src, PGSQL *dst, char *qname, char *condition)
 		.srcWhereClause = condition,
 		.dstQname = qname,
 		.dstAttrList = "",
-		.bytesTransmitted = 0
+		.bytesTransmitted = 0,
+		.bytesTransmittedBeforeSavingProgress = 0,
+		.lastSavingTimeMs = time(NULL),
 	};
 
-	if (!pg_copy(src, dst, &args))
+	if (!pg_copy(src, dst, &args, NULL, NULL))
 	{
 		/* errors have already been logged */
 		return false;
--- a/src/bin/pgcopydb/extensions.c
+++ b/src/bin/pgcopydb/extensions.c
@@ -174,7 +174,6 @@ copydb_copy_ext_table(PGSQL *src, PGSQL *dst, char *qname, char *condition)
 		.srcWhereClause = condition,
 		.dstQname = qname,
 		.dstAttrList = "",
-		.bytesTransmitted = 0,
 	};
 
 	if (!pg_copy(src, dst, &args, NULL, NULL))
--- a/src/bin/pgcopydb/extensions.c
+++ b/src/bin/pgcopydb/extensions.c
@@ -175,8 +175,6 @@ copydb_copy_ext_table(PGSQL *src, PGSQL *dst, char *qname, char *condition)
 		.dstQname = qname,
 		.dstAttrList = "",
 		.bytesTransmitted = 0,
-		.bytesTransmittedBeforeSavingProgress = 0,
-		.lastSavingTimeMs = time(NULL),
 	};
 
 	if (!pg_copy(src, dst, &args, NULL, NULL))
--- a/src/bin/pgcopydb/pgsql.c
+++ b/src/bin/pgcopydb/pgsql.c
@@ -59,8 +59,9 @@ static bool build_parameters_list(PQExpBuffer buffer,
 static void parseIdentifySystemResult(void *ctx, PGresult *result);
 static void parseTimelineHistoryResult(void *ctx, PGresult *result);
 
-static bool pg_copy_data(PGSQL *src, PGSQL *dst, CopyArgs *args, CopyProgressCallback
-						 onProgress, void *context);
+static bool pg_copy_data(PGSQL *src, PGSQL *dst, CopyArgs *args, void *context,
+						 CopyProgressCallback
+						 onProgress);
 
 static bool pg_copy_send_query(PGSQL *pgsql, CopyArgs *args,
 							   ExecStatusType status);
@@ -2663,8 +2664,8 @@ pgsql_truncate(PGSQL *pgsql, const char *qname)
  * referenced by the qualified identifier name dstQname on the target.
  */
 bool
-pg_copy(PGSQL *src, PGSQL *dst, CopyArgs *args, CopyProgressCallback
-		on_copy_progress_hook, void *context)
+pg_copy(PGSQL *src, PGSQL *dst, CopyArgs *args, void *context, CopyProgressCallback
+		on_copy_progress_hook)
 {
 	bool srcConnIsOurs = src->connection == NULL;
 	if (!pgsql_open_connection(src))
@@ -2683,7 +2684,7 @@ pg_copy(PGSQL *src, PGSQL *dst, CopyArgs *args, CopyProgressCallback
 		return false;
 	}
 
-	bool result = pg_copy_data(src, dst, args, on_copy_progress_hook, context);
+	bool result = pg_copy_data(src, dst, args, context, on_copy_progress_hook);
 
 	if (srcConnIsOurs)
 	{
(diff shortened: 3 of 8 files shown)


src/bin/pgcopydb/copydb.h
```diff
@@ -436,15 +436,17 @@ bool copydb_process_table_data_worker(CopyDataSpec *specs);
 bool copydb_process_table_data_with_workers(CopyDataSpec *specs);
 
 bool copydb_copy_table(CopyDataSpec *specs, PGSQL *src, PGSQL *dst,
-					   CopyTableDataSpec *tableSpecs);
+					   CopyTableDataSpec *tableSpecs,
+					   CopyProgressCallback onCopyProgress,
+					   void *onCopyProgressContext);
```
dimitri, line 441: Nitpicking: all the other callback signatures have the context first, the callback function next. What about doing the same here?

src/bin/pgcopydb/extensions.c
```diff
@@ -173,10 +174,12 @@ copydb_copy_ext_table(PGSQL *src, PGSQL *dst, char *qname, char *condition)
 		.srcWhereClause = condition,
 		.dstQname = qname,
 		.dstAttrList = "",
-		.bytesTransmitted = 0
+		.bytesTransmitted = 0,
+		.bytesTransmittedBeforeSavingProgress = 0,
+		.lastSavingTimeMs = time(NULL),
```
dimitri, line 179: This looks like it should be maintained in the callback's context, not here.

src/bin/pgcopydb/pgsql.c
```diff
@@ -2745,6 +2748,8 @@ pg_copy_data(PGSQL *src, PGSQL *dst, CopyArgs *args)
 	bool failedOnSrc = false;
 	bool failedOnDst = false;
 	args->bytesTransmitted = 0;
+	args->bytesTransmittedBeforeSavingProgress = 0;
+	args->lastSavingTimeMs = time(NULL);
```
dimitri, line 2752: That looks like it should be handled in the context for the CopyProgressCallback function.

src/bin/pgcopydb/pgsql.c
```diff
@@ -2849,9 +2854,9 @@ pg_copy_data(PGSQL *src, PGSQL *dst, CopyArgs *args)
 		/*
 		 * If successful PQgetCopyData returns the row length as a result.
 		 */
-		else if (bufsize > 0)
+		else if (bufsize > 0 && on_copy_progress_hook != NULL)
 		{
-			args->bytesTransmitted += bufsize;
+			on_copy_progress_hook(bufsize, context);
```
dimitri, line 2859: The callback function prototype returns a boolean value to indicate success or error. We need to process that return value here. Then there is a choice of logging a WARN/DEBUG message or an ERROR message (plus stopping the COPY operation) in case the callback returns false, but that needs to be explicit.

src/bin/pgcopydb/pgsql.h
```diff
@@ -337,9 +337,16 @@ typedef struct CopyArgs
 	bool truncate;
 	bool freeze;
 	uint64_t bytesTransmitted;
+	uint64_t bytesTransmittedBeforeSavingProgress; /* the bytes transmitted before saving the progress to DB */
+	uint64_t lastSavingTimeMs; /* the last saving time of the progress to the DB */
```
dimitri, line 341: That's a callback responsibility, please consider moving that away from here and within the `CopyProgressContext` structure, where it belongs.

src/bin/pgcopydb/summary.c
```diff
@@ -1866,12 +1866,6 @@ summary_stop_timing(DatabaseCatalog *catalog, TimingSection section)
 			return false;
 		}
 
-		if (!summary_pretty_print_timing(catalog, timing))
-		{
-			/* errors have already been logged */
-			return false;
-		}
-
```
dimitri, line 1874: How does that work now with respect to the `duration_pretty` column that still exists in the SQLite catalogs?

src/bin/pgcopydb/summary.c
```diff
@@ -2347,21 +2322,29 @@ catalog_timing_fetch(SQLiteQuery *query)
 	timing->doneTime = sqlite3_column_int64(query->ppStmt, 3);
 	timing->durationMs = sqlite3_column_int64(query->ppStmt, 4);
 
-	if (sqlite3_column_type(query->ppStmt, 5) != SQLITE_NULL)
+	/*
+	 * Skip reading `ppDuration` from DB and create it from `durationMs`
+	 * in accordance with the single source of truth principle.
+	 */
```
dimitri, line 2328: This principle was already implemented before, where the decision was made to compute the data before SQLite storage and then use what is in the SQLite database. This allows any external tooling to have access to the same data as our internal code. Including commands that run alongside the main `pgcopydb` process rather than as a child process.

IOW I believe you introduced a regression here.

src/bin/pgcopydb/summary.c
```diff
@@ -2347,21 +2322,29 @@ catalog_timing_fetch(SQLiteQuery *query)
...
-	timing->count = sqlite3_column_int64(query->ppStmt, 6);
-	timing->bytes = sqlite3_column_int64(query->ppStmt, 7);
+	timing->count = sqlite3_column_int64(query->ppStmt, 5);
+	timing->bytes = sqlite3_column_int64(query->ppStmt, 6);
 
-	if (sqlite3_column_type(query->ppStmt, 8) != SQLITE_NULL)
+	/*
+	 * Skip reading `ppBytes` from DB and create it from `bytes`
+	 * in accordance with the single source of truth principle.
+	 */
```
dimitri, line 2342: Same as before. What about external processes that might want to access to the telemetry?

src/bin/pgcopydb/table-data.c
```diff
@@ -1196,12 +1250,11 @@ copydb_table_create_lockfile(CopyDataSpec *specs,
...
- * semaphore first. The lockFile is also removed here.
+ * copydb_save_copy_progress saves the duration and the number of
+ * bytes transmitted about the copy progress into the catalog.
  */
 bool
-copydb_mark_table_as_done(CopyDataSpec *specs,
+copydb_save_copy_progress(CopyDataSpec *specs,
 						  CopyTableDataSpec *tableSpecs)
 {
 	DatabaseCatalog *sourceDB = &(specs->catalogs.source);
```
dimitri, line 1260: I do not understand this change. The `summary_finish_table` function is responsible to maintain the `done_time_epoch` field in the SQLite catalogs, and the rest of the code uses that information to determine if the table COPY is finished or not. Now we will update that `done_time_epoch` field multiple times during the COPY. What am I missing?

dimitri (changes requested):
This PR misses the mark on applying the “Separation of Concerns” principle, as detailed in the review. The callback context should handle its own specific data in a way that the CopyArgs does not have to know about. 

Also the PR comment invokes the “Single Source of Truth” principle, but I read the code as a violation of that principle, wherein it was previously made in a way that the single source of truth would be the SQLite catalogs, so that external processes can have a look at the same information as the internal code. Maybe the code evolved in a way that makes this not true anymore, but I think I would like to get back in that direction. For instance, we could build a `pgcopydb watch` command that acts like a `htop` command for pgcopydb: this would be an external process that would ideally just use the SQLite contents.

src/bin/pgcopydb/summary.c
rimbi: Looking at the existing sources, there seems to be ne need to keep the field in the table.

src/bin/pgcopydb/pgsql.c
rimbi: done

src/bin/pgcopydb/summary.c
rimbi: The problem was that the data in these 2 fields were inconsistent. What are the external tools which use these fields?

src/bin/pgcopydb/extensions.c
rimbi: Done. what about `bytesTransmitted`? It does not seem to belong here either.

src/bin/pgcopydb/summary.c
dimitri: - Is there a way to fix the bug wrt inconsistencies here, and keeping the data all in the SQLite database file? 
- Could the inconsistency be an explanation for https://github.com/dimitri/pgcopydb/issues/697?
- I mean “external” as in “not running within the process-tree with access to all the memory etc”, so that means the command `pgcopydb list progress` is already an “external” tool. All it has access to is the SQLite database file contents.

src/bin/pgcopydb/extensions.c
dimitri: You're right. I believe it was a trade-off that made sense with the previous code. Now that we have a separate context to handle this dynamic aspect, it's time to do the proper refactoring here.

src/bin/pgcopydb/table-data.c
rimbi: pushed a change for this.

src/bin/pgcopydb/summary.c
rimbi: - Probably there is. But is it a good idea to keep the 2 versions of the same data in the DB?
- I don't know, maybe.
- If we are talking about the `pgcopydb` then that means we have no problem here.

rimbi:
> This PR misses the mark on applying the “Separation of Concerns” principle, as detailed in the review. The callback context should handle its own specific data in a way that the CopyArgs does not have to know about.
> 
> Also the PR comment invokes the “Single Source of Truth” principle, but I read the code as a violation of that principle, wherein it was previously made in a way that the single source of truth would be the SQLite catalogs, so that external processes can have a look at the same information as the internal code. Maybe the code evolved in a way that makes this not true anymore, but I think I would like to get back in that direction. For instance, we could build a `pgcopydb watch` command that acts like a `htop` command for pgcopydb: this would be an external process that would ideally just use the SQLite contents.

This explanation sounds misleading. If we have the 2 versions of the same data present, that is the symptom of the violation of single source of truth. It caused problems and I spent a lot of time to find this problem. Any application (not only pgcopydb) can make use of the data in the db represented as integer and display it however it wants.

dimitri:
> This explanation sounds misleading. If we have the 2 versions of the same data present, that is the symptom of the violation of single source of truth. It caused problems and I spent a lot of time to find this problem. Any application (not only pgcopydb) can make use of the data in the db represented as integer and display it however it wants.

I think we're just using a different source of truth and then the same argument. Now that you know where the problem comes from, do you see any difficulty or architectural problem in using our SQLite database file as the source of truth?

rimbi:
> > This explanation sounds misleading. If we have the 2 versions of the same data present, that is the symptom of the violation of single source of truth. It caused problems and I spent a lot of time to find this problem. Any application (not only pgcopydb) can make use of the data in the db represented as integer and display it however it wants.
> 
> I think we're just using a different source of truth and then the same argument. Now that you know where the problem comes from, do you see any difficulty or architectural problem in using our SQLite database file as the source of truth?

Yes, there seems to be a misunderstanding. I'm not objecting to using the sqlite as the source of truth, but to using 2 different fields for the same data but with different representations (int and string).

src/bin/pgcopydb/summary.c
dimitri: Now that we use our own version of SQLite and we know we have JSON support in SQL, we could rewrite all the SQLite to JSON transformation code and use a SQL query instead. It was my impression that having the data in two forms in the SQLite database was useful for that.

Another use-case is debugging from the SQLite data file when something strange happens. Then we might appreciate having the data available in pretty printed format for debugging.

The use-case I mentioned before still do exists, it's nice to be able to re-use the data from the SQLite data file either in numeric format or in pretty-printed format without having to re-compute the pretty-printed format every time. It's always the same with caching: do you want to use more CPU or more memory?

src/bin/pgcopydb/table-data.c
```diff
@@ -836,6 +840,57 @@ copydb_table_data_worker(CopyDataSpec *specs)
...
+ */
+static bool
+on_copy_progress_hook(int bytesTransmitted, void *context)
+{
+	CopyProgressContext *ctx = (CopyProgressContext *) context;
+	CopyDataSpec *specs = ctx->specs;
+	CopyTableDataSpec *tableSpecs = ctx->tableSpecs;
+
+	ctx->bytesTransmittedBeforeSavingProgress += bytesTransmitted;
+	ctx->bytesTransmitted += bytesTransmitted;
```
dimitri, line 864: When are the two values different? There is no comment about that, and the naming of the variable does not explain that either. Why do we need the two slots? Can we have a better name for `bytesTransmittedBeforeSavingProgress`? I currently have two problems with it: I don't understand the name in relation with the code, and I would like a shorted one if that's possible.

src/bin/pgcopydb/table-data.c
```diff
@@ -1196,27 +1268,28 @@ copydb_table_create_lockfile(CopyDataSpec *specs,
...
+						  uint64_t bytes,
+						  instr_time lastSavingTimeMs)
 {
 	DatabaseCatalog *sourceDB = &(specs->catalogs.source);
 
-	if (!summary_finish_table(sourceDB, tableSpecs))
-	{
-		/* errors have already been logged */
-		return false;
-	}
```
dimitri, line 1213: I still expected to see a new function for handling updates during the COPY run, and to keep the existing function for marking the operation as done by updating the `done_time_epoch`. As I don't see any changes to the `done_time_epoch` handling (other than lacking an update), can you explain what's happening here?

dimitri: Hint: see `summary_table_count_parts_done()`, or `copydb_objectid_has_been_processed_already`, or `copydb_objectid_has_been_processed_already()`, where we use the `done_time_epoch` field.

dimitri:
> Yes, there seems to be a misunderstanding. I'm not objecting to using the sqlite as the source of truth, but to using 2 different fields for the same data but with different representations (int and string).

I think that's useful (see other comments). I would like to keep that.

src/bin/pgcopydb/table-data.c
rimbi: Basically, we don't call `summary_finish_table` during the copy operation, just at the end as before.

src/bin/pgcopydb/table-data.c
rimbi: Just because you don't understand it doesn't mean others don't, right? So the variable keeps the number of bytes transmitted before saving the progress to the db and reset to zero on each save. If you have a better name, please feel free to suggest one.
Assuming that you mean the vars starting with `bytesTransmitted...` one keeps the total as before (no change here) and the other as explained above.

dimitri: Is that attitude necessary @rimbi ? I can tell you it's not welcome.

Two things for you to consider:
  1. I am asking questions on aspects of your code that is not obvious to me when reviewing the PR. When things are not obvious, we need to either improve the code, or add good comments to help the reader understand what's happening with less efforts.
  2. I have read a lot of code, including some great code in many different programming languages. The great code that I have read always have these two properties: what/how it does is obvious, and it is elegant. Elegance has something personal to it, so I will merge code that I do not find elegant. I will not merge code that I do not find obvious, though. That's not personal.

src/bin/pgcopydb/table-data.c
```diff
@@ -945,16 +998,33 @@ copydb_copy_data_by_oid(CopyDataSpec *specs, PGSQL *src, PGSQL *dst,
 		/*
 		 * 1. Now COPY the TABLE DATA from the source to the destination.
 		 */
+		CopyProgressContext context = {
+			.specs = specs,
+			.tableSpecs = tableSpecs,
+			.bytesTransmitted = 0,
+			.lastSavingTimeMs = 0
+		};
+		INSTR_TIME_SET_CURRENT(context.lastSavingTimeMs);
```
dimitri, line 1007: Can we deal with that responsibility from within the hook function, allowing/expecting the context value to be set to zero by the caller?

dimitri (changes requested):
I think there is still a minor update to make to distribute responsibilities more cleanly (see new comment), and a major blocker to resolve.

Your stating that the human readable pretty-printed versions of the duration and bytes should not be maintained anymore, which I disagree with. But even if we agreed, the proper way to change that would be do open a separate PR and properly remove the two fields from the SQLite database and all the code that uses them, not just the code that maintains the values at UPDATE time!

In the context of this PR, please refrain from making a controversial half-baked change.

src/bin/pgcopydb/table-data.c
rimbi: Just to understand your reasoning: This pattern is quite common in the codebase. What is it that sets this piece of code apart form others?

dimitri: Which pattern do you mean? I can't remember of another place where we use INSTR TIME portability macros for high-precision time measurement from libpq and also a callback function and a context...

Please use the INSTR TIME macros only in the callback function, so that the caller does not need to know why your callback implementation decided to use INSTR TIME macros instead of `time(NULL)` for a 5 seconds duration where accuracy is not a concern, allowing us to change our mind later about how the callback function is implemented without having to also change the code from the caller that has to prepare the context. 

A good init value would be zero, because in C `name = { .specs = specs }` will leave the other fields properly initialized to zero. It is very easy in the callback function to initialize the lastSavingTimeMs to _now_ when called the first time.

dimitri:
Closing for triage purposes. A part of this PR has been superseded in #858, other aspects need to be addressed in their own PR.

PR #842, opened 2024-07-13
Extension config tables are not copied in dependency order

arajkumar:
We ensure that extension config tables are queried from source tables based on their dependency order in
https://github.com/dimitri/pgcopydb/pull/702/files, however while reading from sqlite catalog inorder to copy, we query config tables order by ext config oid which is not correct.

Solution: Read extension config table based on the configuration array index


dimitri:
Closed by #841

Issue #834, opened 2024-07-03
Parsing of the history file contents needs dynamic memory management


hariramt:
One or more errors occurred while migrating 'cloud_volumes_telemetry' : pgcopydb failed with unknown errors. Please contact Microsoft support. Logs from the migration:

```
 2024-07-03 03:36:55.649 76 **ERROR pgsql.c:3553 history file "00000307.history" contains 1547 lines, pgcopydb only supports up to 1023 lines**
2024-07-03 03:36:55.739 75 ERROR follow.c:1058 Subprocess prefetch with pid 76 has exited with error code 12
2024-07-03 03:36:55.744 77 ERROR ld_stream.c:2781 Failed to read stream context file: retry policy expired 
2024-07-03 03:36:56.053 75 ERROR follow.c:655 Some sub-process exited with errors, see above for details 
2024-07-03 03:36:56.053 75 ERROR follow.c:294 Failed to follow changes from source, see above for details 
2024-07-03 03:37:12.599 51 ERROR cli_clone_follow.c:763 follow process 75 has terminated [12]
One or more errors occurred while migrating 'metrics_collector' : pgcopydb failed with unknown errors. Please contact Microsoft support.
```

Logs from the migration: 

```2024-07-03 03:36:55.571 71 ERROR pgsql.c:3553 history file "00000307.history" contains 1547 lines, pgcopydb only supports up to 1023 lines 
2024-07-03 03:36:55.666 70 ERROR follow.c:1058 Subprocess prefetch with pid 71 has exited with error code 12 
2024-07-03 03:36:55.671 72 ERROR ld_stream.c:2781 Failed to read stream context file: retry policy expired 
2024-07-03 03:36:55.983 70 ERROR follow.c:655 Some sub-process exited with errors, see above for details 
2024-07-03 03:36:55.983 70 ERROR follow.c:294 Failed to follow changes from source, see above for details 
2024-07-03 03:37:03.492 32 ERROR cli_clone_follow.c:763 follow process 70 has terminated [12]
```

EDIT NOTES: please use proper formatting and split lines. Also, is it on purpose that you submitted the same logs twice?

sharonl1970:
[00000307.zip](https://github.com/user-attachments/files/16084985/00000307.zip)


File uploaded

sharonl1970:
This is the history file that is compressed.

kaman05010:
Do we have an update on the GitHub bug?

hanefi:
Completed in #848

PR #848, opened 2024-07-19
Store timelines in internal catalogs

hanefi:
This is a WIP PR that still has some rough edges.

TODO:
- [x] It does not look acceptable to include "schema.h" or "catalog.h" in file `pgsql.c`. I'll try to find a workaround
	- I ended up creating a new module. 
- [x] Having to pass the internal database details in a context does not look elegant.
	- I guess that is a compromise that is necessary.
- [x] Missing some catalog functions that can be used to populate data structures on current timeline. I do not see the value in storing a value that is not used anywhere.
	- I store the current timeline details in memory now. I believe having only the current timeline in memory is enough for today.


dimitri:
> This is a WIP PR that still has some rough edges.
> 
> TODO:
> 
>     * [ ]  It does not look acceptable to include "schema.h" or "catalog.h" in file `pgsaql.c`. I'll try to find a workaround

Usually that involves creating a new C module that will include all the needed headers. Here, that would be something like a “pgsql_timeline.c” module I suppose.

>     * [ ]  Missing some catalog functions that can be used to populate data structures on current timeline. I do not see the value in storing a value that is not used anywhere.

At the moment we fetch the timeline mostly for DEBUG purposes. Supporting a TLI change in Logical Decoding is still an open topic in pgcopydb. We need to have enough information to decide if we can follow a TLI change, depending on what we replayed already in the past etc. So for the moment, we just store the information in a way that's easy to review/debug, but we not use the information anywhere yet.

dimitri:
See #834

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,363 @@
...
+	{
+		log_error("Received a timeline history file of %zu bytes, "
+				  "pgcopydb is limited to files of up to %zu bytes.",
+				  strlen(value),
+				  sizeof(context->content));
+		context->parsedOk = false;
+	}
+	strlcpy(context->content, value, sizeof(context->content));
+
+	context->parsedOk = true;
```
dimitri, line 252: Can we fetch the value length and dynamically allocate a buffer of the proper size here, using `calloc`?

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,363 @@
...
+	entry->begin = prevend;
+	entry->end = InvalidXLogRecPtr;
+
+	log_trace("parseTimelineHistory[%d]: tli %d [%X/%X %X/%X]",
+			  timelineCount,
+			  entry->tli,
+			  LSN_FORMAT_ARGS(entry->begin),
+			  LSN_FORMAT_ARGS(entry->end));
+
+	catalog_add_timeline_history(context, entry);
```
dimitri, line 360: We need to either add `(void) ` as a prefix to mean we don't care about the result, or rather properly handle the boolean value returned and error out when it's not true.

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,363 @@
...
+	IdentifySystem *system;
+} IdentifySystemResult;
+
+
+typedef struct TimelineHistoryResult
+{
+	char sqlstate[6];
+	bool parsedOk;
+	char filename[MAXPGPATH];
+	char content[BUFSIZE * BUFSIZE]; /* 1MB should get us quite very far */
```
dimitri, line 30: That should now be a dynamic memory allocation, otherwise I don't think this PR will solve the bug it's been assigned to fix?

src/bin/pgcopydb/catalog.c
```diff
@@ -7700,6 +7707,176 @@ catalog_count_summary_done_fetch(SQLiteQuery *query)
 }
 
 
+/*
+ * catalog_add_timeline_history is a wrapper for
+ * catalog_add_timeline_history_internal so that callers can be agnostic of the
+ * TimelineHistoryContext.
```
dimitri, line 7713: I'm not sure I see the point about that yet. Will get back to it in another round of review I suppose.

dimitri (changes requested):
Looks good! I think there is a major aspect that needs a revisit, about the allocation strategy for the timeline history bits. We receive them in a single chunk of memory, but having a static size won't be good enough.

Other than than, the usual amount of nitpicking.

src/bin/pgcopydb/pgsql_timeline.c
dimitri: Apparently we can use (PQgetlength)[https://www.postgresql.org/docs/current/libpq-exec.html#LIBPQ-PQGETLENGTH] for that.

src/bin/pgcopydb/catalog.c
hanefi: After introducing a new module, this dirty workaround is not really necessary. I merged the two functions into one.

src/bin/pgcopydb/catalog.c
```diff
@@ -7700,6 +7707,176 @@ catalog_count_summary_done_fetch(SQLiteQuery *query)
...
+	SQLiteQuery query = {
+		.context = entry,
+		.fetchFunction = &catalog_timeline_history_fetch
+	};
+
+	char *sql =
+		"  select tli, startpos, endpos"
+		"    from timeline_history"
+		"   where tli = $1"
+		"   limit 1";
```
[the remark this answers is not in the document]
hanefi: Makes sense. Removed.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: Fair point. This piece of memory needs to be handled dynamically.

The reported bug was about the limitation on the number of lines (PGCOPYDB_MAX_TIMELINES). This line is about the total size of the timeline file (PGCOPYDB_MAX_TIMELINE_CONTENT).

On that note, I will remove both of these constants in a following commit as they are not used anywhere in the code.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: Fixed.

This is not the first time I missed checking return values on a function. I wonder I should use `-Wunused-result` or set up a linter locally.

src/bin/pgcopydb/catalog.c
```diff
@@ -7700,6 +7704,169 @@ catalog_count_summary_done_fetch(SQLiteQuery *query)
...
 
+/*
+ * catalog_add_timeline_history inserts a timeline history entry to our
+ * internal catalogs database.
+ */
+bool
+catalog_add_timeline_history(void *ctx, TimelineHistoryEntry *entry)
+{
+	TimelineHistoryContext *context = (TimelineHistoryContext *) ctx;
+	DatabaseCatalog *catalog = context->source;
```
dimitri, line 7715: Can we pass just the `DatabaseCatalog *catalog` as the first argument to this function?

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,371 @@
...
+			return false;
+		}
+
+		TimelineHistoryEntry *current = &(system->currentTimeline);
+
+		log_sql("TIMELINE_HISTORY: \"%s\", timeline %d started at %X/%X",
+				hContext.filename,
+				current->tli,
+				(uint32_t) (current->begin >> 32),
+				(uint32_t) current->begin);
```
dimitri, line 131: Let's use LSN_FORMAT_ARGS here?

src/bin/pgcopydb/schema.h
```diff
@@ -389,6 +389,11 @@ typedef struct DatabaseCatalog
 	Semaphore sema;
 } DatabaseCatalog;
 
+typedef struct TimelineHistoryContext
+{
+	DatabaseCatalog *source;
+} TimelineHistoryContext;
```
dimitri, line 395: I wonder why we don't just pass a `DatabaseCatalog *` argument around?

src/bin/pgcopydb/pgsql_timeline.c
dimitri: Or just `strdup` ;-)

src/bin/pgcopydb/schema.h
hanefi: Fair comment.

This struct used to have some more attributes that got removed. At some point I used to store the current timeline identifier for lookups in this context so that I can fetch the correct record from internal catalog. It no longer makes sense to introduce a new struct here at all.

src/bin/pgcopydb/pgsql_timeline.c
dimitri: Yeah I realize it was done that way before your PR. You just happen to have a good opportunity to fix this in passing, thanks for doing that!

src/bin/pgcopydb/catalog.c
```diff
@@ -7700,6 +7704,169 @@ catalog_count_summary_done_fetch(SQLiteQuery *query)
...
 
 
+/*
+ * catalog_add_timeline_history inserts a timeline history entry to our
+ * internal catalogs database.
+ */
+bool
+catalog_add_timeline_history(void *ctx, TimelineHistoryEntry *entry)
+{
+	DatabaseCatalog *catalog = (DatabaseCatalog *) ctx;
```
dimitri, line 7714: Why use a `void *` context that actually is a `DatabaseCatalog *` argument? It would be better to skip the cast and have the compiler validation for us... is there a reason to do this that I can't see and that is better than having the compiler checks?

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,370 @@
...
+		context->parsedOk = false;
+		return;
+	}
+
+	/* filename (text) */
+	char *value = PQgetvalue(result, 0, 0);
+	strlcpy(context->filename, value, sizeof(context->filename));
+
+	/* content (bytea) */
+	value = PQgetvalue(result, 0, 1);
```
dimitri, line 239: Is there a point to be made that we should maybe store the timeline history to a local file with the Postgres filename and the content received, no edits, for more debugging abilities? If not, then let's consider adding a comment about it here.

src/bin/pgcopydb/pgsql_utils.h
```diff
@@ -0,0 +1,9 @@
+/*
+ * src/bin/pgcopydb/pgsql_utils.h
+ *     Common helper functions for interacting with a postgres server
+ */
+
+#include "libpq-fe.h"
+
+bool is_response_ok(PGresult *result);
+bool clear_results(PGSQL *pgsql);
```
dimitri, line 9: Maybe add a comment line that says where the functions are implemented? I know modern IDEs allow to easily follow that, but I like to help human code readers first.

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,370 @@
...
+		}
+
+		if (*ptr == '\0' || *ptr == '#')
+		{
+			continue;
+		}
+
+		log_trace("parseTimelineHistory line %lld is \"%s\"",
+				  (long long) lineNumber,
+				  lbuf.lines[lineNumber]);
```
dimitri, line 295: If we get to store the history file as-is on-disk then we should be able to remove that line entirely.

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,370 @@
...
+}
+
+
+/*
+ * parseTimelineHistory parses the content of a timeline history file.
+ */
+bool
+parseTimelineHistory(const char *filename, const char *content,
+					 IdentifySystem *system, void *context)
+{
```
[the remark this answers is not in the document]
dimitri: Also is the filename necessary here at all?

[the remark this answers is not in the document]
hanefi: I also think that it makes sense to use `DatabaseCatalog *` here. However, it is not trivial to solve the circular dependency. We have similar issues in other occurences of void pointers in this branch. Sharing an example:

- `parseTimelineHistory` declaration is at `pgsql.h`, implementation is at `pgsql_timeline.c`
- `DatabaseCatalog` definition is in `schema.h`
- `schema.h` includes `pgsql.h`
- If I want to add an include directive for `schema.h` in `pgsql.h`, we got a circular dependency.

Shall I move things to a separate module and try to come up with a better hierarchy?

```mermaid
	graph TD;
parseTimelineHistory --> pgsql.h
parseTimelineHistory --> pgsql_timeline.c
DatabaseCatalog --> schema.h
schema.h --> pgsql.h
pgsql.h ==does not work==> schema.h
```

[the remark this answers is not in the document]
hanefi: I can also use forward declarations for problematic structs. I never liked them, and I avoid them usually, but there may be no other option.

I guess there is no other way around it. I'll add the following to `pgsql.h`:

```c
/* forward declaration */
typedef struct DatabaseCatalog DatabaseCatalog;
```

src/bin/pgcopydb/catalog.c
hanefi: I could not figure out the circular dependency in our headers. In the end I ended up writing some forward declaration for `DatabaseCatalog` to circumvent that.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: I added the logic to write the contents to disk earlier in the stack trace. See relevant discussion at https://github.com/dimitri/pgcopydb/pull/848#discussion_r1688624052

src/bin/pgcopydb/catalog.c
dimitri: We might need a new header `pgsql_timeline.h` that could use both the catalogs and the lower-level pgsql header?

src/bin/pgcopydb/pgsql_timeline.c
dimitri: Yeah that's why the placement of my comment about storing to disk was at this seemingly random place. I would like to store the file exactly as Postgres makes it available to us, only for debugging purposes, in case our parsing of the file were to fail.

src/bin/pgcopydb/pgsql_timeline.c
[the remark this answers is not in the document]
dimitri: See my other comment about `pgsql_timeline.h` maybe?

src/bin/pgcopydb/pgsql_timeline.c
dimitri: I see that now, looks like what I had in mind, thanks!

src/bin/pgcopydb/pgsql_timeline.c
dimitri: I see now what you've done, and I think that's good. When implementing debug traces that are not used in the normal flow of operations, or even in the code, I don't think separation of concerns is as imperative. Here, we want to log-to-disk the actual result from the Postgres protocol, before any kind of processing...

src/bin/pgcopydb/catalog.c
hanefi: I created a new header in a new commit, and removed the forward declaration for `DatabaseCatalog`

It contains the declaration of a function defined in `psql.c` just because I can not include the file that defines the struct `DatabaseCatalog` in `psql.h`.

```c
/* pgsql.c */
bool pgsql_start_replication(LogicalStreamClient *client, DatabaseCatalog *catalog,
							 char *cdcPathDir);
```

src/bin/pgcopydb/pgsql_timeline.c
hanefi: After reading the following sentence in the last comment, I realized that I did not print a debug message when writing the tli history file.

> When implementing debug traces that are not used in the normal flow of operations, or even in the code, I don't think separation of concerns is as imperative.

I pushed a new commit 2eacb24 that adds a debug message similar to what we had prior to this PR.

dimitri: Only thinking about it now, sorry: what could make sense is to store the timeline data received from the protocol directly to disk, no parsing, and then have a function that reads the file one line at a time and parses the content etc.

This would be better at memory management, and also separation of concerns and internal API/headers.

What do you think?

hanefi: I actually write contents to disk before parsing it.

Currently `pgsql_identify_system` function does the following calls:

- parseTimelineHistoryResult parses PGResult into a TimelineHistoryResult
- writeTimelineHistoryFile writes the TimelineHistoryResult content that holds raw data to disk
- parseTimelineHistoryfile parses the TimelineHistoryResult and stores it in internal catalog

src/bin/pgcopydb/catalog.c
```diff
@@ -7700,6 +7704,167 @@ catalog_count_summary_done_fetch(SQLiteQuery *query)
...
+		/* errors have already been logged */
+		return false;
+	}
+
+	return true;
+}
+
+
+bool
+catalog_timeline_history_fetch(SQLiteQuery *query)
```
dimitri, line 7831: Missing function level comment.

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,396 @@
...
+		if (!writeTimelineHistoryFile(hContext.filename, hContext.content, cdcPathDir))
+		{
+			log_error("Failed to write contents of TIMELINE_HISTORY command to disk");
+			PQfinish(connection);
+			return false;
+		}
+
+		log_debug("Wrote tli %s timeline history file \"%s/%s\"",
+				  system->currentTimeline.tli, cdcPathDir, hContext.filename);
+
```
dimitri, line 130: This should be done in `writeTimelineHistoryFile` and *before* the call to `write_file`, so that if the writing to file happens fail or take a long time, we already would have printed the log information about it.

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,396 @@
...
+	}
+
+	/* filename (text) */
+	char *value = PQgetvalue(result, 0, 0);
+	strlcpy(context->filename, value, sizeof(context->filename));
+
+	/* content (bytea) */
+	value = PQgetvalue(result, 0, 1);
+	context->content = strdup(value);
+
```
dimitri, line 267: I was thinking we might want to write to file at this place in the code, then skipping the `strdup` entirely, and keeping only the filename in our internal structure.

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,396 @@
...
+					 DatabaseCatalog *catalog)
+{
+	LinesBuffer lbuf = { 0 };
+
+	if (!splitLines(&lbuf, (char *) content))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
```
dimitri, line 295: If we write the file to disk as soon as receiving it, and without duplicating the contents in-memory, then we don't need to `splitLines` here and instead of looping through the array lines we could read the file line-by-line using the iterator and a callback function.

src/bin/pgcopydb/ld_stream.c
```diff
@@ -490,7 +491,7 @@ startLogicalStreaming(StreamSpecs *specs)
 				  OutputPluginToString(specs->slot.plugin),
 				  specs->pluginOptions.count);
 
-		if (!pgsql_start_replication(&stream))
+		if (!pgsql_start_replication(&stream, specs->sourceDB, specs->paths.dir))
```
dimitri, line 494: The whole idea behing the latest opened discussion in this review is to allow to skip using the sourceDB when doing `pgsql_start_replication()`, getting back to your point about separation of concerns. At replication start, only store the timeline history to a file.

In `stream_write_context()` (or maybe in a new function, if needed?) we can then parse the history file with a file iterator and store our own format in the SQLite catalogs.

src/bin/pgcopydb/pgsql_timeline.c
dimitri: See the other comments I have added, with more details. I think we have an opportunity to implement things in a way that `start_logical_decoding` does not need access to our SQLite catalogs.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: Ok. I'll move this before the call to `write_file` and change wording slightly (e.g. Wrote -> Writing).

In my defense, in `stream_write_context` we print the logs only after successfully writing the files to disk.

src/bin/pgcopydb/catalog.c
hanefi: Added them.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: Done. After implementing the iterator API from scratch, I needed to move the filename to another structure. However, I think it is fine as I did not duplicate the file content and store only the file name.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: I implemented the iterator API, and removed the `splitLines` here in favor of our file iterators.

If I can share some context, I did not really write this piece of code myself. I just moved it from one file to another as it made sense to group all timeline related functions to the same module. We had a function who was good at parsing the complete file by itself, and know we have `timeline_iter_history[|_init|_next|_finish]` functions that handle portions of it.

src/bin/pgcopydb/pgsql.c
```diff
@@ -4182,7 +3816,7 @@ pgsql_timestamptz_to_string(TimestampTz ts, char *str, size_t size)
  * Send the START_REPLICATION logical replication command.
  */
 bool
-pgsql_start_replication(LogicalStreamClient *client)
+pgsql_start_replication(LogicalStreamClient *client, char *cdcPathDir)
```
dimitri, line 3819: Can we add the cdcPathDir to the LogicalStreamClient structure?

src/bin/pgcopydb/pgsql_timeline.h
```diff
@@ -0,0 +1,46 @@
...
+						   char *cdcPathDir);
+bool timeline_iter_history(char *filename,
+						   ParseTimelineHistoryContext *context,
+						   TimelineHistoryFun *callback);
+
+bool timeline_iter_history_init(TimelineHistoryIterator *iter);
+bool timeline_iter_history_next(TimelineHistoryIterator *iter);
+bool timeline_iter_history_finish(TimelineHistoryIterator *iter);
+
+bool timeline_history_add_hook(void *context, TimelineHistoryEntry *entry);
```
dimitri, line 41: I'm not fond of the name. Not a showstopper. Will think. The idea is to describe what the hook does, not where we call it...

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,513 @@
...
+		(TimelineHistoryIterator *) malloc(sizeof(TimelineHistoryIterator));
+
+	if (iter == NULL)
+	{
+		log_error("Failed to allocate memory for timeline history iterator");
+		return false;
+	}
+
+	iter->filename = filename;
+	iter->currentTimeline = context->currentTimeline;
```
dimitri, line 278: The iterator should never look into the context.

dimitri (commented):
Quick review from the metro, hope it's useful. I am surprised we need to define our own iterator. Expected that a callback to the file line iterator would do.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: This is a great point. Applying a fix for this allowed me to make the following improvements:
- I passed the value as an argument, I had a slimmer context.
- Now that the context had a single attribute, I deleted that completely.
- A function with a bad name `timeline_history_add_hook` that was a wrapper for `catalog_add_timeline_history` was no longer needed as I did not need to have the logic to access a subset of values in a context.

When the number of parameters get large, I start to think of context structs as a means to group them. This was a wrong way to think, and I see that now.

src/bin/pgcopydb/pgsql_timeline.h
hanefi: I ended up removing this completely. Relevant discussion at other comment: https://github.com/dimitri/pgcopydb/pull/848#discussion_r1698482611

hanefi:
> I am surprised we need to define our own iterator. Expected that a callback to the file line iterator would do.

I had 2 main issues that were not easy to overcome if I used the current file line iterators, and `file_iter_lines` function.

Problem 1: 

When parsing a line, I needed to store some values in the iterator, and it was not possible with the current file iterator. I believe we need to be able to pass custom iterators for that to be useful.

I considered updating the current implementation of `file_iter_lines` to accept an iterator instead of some fixed set of arguments that are used to construct the iterator in the function body. That would have worked out for me.

Problem 2:

I need to be able to call the callback function after all the lines are processed. This way, we are able to add a final timeline entry for the current timeline. This was not easy to accomplish in the current implementation of `file_iter_lines`.

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,493 @@
...
+
+			PQfinish(connection);
+
+			return false;
+		}
+
+		(void) parseTimelineHistoryResult((void *) &hContext, result, cdcPathDir);
+
+		sformat(system->timelineHistoryFilename, MAXPGPATH,
+				"%s/%s", cdcPathDir, hContext.filename);
```
dimitri, line 111: The same `sformat` is done twice. Can we pass the system down to `parseTimelineHistoryResult` so that we build the timeline history absolute pathname only once?

src/bin/pgcopydb/pgsql_timeline.c
```diff
@@ -0,0 +1,493 @@
...
+
+/*
+ * writeTimelineHistoryFile writes the content of a timeline history file to disk.
+ * The filename is determined by the PostgreSQL TIMELINE_HISTORY command.
+ */
+static bool
+writeTimelineHistoryFile(char *filename, char *content, char *cdcPathDir)
+{
+	char path[MAXPGPATH] = { 0 };
+	sformat(path, MAXPGPATH, "%s/%s", cdcPathDir, filename);
```
dimitri, line 141: With the previous comment, this would take a single `filename` that would have already been computed to be in the right directory.

src/bin/pgcopydb/pgsql_timeline.c
hanefi: I agree that I should do this formatting only once. However, I'd rather not push `system` to `parseTimelineHistoryResult` as accessing that is not really necessary for parsing. Instead, I copy the string after this function completes.

dimitri (approved):

dimitri:
Thanks for pushing all the work Hanefi, I like what we ended-up with!

PR #849, opened 2024-07-22
Support timescaledb

Kesarwani-aditi:
This pull request adds support for the TimescaleDB extension in pgcopydb. The update includes specific pre and post pg_restore steps needed for TimescaleDB, requires data fetching and pg_dump operations. The changes ensure compatibility with TimescaleDB, enhancing the utility of pgcopydb for users leveraging TimescaleDB's capabilities.


src/bin/pgcopydb/catalog.c
```diff
@@ -6487,6 +6487,66 @@ catalog_add_s_extension_config(DatabaseCatalog *catalog,
...
 
 
+/*
+ *   catalog_iter_s_extesnion_timescaledb_checker , iterates over the list of extensions in our catalogs
+ *   and checks for presence of timescaledb extension.
+ */
+bool
+catalog_iter_s_extension_timescaledb_checker(DatabaseCatalog *catalog,
+											 bool *timescaledb)
+{
```
dimitri, line 6497: That looks like it should be a lookup function, not a iterator function that short-circuit the iteration when it finds a specific value?

dimitri (commented):
Thanks for working on timescaledb support! I think I remember that even when the PR used to work in the CI testing, it was failing for real users. Do you have a good testing setup outside of the CI?

cc @arajkumar

src/bin/pgcopydb/catalog.c
arajkumar: How about calling this as `catalog_has_timescaledb_extension`?

src/bin/pgcopydb/catalog.c
```diff
@@ -6487,6 +6487,66 @@ catalog_add_s_extension_config(DatabaseCatalog *catalog,
...
+			if (!catalog_iter_s_extension_finish(iter))
+			{
+				/* errors have already been logged */
+				return false;
+			}
+
+			break;
+		}
+
+		if (strcmp(ext->extname, "timescaledb") == 0)
```
arajkumar, line 6532:
```suggestion
		if (streq(ext->extname, "timescaledb"))
```

arajkumar (changes requested):

Kesarwani-aditi:
> Thanks for working on timescaledb support! I think I remember that even when the PR used to work in the CI testing, it was failing for real users. Do you have a good testing setup outside of the CI?
> 
> cc @arajkumar

Kesarwani-aditi:
Thank you for your feedback! I appreciate it. 
Yes, we have set up manual testing outside of the CI environment to ensure that everything works correctly for real users. This includes various scenarios and edge cases to verify robustness and reliability.
Please let me know if there are any specific areas you’d like us to focus on or if you have any additional concerns.

Kesarwani-aditi:
> Thanks for working on timescaledb support! I think I remember that even when the PR used to work in the CI testing, it was failing for real users. Do you have a good testing setup outside of the CI?
> 
> cc @arajkumar

Thank you for your feedback! I appreciate it. 
Yes, we have set up manual testing outside of the CI environment to ensure that everything works correctly for real users. This includes various scenarios and edge cases to verify robustness and reliability.
Please let me know if there are any specific areas you’d like us to focus on or if you have any additional concerns.

src/bin/pgcopydb/catalog.c
Kesarwani-aditi: > How about calling this as `catalog_has_timescaledb_extension`?
I’ve updated the name to catalog_has_timescaledb_extension as suggested.

Kesarwani-aditi:
This pull request adds support for the TimescaleDB extension in pgcopydb. The update includes specific pre and post pg_restore steps needed for TimescaleDB ,requires data fetching and pg_dump operations. The changes ensure compatibility with TimescaleDB, enhancing the utility of pgcopydb for users leveraging TimescaleDB's capabilities.

tests/timescaledb/copydb.sh
```diff
@@ -0,0 +1,56 @@
...
+# take a snapshot using role tsdb on source database
+coproc ( pgcopydb snapshot --debug )
+
+sleep 1
+
+# copy the extensions separately, needs superuser (both on source and target)
+pgcopydb list extensions
+
+# now clone with superuser privileges, seems to be required for timescaledb
+pgcopydb clone --skip-extensions \
```
arajkumar, line 44: I don't see anywhere the extension being copied., adding --skip-extension wouldn't migrate TimescaleDB catalog tables.

tests/timescaledb/copydb.sh
```diff
@@ -0,0 +1,56 @@
...
+copy="\COPY rides FROM nyc_data_rides.10k.csv CSV"
+psql -d ${PGCOPYDB_SOURCE_PGURI} -1 -c "${copy}"
+
+# take a snapshot using role tsdb on source database
+coproc ( pgcopydb snapshot --debug )
+
+sleep 1
+
+# copy the extensions separately, needs superuser (both on source and target)
+pgcopydb list extensions
```
arajkumar, line 41: This won't copy the extensions, rather it would just list the extension on the console. Did you intent to use `pgcopydb copy extensions`?

tests/timescaledb/copydb.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# This script expects the following environment variables to be set:
+#
+#  - PGCOPYDB_SOURCE_PGURI
+#  - PGCOPYDB_TARGET_PGURI
+#  - PGCOPYDB_TABLE_JOBS
+#  - PGCOPYDB_INDEX_JOBS
+
+# make sure source and target databases are ready
+pgcopydb ping --source ${PGCOPYDB_SOURCE_PGURI} --target ${PGCOPYDB_TARGET_PGURI}
```
arajkumar, line 14: I didn't understand why PGCOPYDB_SOURCE_PGURI and PGCOPYDB_TARGET_PGURI being used explicitly? Removing them should work fine.

tests/timescaledb/copydb.sh
```diff
@@ -0,0 +1,56 @@
...
+create database tsdb owner tsdb;
+EOF
+
+# copying roles needs superuser
+# and we use the postgres database here still
+pgcopydb copy roles --source ${PGCOPYDB_SOURCE_PGURI} --target ${PGCOPYDB_TARGET_PGURI}
+
+# now create the tsdb database on the target, owned by new role tsdb
+psql ${PGCOPYDB_TARGET_PGURI} <<EOF
+create database tsdb owner tsdb;
```
arajkumar, line 27: What is the purpose of this database? I don't see them being used.

arajkumar (changes requested):

Kesarwani-aditi:
Hi @arajkumar, 
I’ve encountered an issue with hypertables after migrating data. On the database where I originally created hypertables, executing the query "SELECT * FROM timescaledb_information.hypertables;" shows the list of hypertables correctly. However, when I run the same query on the database that contains the migrated data, the list does not include the migrated hypertable names.
Could you please help me with this?
Thanks

tests/timescaledb/copydb.sh
Kesarwani-aditi: Yes, I intended to use `pgcopydb copy extensions`. I have made the changes from `pgcopydb list extensions` to `pgcopydb copy extensions` to ensure that the extensions are properly copied from the source to the target. Thanks!

tests/timescaledb/copydb.sh
Kesarwani-aditi: Yes, I have removed the URLs.

tests/timescaledb/copydb.sh
Kesarwani-aditi: The purpose of the `tsdb` database was to simply check that the creation of the database is happening properly without affecting the data in the source and target. It had no effect on the migration, which is why I removed it.

tests/timescaledb/copydb.sh
```diff
@@ -0,0 +1,44 @@
...
+
+kill -TERM ${COPROC_PID}
+wait ${COPROC_PID}
+
+s=/tmp/fares.out
+t=/tmp/fares.out
+psql -o $s -d ${PGCOPYDB_SOURCE_PGURI} -f /usr/src/pgcopydb/fares.sql
+psql -o $t -d ${PGCOPYDB_TARGET_PGURI} -f /usr/src/pgcopydb/fares.sql
+
+diff $s $t
```
arajkumar, line 44: Hopefully, last few comments!

Could you please also check the rows of the below relations are same between source and target?

timescaledb_information.hypertables
timescaledb_information.chunks

tests/timescaledb/rides.sql
```diff
@@ -0,0 +1,59 @@
...
+);
+INSERT INTO rates(rate_code, description) VALUES
+(1, 'standard rate'),
+(2, 'JFK'),
+(3, 'Newark'),
+(4, 'Nassau or Westchester'),
+(5, 'negotiated fare'),
+(6, 'group ride');
+
+COMMIT;
```
arajkumar, line 59: Please add TODOs to add tests for other timescaledb features (like compression, continous aggregation, background user defined jobs, retention policy).

tests/timescaledb/README.md
```diff
@@ -0,0 +1,3 @@
+# Testing extensions
+
+* PostGIS test data from https://github.com/adityatoshniwal/postgis-sample-dataset
```
arajkumar, line 3: This is incorrect, either remove the file or better add the right information.

arajkumar (changes requested):

src/bin/pgcopydb/catalog.c
```diff
@@ -6487,6 +6487,66 @@ catalog_add_s_extension_config(DatabaseCatalog *catalog,
...
 
 
+/*
+ *   catalog_iter_s_extesnion_timescaledb_checker , iterates over the list of extensions in our catalogs
+ *   and checks for presence of timescaledb extension.
+ */
+bool
+catalog_has_timescaledb_extension(DatabaseCatalog *catalog,
+											 bool *timescaledb)
+{
```
dimitri, line 6497: The new name needs updating in the comment also. But the problem that I have with that code is not even with the function name. Why use an iterator over the full list of extensions in our SQLite catalogs when searching for the presence of only one, that we know the name of? This should be `SELECT 1 FROM s_extensions WHERE extname = $1`, like I said before: a **lookup** function.

dimitri (changes requested):

src/bin/pgcopydb/extensions.c
```diff
@@ -407,6 +407,39 @@ copydb_parse_extensions_requirements(CopyDataSpec *copySpecs, char *filename)
 }
 
 
+/* source_has_timescaledb_extension will check if timescaledb extension is present in 
+ * source database, if it exists then prepare the atrget database for timescale data
+ */
+bool 
+source_has_timescaledb_extension(CopyDataSpec *copySpecs, bool *hasTimescaledb)
```
arajkumar, line 414: I think @dimitri meant to query the sqlite catalog not the source again.

tests/timescaledb/copydb.sh
Kesarwani-aditi: Yes, I have checked and the rows of timescaledb_information.hypertables and timescaledb_information.chunks are same for between source and target database.

arajkumar: @Kesarwani-aditi Sorry for being unclear. I meant to add checks in the copydb.sh to verify those tables has same contents.

tests/timescaledb/rides.sql
Kesarwani-aditi: Yes, for sure! I'll add the TODOs to include tests for other TimescaleDB features.

tests/timescaledb/README.md
Kesarwani-aditi: Sorry, I didn't look carefully. I've removed the file now.

src/bin/pgcopydb/catalog.c
Kesarwani-aditi: We have created a lookup function as you asked. It now uses `SELECT 1 FROM s_extensions WHERE extname = $1` to efficiently check for the presence of an extension.

src/bin/pgcopydb/extensions.c
Kesarwani-aditi: Yes, I misunderstood it earlier. I have now implemented the lookup function to query the SQLite catalog instead of the source. Thank you for pointing that out!

tests/timescaledb/copydb.sh
arajkumar: Could you please add this?

Kesarwani-aditi:
@arajkumar 
Could you please review my code when you get a chance.

arajkumar:
> @arajkumar Could you please review my code when you get a chance.

I see the following comment is not addressed yet.
https://github.com/dimitri/pgcopydb/pull/849/files#r1693158692

src/bin/pgcopydb/extensions.c
```diff
@@ -405,3 +405,122 @@ copydb_parse_extensions_requirements(CopyDataSpec *copySpecs, char *filename)
...
+{
+	PGSQL dst = { 0 };
+
+	if (!pgsql_init(&dst, copySpecs->connStrings.target_pguri, PGSQL_CONN_TARGET))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	char *sql = "SELECT timescaledb_post_restore()";
```
arajkumar, line 517: This doesn't work when the timescaledb extension is installed on non "public" schema. You may have factoring the schema at which timescaledb is installed.

e.g.
```
create extension timescaledb with schema "foo" version "2.13.0";
```

VaibhaveS: done

tests/timescaledb/copydb.sh
VaibhaveS: Hey @arajkumar, do you think adding pgcopydb compare would do the job here?

```
pgcopydb compare schema
pgcopydb compare data
```

arajkumar: @VaibhaveS I'm not sure compare schema compares extension catalog table as well. It is better to explicitly check them.

src/bin/pgcopydb/extensions.c
```diff
@@ -407,3 +407,168 @@ copydb_parse_extensions_requirements(CopyDataSpec *copySpecs, char *filename)
...
+	if (!pgsql_init(&dst, copySpecs->connStrings.target_pguri, PGSQL_CONN_TARGET))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	PQExpBuffer sql = createPQExpBuffer();
+	char *sqlTemplate = "select %s.timescaledb_pre_restore()";
+
+	appendPQExpBuffer(sql, sqlTemplate, extension->extnamespace);
```
arajkumar, line 513: nit:

How about using a stack char array & sformat instead of createPQExpBuffer? IMHO, that would simplify this function a lot.

```suggestion
    char sql[128] = { 0 };
    char *sqlTemplate = "select %s.timescaledb_pre_restore()";
    sformat(sql, sizeof(sql), sqlTemplate, extension->extnamespace);
```

VaibhaveS: yeah makes sense

tests/timescaledb/copydb.sh
VaibhaveS: yeah, it checks only `_timescaledb_internal._hyper_1_2_chunk` and `_timescaledb_internal._hyper_1_1_chunk` in the tests

VaibhaveS: Added.

src/bin/pgcopydb/dump_restore.c
```diff
@@ -188,6 +188,16 @@ copydb_target_prepare_schema(CopyDataSpec *specs)
 		return false;
 	}
 
+	/*
```
arajkumar, line 191: I just found that the `copydb_prepare_extensions_restore` should be executed after the `pg_restore_db`. 

i.e. after the pre-data restore

All other changes looks good.

VaibhaveS: oh

VaibhaveS: done

arajkumar: The problem is the pre-data restore is executing `ALTER TABLE _timescaledb_internal._hyper_1_1_chunk OWNER TO postgres` on the hypertable chunk which is not supported.

Adding `--no-owner` to clone would help.

VaibhaveS: when the pre_restore() function is called before then this is not occurring...

VaibhaveS: how do we make this compatible with owners though?

arajkumar: >when the pre_restore() function is called before then this is not occurring...

If you want to reproduce the error, use `timescale/timescaledb:2.13.0-pg13` docker image from compose.yaml.

>how do we make this compatible with owners though?

I don't think that is possible as of now.

arajkumar (approved):

src/bin/pgcopydb/extensions.c
```diff
@@ -407,3 +407,146 @@ copydb_parse_extensions_requirements(CopyDataSpec *copySpecs, char *filename)
...
+{
+	PGSQL dst = { 0 };
+
+	if (!pgsql_init(&dst, copySpecs->connStrings.target_pguri, PGSQL_CONN_TARGET))
+	{
+		/* errors have already been logged */
+		return false;
+	}
+
+	char sql[128] = { 0 };
```
dimitri, line 510: Please use BUFSIZE here not 128. The schema name can be 64 chars, but when properly quoted in theory you need to double that.

src/bin/pgcopydb/extensions.c
```diff
@@ -407,3 +407,146 @@ copydb_parse_extensions_requirements(CopyDataSpec *copySpecs, char *filename)
...
```
dimitri, line 539: Same, BUFSIZE looks like a better choice than the constant 128.

dimitri (changes requested):
Nitpicking: some constant size for strings need an adjustment.

dimitri (approved):

arajkumar:
Nice work @VaibhaveS and @Kesarwani-aditi !

Thanks @dimitri for your early work on this.

PR #850, opened 2024-07-22
Fixes literalWithCharPtrCompare cppcheck warnings

rimbi:
The reason for the warning is the comparison of string literals like "ctid" with NULL, which appeared at 23 points as follows:
```bash
src/bin/pgcopydb/cli_list.c:1385:6: warning: String literal compared with variable 'NULL'. Did you intend to use strcmp() instead? [literalWithCharPtrCompare]
 if (streq(table->partKey, "ctid"))
     ^
```

### DIFF
--- a/src/bin/pgcopydb/string_utils.h
+++ b/src/bin/pgcopydb/string_utils.h
@@ -11,9 +11,11 @@
 
 #define NULL_AS_EMPTY_STRING(str) (str == NULL ? "" : str)
 
+/* casting to const char * is necessary to avoid literalWithCharPtrCompare cppcheck finding */
 #define streq(a, b) \
 	(((const char *) a != NULL) && ((const char *) b != NULL) && (strcmp(a, b) == 0))
 
+/* casting to const char * is necessary to avoid literalWithCharPtrCompare cppcheck finding */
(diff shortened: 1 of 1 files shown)


dimitri (commented):
Nice, thanks for a simple and direct fix!

I think it would be best to add a comment about why the macro is using this cast (basically the text from the PR would probably do). Sometimes repeating the same information in the git message and in the code is useful, and I think that's one of this times.

hanefi:
@rimbi can you add a comment for the macro definition as it is not really obvious why we need the casts there?

After that, this one is good to merge IMO.

dimitri:
I wanted to improve the comment, and I'm not sure I understand the cppcheck warning correctly. The literal string is a `char []` in memory I believe, but will this cppcheck error be reported when using a `char *` to `char[]` comparison actually?

rimbi:
What the `cppcheck` does not like is, when a statement like `streq("left", "right")` is expanded, at some point it sees a statement like `"left" != NULL` which does not look right. IMHO the best thing to avoid all of these would be to turn these macros into `inline functions` but I don't want to dive into another discussion series at the moment.

PR #851, opened 2024-07-22
Add github action for cppcheck

rimbi:


hanefi:
Really nice idea. What is the plan to get this ready to merge?

How about:

1. Fix errors, and warnings in small PRs
	- [ ] `uninitvar`. I do not really understand having a character array of size zero that is interpreted as uninitialized.
	- [ ] `unknownMacro`. cppcheck could not find the definition of `PG_MAJORVERSION`. Safe to add an ignore rule, I suppose.
	- [ ] `literalWithCharPtrCompare` #850
2. Add exclusion comments if we can not fix the code in this PR
3. In case of errors or warnings, store output in `GITHUB_STEP_SUMMARY` so that the error message is easier to access for the developers.

.github/workflows/run-tests.yml
```diff
@@ -91,3 +91,16 @@ jobs:
...
+
+  cppcheck:
+    name: cppcheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Execute cppcheck
+        uses: rimbi/cppcheck-action@master
```
dimitri, line 103: Do you plan on contributing that upstream? I think we should consider merging only after that merge, would you agree?

rimbi: We can handle it internally without depending on external actions. I'll change it later.

hanefi: It is possible to have the definition of the action inside the project.

Citus has definitions of several actions in the following directory: https://github.com/citusdata/citus/tree/main/.github/actions

rimbi: Sounds good. We can apply the same approach unless there is a simpler solution. 👍

PR #852, opened 2024-07-23
Fix test_decoding schema parsing with quoted identifier

arajkumar:
The current parser fails when the DML message table schema is escaped with quotes.

Remove the quote check, as the existing logic already covers schemas with quotes.


dimitri:
The problem with quoted identifier is that our parser is not smart enough about identifiers that contain either dots or columns characters in their name, which now could happen. We need to implement an actual parser for a quoted identifier that knows how to walk through it and skip the right set of characters etc. I think it would be best to include that work in the current PR.

arajkumar:
> The problem with quoted identifier is that our parser is not smart enough about identifiers that contain either dots or columns characters in their name, which now could happen. We need to implement an actual parser for a quoted identifier that knows how to walk through it and skip the right set of characters etc. I think it would be best to include that work in the current PR.

@dimitri Should we an equivalent parser like Postgres [parse_ident](https://github.com/postgres/postgres/blob/3937cadfd4383fd32f3fd8d8dc5efc32b3d12664/src/backend/utils/adt/misc.c#L861)? or may be we can assume we will a valid identifier here and implement some dumb parser which finds first dot after all escaping?

dimitri:
> @dimitri Should we build an equivalent parser like Postgres [parse_ident](https://github.com/postgres/postgres/blob/3937cadfd4383fd32f3fd8d8dc5efc32b3d12664/src/backend/utils/adt/misc.c#L861)? or may be we can assume we will a valid identifier here and implement some dumb parser which finds first dot after all escaping?

We should assume a valid identifier here, for sure. I think all we need is to be able to navigate through the chars knowing if we are still within the quoted identifier text, reading the delimiter quote, or reading something that's not in the quoting (the dot, for instance, is never quoted).

arajkumar:
@dimitri I attempted to parse the identifier with the assumption that it would be well-formed.

src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -300,6 +302,55 @@ parseTestDecodingMessage(StreamContext *privateContext,
 }
 
 
+/*
+ * findIdentifierEndPos returns the first position of the separator in the given
+ * message after navigating through all the quotes. It assumes that the message
+ * contains a well-formed identifier and doesn't attempt to handle errors.
+ */
+static int
+findIdentifierEndPos(const char *message, char separator)
```
dimitri, line 311: I think it would be best if that handled being passed a NULL message, which would hint towards our usual approach of returning a boolean (success). Also, I suppose we could return the position where the dot is and the position of the first character after the identifier (or NULL), so we could probably use a `typedef struct` as a (modifiable) argument to the function.

src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -300,6 +302,55 @@ parseTestDecodingMessage(StreamContext *privateContext,
...
+		 * separator: '.'
+		 * "Foo Bar.Baz".hello: UPDATE:
+		 *		        ^
+		 * separator: ':'
+		 * "Foo Bar.Baz": UPDATE:
+		 *		        ^
+		 */
+		else if (message[pos] == separator && quoteCount % 2 == 0)
+		{
+			break;
```
dimitri, line 346: On a first look, I believe this misses the possibility of an identifier embedding quotes using the Postgres syntax `"Foo""Bar"` for `Foo"Bar`.

src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -319,25 +370,19 @@ parseTestDecodingMessageHeader(TestDecodingHeader *header, const char *message)
...
 	char *idp = (char *) message + strlen("table ");
-	char *dot = strchr(idp, '.');
-	char *sep = strchr(idp, ':');
+
+	int schemaEnd = findIdentifierEndPos(idp, '.');
+	char *dot = idp + schemaEnd;
+
+	int tableEnd = findIdentifierEndPos(dot, ':');
+	char *sep = dot + tableEnd;
+
```
dimitri, line 379: What about the idea of passing a struct to fill-in and calling the function only once?

src/bin/pgcopydb/ld_test_decoding.c
dimitri: On-top of that we have to take care of unicode special syntax, see https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

arajkumar: Right, I will check on the unicode part, but "Foo""Bar" is working as expected. I modified the test to include "Foo""Bar".

src/bin/pgcopydb/ld_test_decoding.c
arajkumar: The idea is to support identifier parsing for the following 3 scenarios with different end separator.

1) schema name parsing with separator  '.'
2) table name parsing with separator ':'
3) column name parsing with separator '['

IMHO, a single function with separator as a argument makes more sense here.

src/bin/pgcopydb/ld_test_decoding.c
arajkumar: @dimitri The current implementation also seem to handle unicode chars, I've also added tests with unicode chars which covers unicode in schema, table name, column and types.

src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -300,6 +302,65 @@ parseTestDecodingMessage(StreamContext *privateContext,
...
+		 *		        ^
+		 */
+		else if (message[pos] == separator && quoteCount % 2 == 0)
+		{
+			*position = pos;
+			return true;
+		}
+	}
+
+	return false;
```
dimitri, line 360: Please consider adding a comment about why we would return false here.

src/bin/pgcopydb/ld_test_decoding.c
```diff
@@ -300,6 +302,65 @@ parseTestDecodingMessage(StreamContext *privateContext,
...
+		{
+			quoteCount++;
+		}
+
+		/*
+		 * We are looking for the first 'separator' in the message which
+		 * should be outside the quotes.
+		 *
+		 * When there is 'separator' inside quotes, quoteCount will be odd,
+		 * and we need to account it as a part of the identifier.
```
dimitri, line 338: I think that only works because of identifier escaping rules: it's using `""` to represent a `"` character, not `\"` like in other places, and also the separator itself can not be `"`. Can we add a check about the separator not being a double-quote character?

src/bin/pgcopydb/ld_test_decoding.c
dimitri: Thanks for the added test case! Also I now realize why `"Foo""Bar"` works and my initial reading of your code was wrong. I have added a comment about that, I think we should make sure that the separator can't be `"`.

src/bin/pgcopydb/ld_test_decoding.c
dimitri: Ah yes, you're right. Thanks!

dimitri (approved):

PR #853, opened 2024-07-23
Remove redundant NULL pointer checks

rimbi:
This is a `cppcheck` finding.

### DIFF
--- a/src/bin/pgcopydb/pgcmd.c
+++ b/src/bin/pgcopydb/pgcmd.c
@@ -1676,14 +1676,8 @@ tokenize_archive_list_entry(ArchiveToken *token)
 		char *ptr = line;
 
 		/* advance ptr as long as *ptr is a digit */
-		for (; ptr != NULL && isdigit(*ptr); ptr++)
+		for (; isdigit(*ptr); ptr++)
 		{ }
-
-		if (ptr == NULL)
-		{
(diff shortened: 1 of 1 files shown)


src/bin/pgcopydb/pgcmd.c
```diff
@@ -1676,15 +1676,9 @@ tokenize_archive_list_entry(ArchiveToken *token)
...
-		for (; ptr != NULL && isdigit(*ptr); ptr++)
+		for (; isdigit(*ptr); ptr++)
 		{ }
 
-		if (ptr == NULL)
-		{
-			log_error("Failed to tokenize Archive Item line: %s", line);
-			return false;
-		}
-
```
hanefi, line 1687: I'd be happier if you briefly shared why you wanted to remove this check here.

If we keep this check here, we'd have guarantees that there are some more characters after the token that we parsed. I do not really see why we need this check, but I am unsure if it makes sense to remove these lines here.

rimbi: I see. The reason is that there is no way for `ptr` to be `NULL`. If it is `NULL`, then it means that `line` is also `NULL`, which means that the preceding `isdigit(*line)` ends with an error.

hanefi (approved):

PR #867, opened 2024-08-12
Increase max allowed chars in Archive TOC lines

hanefi:
The maximum line length for an Archive TOC entry is currently limited to 1024, which is too short for some objects such as functions and procedures with a high number of parameters.

Increase the maximum line length to 8 kilobytes, which should be enough for all objects.


hanefi:
Here is a description of the commits:

- a4ea5d4 adds a failing test case. The error message indicate that we fail to catch lines of Archive TOC that does not fit into our buffers. This needs to be investigated.
- ef6f20c increases max allowed characters from 1024-1 to (8*1024)-1. After this change, CI is able to parse the TOC, but we observe some regression due to txn ids changing.
- c93d422 updates the test outputs to have the CI pass again.
- eb31de8 changes the logic for comparing logical decoding outputs so that we normalize transaction ids in logical decoding outputs. The tests still pass
- 79c1850 revert changes in logical decoding outputs, and observe tests are still passing. This is an indication that the normalization logic is sane.

dimitri (approved):

Issue #710, opened 2024-03-19
Issue with toast column in cdc apply for test_decoding


pratikshetty1994:
We are using only the follow the command with test_decoding to capture cdc from aws rds  postgres 14.5 to aws rds postgres 14.5.
During the prefect phase everything works perfectly but when i enable the apply process. We are observing that TOAST columns that are unchanged are handled differently in test_decoding which is resulting the apply process failing.
`'unchanged-toast-datum'` seems to be the cause of issue.
Reference: https://debezium.io/blog/2019/10/08/handling-unchanged-postgres-toast-values/

```
Postgres result status is PGRES_FATAL_ERROR
21:35:33.920 11 ERROR  pgsql.c:2028              [TARGET 12277] ERROR:  invalid input syntax for type json
21:35:33.920 11 ERROR  pgsql.c:2028              [TARGET 12277] DETAIL:  Token "unchanged" is invalid.
21:35:33.920 11 ERROR  pgsql.c:2028              [TARGET 12277] CONTEXT:  JSON data, line 1: unchanged...
21:35:33.920 11 ERROR  pgsql.c:2028              [TARGET 12277] unnamed portal parameter $16 = '...'
21:35:33.920 11 ERROR  pgsql.c:2039              [TARGET 12277] SQL query: EXECUTE 4756fb82;
21:35:33.920 11 ERROR  pgsql.c:2047              [TARGET 12277] SQL params: '###', '###', '## - ### ##', NULL, NULL, NULL, NULL, 'success', '###', NULL, '2024-03-01 18:09:24.323412', '2024-03-01 18:12:15.94', '10000.00', '####', NULL, 'unchanged-toast-datum', 
21:35:33.920 11 ERROR  ld_apply.c:523            Failed to apply SQL from file "/dbhome/tmp//cdc/00000001000074990000000D.sql", see above for details


```

apduvuri:
Hi @dimitri and @marikkan-microsoft,

We got different errors during migration from On-Premises -

`Migration-1`

```bash
2024-07-16T03:59:44.205Z	testdb	153	follow.c	NOTICE	             Starting the catchup sub-process
2024-07-16T03:59:45.785Z	testdb	153	ld_apply.c	INFO	             Replaying changes from LSN 59D/90001E50
2024-07-16T14:21:00.570Z	testdb	153	pgsql.c	    ERROR	             [TARGET 7741] [22P02] ERROR:  malformed array literal: "unchanged-toast-datum"
[...]

PR #877, opened 2024-09-02
test_decoding transform: skip toast columns with unchanged values

hanefi:
test_decoding plugin may not include the value of a TOAST column if it has not changed. In that case, instead of the value of the column the string "unchanged-toast-datum" is included in the output. In that case, we should remove the column from the range table entry list in the generated SQL queries instead of including it and setting that column to the value "unchanged-toast-datum".

There is one strange case that if a table has only toast columns, and an update query that does not update any value is executed, we end up filtering all the range table entries and skip executing any sql queries as it will be no-op.

Fixes: #350 
Fixes: #710


dimitri (approved):
It looks like it could be worth adding a SQL query to the tests to show that the values are indeed “uncompressed external” or “compressed external”. I can't remember how to do that, but I think it is possible to check at run-time?

hanefi:
> It looks like it could be worth adding a SQL query to the tests to show that the values are indeed “uncompressed external” or “compressed external”. I can't remember how to do that, but I think it is possible to check at run-time?

Checking for storage type:

```sql
select att.attname,
       case att.attstorage
          when 'p' then 'plain'
          when 'm' then 'main'
          when 'e' then 'external'
          when 'x' then 'extended'
       end as attstorage
from pg_attribute att
  join pg_class tbl on tbl.oid = att.attrelid
  join pg_namespace ns on tbl.relnamespace = ns.oid
where tbl.relname = 'xpto'
  and ns.nspname = 'public'
  and not att.attisdropped
order by 1;

+--------------+------------+
|   attname    | attstorage |
+--------------+------------+
| cmax         | plain      |
| cmin         | plain      |
| ctid         | plain      |
| id           | plain      |
| rand1        | plain      |
| rand2        | plain      |
| tableoid     | plain      |
| toasted_col1 | extended   |
| toasted_col2 | extended   |
| xmax         | plain      |
| xmin         | plain      |
+--------------+------------+
```

```c
#define  TYPSTORAGE_PLAIN		'p' /* type not prepared for toasting */
#define  TYPSTORAGE_EXTERNAL	'e' /* toastable, don't try to compress */
#define  TYPSTORAGE_EXTENDED	'x' /* fully toastable */
#define  TYPSTORAGE_MAIN		'm' /* like 'x' but try to store inline */
```

--------------------------------

Checking for column compression:

```sql
select id,
       pg_column_compression(toasted_col1) toasted_col1_compression,
       pg_column_compression(toasted_col2) toasted_col2_compression
from xpto
order by 1;

+----+--------------------------+--------------------------+
| id | toasted_col1_compression | toasted_col2_compression |
+----+--------------------------+--------------------------+
|  1 | (null)                   | (null)                   |
|  2 | (null)                   | pglz                     |
+----+--------------------------+--------------------------+
```

It turns out that the comments were partially wrong. We have extended storage instead of external. I guess the 11 year old PG tests are also misleading as I copied the table definitions from them. https://github.com/postgres/postgres/blob/82b07eba9e8b863cc05adb7e53a86ff02b51d888/contrib/test_decoding/sql/toast.sql#L17-L21

dimitri:
Merged as it is, please see if you want to add tests coverage for actual storage options of the columns in the unit tests. The goal I have in mind is to make sure that the actual Postgres behavior conforms with the unit tests expectations: limits could change in a new major version of Postgres in a way that we need to adjust the test data.

Issue #750, opened 2024-04-18
Migration failed while update for schema containing composite primary key


tanyaranjan1995:
Issue: Migration failed while update for schema containing composite primary key
Version: Running pgcopydb version 0.15 from "/usr/bin/pgcopydb"

**Repro steps:**
1. Initiate migration with: pgcopydb clone --follow --drop-if-exists --restart --plugin test_decoding --table-jobs 4 --index-jobs 4
Source database containing:
```create table test_002
(id integer,
	name text,
 constraint test_pk PRIMARY KEY(id,name)
);
insert into test_002 (id,name) values (1,'postgres');
insert into test_002 (id,name) values (2,'oracle');
insert into test_002 (id,name) values (3,'mysql');
insert into test_002 (id,name) values (4,'couchbase');
insert into test_002 (id,name) values (5,'redis');
```
2. During the cutover phase executed the following commands-
```
Insert into test_002 (id,name) values (8,'AWS RDS12');
Insert into test_002 (id,name) values (9,'AWS Aurora123');
```
Executing the above statements was successful.

Migration failed when executing-
```
update test_002
set id = id
where 1 = 1
```

Error Logs:
```
15:22:14.907 8682 INFO   Reported write_lsn A/20000848, flush_lsn A/20000818, replay_lsn A/200007E0
15:22:24.911 8682 INFO   Reported write_lsn A/20000880, flush_lsn A/20000848, replay_lsn A/20000818
15:22:30.588 8683 ERROR  Failed to parse decoding message for UPDATE on table public.test_002: SET clause columns not found
15:22:30.589 8683 ERROR  Failed to parse UPDATE new-tuple columns for logical message table public.test_002: UPDATE: id[integer]:1 name[text]:'postgres'
15:22:30.589 8683 ERROR  Failed to parse test_decoding UPDATE message: table public.test_002: UPDATE: id[integer]:1 name[text]:'postgres'
15:22:30.589 8683 ERROR  Failed to parse test_decoding message, see above for details
15:22:30.589 8683 ERROR  Failed to parse JSON message: {"action":"U","xid":"0","lsn":"A/20000880","timestamp":"2024-04-18 09:52:31.251981+0530","message":"table public.test_002: UPDATE: id[integer]:1 name[text]:'postgres'"}
15:22:30.589 8683 ERROR  Failed to transform JSON messages from input stream, see above for details
15:22:30.589 8683 INFO   Transform process has terminated
[...]

Issue #844, opened 2024-07-16
CDC update on tables with  column type as "generated always as identity" fails


arajkumar:
Updating a row value in the following schema fails with error

```
create table metrics( "SerialNumber" integer generated always as identity, name text);
```

```
ERROR:  column "SerialNumber" can only be updated to DEFAULT
DETAIL:  Column "SerialNumber" is an identity column defined as GENERATED ALWAYS.
```

Currently, we handle only:
  - Generated columns with is_generated = 'ALWAYS' for INSERT and UPDATE (worked on #664 )
  - IDENTITY columns for INSERT using "overriding system value"

PR #883, opened 2024-09-18
Fix invalid UPDATE stmt when none of the columns change

arajkumar:
When all column values in the SET clause are equal to those in the WHERE clause, we remove all columns from the SET clause. This results in an invalid UPDATE statement like the one shown below:
```
UPDATE table SET WHERE "id" = 1;
```

Usually, the above could happen when REPLICA IDENTITY is set to FULL, and the UPDATE statement executed with the same values as the old ones.
For e.g.
```
UPDATE table SET "id" = 1 WHERE "id" = 1;
```

**Solution**: Skip the UPDATE when all column values in SET clause are equal to the WHERE clause values.

Fixes https://github.com/dimitri/pgcopydb/issues/750


arajkumar:
Hmmm, I think the alternative approach is the right one.

dimitri:
Do we need to worry about triggers on the target side? UPDATE statement that look like they change nothing can be used to execute a trigger... on the replica system though, this would have to be a specifically crafted trigger that still is executed with session_replication_role set to replica.

arajkumar:
@dimitri In this PR, I'm skipping the UPDATE stmt when all columns in SET clause are duplicates of WHERE clause values. 

Once we fix https://github.com/dimitri/pgcopydb/issues/844, we shall also remove the duplicate removal logic.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2227,12 +2227,15 @@ stream_write_update(FILE *out, LogicalMessageUpdate *update)
 						  update->table.relname);
 		int pos = 0;
 
+		int skippedColumns = 0;
```
dimitri, line 2230: Can re rename to `skipColumnCount` or something? at the moment this reads more like a boolean value to me.

src/bin/pgcopydb/ld_transform.c
```diff
@@ -2360,20 +2367,44 @@ stream_write_update(FILE *out, LogicalMessageUpdate *update)
...
+		 * Skip the UPDATE statement in such cases.
+		 */
+		bool skipUpdate = (skippedColumns == new->attributes.count);
 
-		FFORMAT(out, "PREPARE %x AS %s;\n", hash, buf->data);
+		if (skipUpdate)
+		{
+			log_warn("Skipping UPDATE statement as all columns are "
+					 "the same as the old");
+		}
```
dimitri, line 2390: If we would `return true;` from here, we wouldn't need an else branch at all. What do you think?

dimitri (commented):
Finally had time to review. Thanks for the great work again!

I was initially confused about what we already do in pgcopydb before your PR and what's added in your PR. I now got clarity on that, and appreciate your changes. Apart from the nitpicking in the review, I think we should add some mention of what we're doing here in the docs.

Maybe between https://pgcopydb.readthedocs.io/en/latest/ref/pgcopydb_follow.html#replica-identity-and-lack-of-primary-keys and https://pgcopydb.readthedocs.io/en/latest/ref/pgcopydb_follow.html#logical-decoding-pre-fetching in a new section titled “Statement Rewriting” or something.

arajkumar:
@dimitri Thanks for reviewing this PR.

>  I think we should add some mention of what we're doing here in the docs.

I don't think we should add to the doc, This is just a big fix to avoid generation of invalid SQL UPDATE statement. Also, I believe we should just implement skipping identity columns being included as part of UPDATE's SET clause(https://github.com/dimitri/pgcopydb/issues/844) and remove the deduplication logic all together. WDYT?

dimitri (approved):

dimitri:
> I don't think we should add to the doc, This is just a big fix to avoid generation of invalid SQL UPDATE statement.

As a change of behavior, I think you're right, no need to document the fix. That said, I think our overall approach should be documented. Saying that makes me realize it's not fair to add that burden to your PR (now merged), I will think more about what I think is needed and take care of it.

> Also, I believe we should just implement skipping identity columns being included as part of UPDATE's SET clause(#844) and remove the deduplication logic all together. WDYT?

First part, agreed. Second part (remove the deduplication logic), I'm not sure I understand it. Can you provide a couple examples showing what we do, why it's wrong, and what you propose we should do instead?

Issue #899, opened 2025-02-05
Segmentation fault when pgcopydb clone hits an error


Bodobolero:
## Scenario:

```bash
export "PGCOPYDB_SOURCE_PGURI=postgresql://..."
export "PGCOPYDB_TARGET_PGURI=postgresql://..."
export PGOPTIONS="-c maintenance_work_mem=8388608 -c max_parallel_maintenance_workers=7"

PG_CONFIG=/usr/local/pgsql/bin/pg_config pgcopydb clone \
        --resume \
        --not-consistent \
        --dir /home/admin/logs \
        --skip-vacuum \
        --skip-extensions \
        --no-owner \
        --no-acl \
        --skip-db-properties \
        --table-jobs 4 \
        --index-jobs 4 \
        --restore-jobs 4 \
        --split-tables-larger-than 10GB \
        --skip-extensions \
        --drop-if-exists \
        --use-copy-binary 2>&1 | tee /home/admin/logs/pgcopydb.log

## Problem:

```bash
2025-02-05 09:50:40.559 50441 ERROR  copydb.c:772              Sub-process 50445 exited with code 0 and signal Segmentation fault
```

gdb output

```bash
#0  0x0000aaaac16b7a9c in strlcpy ()
#1  0x0000aaaac15daafc in pgcopy_log_error (pgsql=pgsql@entry=0xfffff6625600, res=res@entry=0xaaaaf51fee90,
    context=0xaaaaf528b5d0 "copy workflows.vertices(uuid,created_at,condition_block_id,operator,has_been_visited,reference_id,workflow_id,meta_data,id,action_block_id) from stdin with (format binary)") at pgsql.c:3181
#2  0x0000aaaac15dae08 in pg_copy_send_query (pgsql=pgsql@entry=0xfffff6625600, args=args@entry=0xffff8001bce8, status=status@entry=PGRES_COPY_IN) at pgsql.c:3151
#3  0x0000aaaac15dd0f0 in pg_copy_data (callback=0xaaaac15f03c0 <copydb_update_copy_stats_hook>, context=0xfffff66250c0, stats=0xfffff66250b0, args=0xffff8001bce8, dst=0xfffff6625600, src=0xfffff6634bb0) at pgsql.c:2744
[...]

PR #902, opened 2025-02-15
Fix a segfault when SQL state is NULL

Bodobolero:


src/bin/pgcopydb/pgsql.c
```diff
@@ -3133,7 +3133,15 @@ pgcopy_log_error(PGSQL *pgsql, PGresult *res, const char *context)
 	if (res != NULL)
 	{
 		char *sqlstate = PQresultErrorField(res, PG_DIAG_SQLSTATE);
-		strlcpy(pgsql->sqlstate, sqlstate, sizeof(pgsql->sqlstate));
+		if (sqlstate == NULL)
+		{
+			// PQresultErrorField returned NULL!
```
dimitri, line 3138: Code style checks prevents usage of C++ style comments. Please use the C syntax `/* ... */` to fix CI checks.

dimitri (changes requested):

src/bin/pgcopydb/pgsql.c
Bodobolero: I could not get your style checkers to run on my mac, including citus-indent, you may want to add a contributors instructions for those using mac and not Linux

dimitri (approved):

dimitri:
Fixes #899

PR #911, opened 2025-04-03
add arm64 build for docker image

dhanushreddy291:
thought it would be beneficial to have an `arm64` build of `pgcopydb`, as currently only the `amd64` version is available

image

also added cache if it helps (or you can remove it. Thanks)

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -85,6 +85,10 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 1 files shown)


dimitri (approved):

dimitri:
Thanks for your contribution @dhanushreddy291 ; the build seems to be rather slow. Is it expected and worth it in your opinion?

dimitri:
Okay, let's try it then and see how it goes. Worst case, I might have to revert this commit. Crossing my fingers we don't have to! Thanks for your contribution @dhanushreddy291 !

dhanushreddy291:
awseome @dimitri, I guess first one won't be having any cache and will take more time, but if you retry github action, or push any dummy commit, it will be faster as it uses cache

PR #925, opened 2025-09-25
Fix Syntax error at or near "\"

ahmad-sanad:
Fixes https://github.com/dimitri/pgcopydb/issues/922 and tests

Postgres 17.6 (and 16.10, etc) added `\restrict` and `\unrestrict` metacommands to pg_dump output to mitigate a security issue. This code skips these lines when restoring roles.

17.6 change notes: https://www.postgresql.org/docs/17/release-17-6.html#RELEASE-17-6-CHANGES
CVE: https://www.postgresql.org/support/security/CVE-2025-8714/

PR #926, opened 2025-09-26
Fix catalog mismatch issue

### LINKED ISSUE #869: pgcopydb list table-parts fails when filtering has been used to initialize the catalog
When filtering has been used for a migration, `pgcopydb list table-parts` fails with:
```sh
# prior to running list table-parts we use filtering
pgcopydb clone --filters filter.ini ...

# at this point the catalog is initialized and list table-parts will fail
pgcopydb list table-parts --schema-name public --table-name pairs --split-tables-larger-than "50GB"

13:50:02.203 13886 INFO   Running pgcopydb version 0.17-1.pgdg22.04+1 from "/usr/bin/pgcopydb"
13:50:02.234 13886 INFO   Using work dir "/tmp/pgcopydb"
13:50:02.235 13886 INFO   Current filtering setup is: {"type":"SOURCE_FILTER_TYPE_NONE"}
13:50:02.235 13886 INFO   Catalog filtering setup is: {"type":"SOURCE_FILTER_TYPE_INCL","include-only-table":[{"schema":"public","name":"pairs"}]}
13:50:02.235 13886 ERROR  Catalogs at "/tmp/pgcopydb/schema/source.db" have been setup for a different filtering than the current command, see above for details
13:50:02.235 13886 ERROR  Failed to initialize pgcopydb internal catalogs

```

ahmad-sanad:
Fixes https://github.com/dimitri/pgcopydb/issues/869

Caused by using commands that support filter after ones that don't, or vice versa. For example, `snapshot` then `clone --filter` or `clone --filter` then `stream sentinel`. 

This code tries to deduce which filters should be used or gives an error when that isn't straightforward.

Note: tests fail for me without https://github.com/dimitri/pgcopydb/pull/925

### DIFF
--- a/src/bin/pgcopydb/catalog.c
+++ b/src/bin/pgcopydb/catalog.c
@@ -6,6 +6,7 @@
 #include <inttypes.h>
 #include <limits.h>
 #include <stdlib.h>
+#include <string.h>
 #include <sys/select.h>
 #include <time.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


dimitri (approved):

dimitri:
I think you need to allow edits from mainteners so that I can “update branch” to current main status and run the CI.

dimitri:
Merged in #952.

PR #933, opened 2025-12-15
test_decoding: Fix parsing quoted null

### LINKED ISSUE #931: In follow mode, null as text (i.e. not actual NULL) is deserialized as actual NULL
If user inserts a new row which has a text or jsonb column with value "null" as string, (i.e. not actual `NULL`), it is deserialized as actual `NULL`, effectively corrupting the data.

Steps to reproduce:
Before `clone --follow`:
```sql
CREATE TABLE t(c text);
```

During `clone --follow`:
```sql
INSERT INTO t(c) VALUES("null");
```

On source:
```sql
SELECT * FROM t;
  c   
------
 null
(1 row)
```

On target:
```sql
SELECT * FROM t;
  c   
------

(1 row)
```

The test_decoding output seems correct:
```
{"action":"I","xid":"0","lsn":"0/5E0000A8","timestamp":"2025-10-09 20:00:59.240677+0000","message":"table public.t: INSERT: c[text]:'null'"}
```

For actual `NULL`, the test_decoding output is as follows:
```
The test_decoding output seems correct:
{"action":"I","xid":"0","lsn":"0/5E0000A8","timestamp":"2025-10-09 20:00:59.240677+0000","message":"table public.t: INSERT: c[text]:null"}
```

Howewe, they both generate the same SQL:
```sql
PREPARE a75e754c AS INSERT INTO public.t (c) overriding system value VALUES ($1);
EXECUTE a75e754c[null];
```

It seems the issue is combination of these two pieces of code:
- First we ignore the quotes while reading the test_decoding output [here](https://github.com/dimitri/pgcopydb/blob/c1e07da3d257d0d3a1fc321f3c7e1a4063eed6bd/src/bin/pgcopydb/ld_test_decoding.c#L885),
- Then we set isNull to true if the value we read equals to "null" [here](https://github.com/dimitri/pgcopydb/blob/c1e07da3d257d0d3a1fc321f3c7e1a4063eed6bd/src/bin/pgcopydb/ld_test_decoding.c#L1018)

matixlol:
Fixes https://github.com/dimitri/pgcopydb/issues/931


scottjacobsen:
Hit the same bug in production with a quoted email-shaped value (a real address starting with `null…@…`) → SQL `null` → NOT NULL constraint violation on the target. Confirmed your fix here resolves it (the `valueLen == 4` guard makes the prefix-with-suffix case fall through correctly).

One small thing the existing `null_texts` fixture in this PR doesn't explicitly cover: quoted strings whose contents *start with* `null` but are longer than 4 chars (e.g. `'nullable@example.com'`, `'nulls'`). The fix handles them, but they're not exercised by the test. To round out regression coverage at the parser layer (no docker round-trip), I added a fixture-based test in `tests/cdc-test-decoding/` plus one tiny refinement: defensive `cols->isQuoted = false` at the top of `parseNextColumn` (lifted from teknogeek0/pgcopydb#9 — covers the raw/unchanged-toast paths without depending on caller-side zero-init).

Two commits on a branch on my fork — happy to roll them into this PR, open a follow-up, or just leave them here as a reference:

- Test fixture: https://github.com/scottjacobsen/pgcopydb/commit/5cb15433b18e
- Fix (mirrors yours + defensive init): https://github.com/scottjacobsen/pgcopydb/commit/0b84a34e1950

Cases covered by the new fixture (single INSERT round-trip):
- `'null'` (the case in this PR)
- `'nullable@example.com'` (synthetic; the prefix-match, len > 4 case)
- `'nulls'` (5-char prefix)
- `'null '` (trailing space inside quotes)
- `'NULL'` (uppercase sanity)
- `''` (empty string)
- unquoted `null` (must remain NULL — regression guard)

Thanks for the fix — would be great to see this land.

dimitri (approved):
Looks good, thanks for the fix! As the comment says, would be good to improve the test case coverage a little more.

Issue #772, opened 2024-05-08
bug?: include-table filter fails due to oid collision with postgis extension


alefminus:
Hi,

 I have no produced a simple reproduction for this yet, but using latest commit (4f41dd430c8160876b89c26a99d81fc068d54c99) and a include-table with a single exclude-schema filter I got the table excluded. Some added log messages and I saw that the exclusion was based on oid from ```copydb_objectid_is_filtered_out``` in branch ```if (result.oid != 0)```.
Looking at the sqlite filter contained that oid but the restore_list_name is "postgis" which is unrelated to the table.

Is this a known issue?

More details:
I'm running
```bash
./src/bin/pgcopydb/pgcopydb clone \
  --verbose \
  --no-role-passwords \
  --no-owner \
  --no-acl \
  --dir $DIR \
  --source $SOURCE \
  --target $TARGET \
  --skip-extensions \
  --filters $FILTERS \
   2>&1| tee clone.$(git describe).$(date +%Y%m%d_%H%M%S).log
```
DIR=test3

Where $FILTERS points to a file containing:
```
[include-only-table]
public.sensor_data_raw
[exclude-schema]
grafana
```

And the output shows that the sensor_data_raw table is filtered out:
```
2024-05-08 16:09:00.301 3968766 NOTICE copydb_schema.c:914       copydb_objectid_filtered_out based on oid 16417
2024-05-08 16:09:00.301 3968766 NOTICE dump_restore.c:718        Skipping filtered-out dumpId 662: TABLE 1259 16417 public sensor_data_raw postgres
```

The additional prints are due to this patch:
```patch
--- a/src/bin/pgcopydb/copydb_schema.c
+++ b/src/bin/pgcopydb/copydb_schema.c
@@ -911,6 +911,7 @@ copydb_objectid_is_filtered_out(CopyDataSpec *specs,

                if (result.oid != 0)
                {
+                       log_notice("copydb_objectid_filtered_out based on oid %d", result.oid);
                        return true;
                }
        }
@@ -925,6 +926,7 @@ copydb_objectid_is_filtered_out(CopyDataSpec *specs,

                if (!IS_EMPTY_STRING_BUFFER(result.restoreListName))
                {
+                       log_notice("copydb_objectid_filtered_out based on rlname");
                        return true;
                }
        }
```

But filters.db shows:
```bash
❯ sqlite3 test3/schema/filter.db 'select * from filter where oid = 16417'
16417|postgis|extension
```

Thanks,
[...]

PR #945, opened 2026-04-16
Fix --skip-extensions OID collision filtering out unrelated tables

scottjacobsen:
## Summary

- Fixes `--skip-extensions` incorrectly excluding tables/sequences from `pg_restore` when their OID collides with an extension OID
- Also fixes the same potential issue with `--skip-collations`
- Guards both the OID and `restoreListName` lookup paths

## Problem

PostgreSQL uses independent OID sequences for different catalog tables (`pg_extension`, `pg_class`, `pg_collation`). The same OID value can appear in multiple catalogs. When `--skip-extensions` inserts extension OIDs into pgcopydb's filter table, `copydb_objectid_is_filtered_out()` would match any archive item with that OID — including TABLE and SEQUENCE items from `pg_class` that happen to share the OID.

This causes `pg_restore` to fail because dependent objects (SEQUENCE OWNED BY, DEFAULT, constraints, indexes, GRANTs) are still included in the restore list, but the table they reference was incorrectly excluded.

Reproduces on real-world databases where early-created extensions get low OIDs that overlap with user tables. See #772 for a prior report of this bug.

## Fix

Added `filter_kind_matches_archive_desc()` which checks the filter entry's `kind` against the archive item's `description` before filtering:
- `kind = 'extension'` only matches `EXTENSION` archive items
- `kind = 'coll'` only matches `COLLATION` archive items  
- All other kinds (table, index, sequence, etc.) match freely since they use `pg_class` OIDs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/bin/pgcopydb/copydb.h
+++ b/src/bin/pgcopydb/copydb.h
@@ -405,7 +405,8 @@ bool copydb_prepare_sequence_specs(CopyDataSpec *specs, PGSQL *pgsql, bool reset
 bool copydb_fetch_schema_and_prepare_specs(CopyDataSpec *specs);
 bool copydb_objectid_is_filtered_out(CopyDataSpec *specs,
 									 uint32_t oid,
-									 char *restoreListName);
+									 char *restoreListName,
+									 const char *archiveDesc);
 bool copydb_matview_refresh_is_filtered_out(CopyDataSpec *specs,
 											uint32_t oid);
 
--- a/src/bin/pgcopydb/copydb_schema.c
+++ b/src/bin/pgcopydb/copydb_schema.c
@@ -1012,13 +1012,51 @@ copydb_matview_refresh_is_filtered_out(CopyDataSpec *specs, uint32_t oid)
 
 
 /*
+ * filter_kind_matches_archive_desc returns true when a filter entry's kind
+ * is compatible with the archive item description being checked. This prevents
+ * OID collisions between different catalog types (e.g., pg_extension and
+ * pg_class can have the same OID value) from incorrectly filtering out
+ * unrelated objects.
+ *
+ * Filter kinds whose OIDs come from non-pg_class catalogs (extension from
+ * pg_extension, coll from pg_collation) must only match their own archive
+ * item type. Other kinds (table, index, sequence, etc.) use pg_class OIDs
+ * and match freely since they share the same OID namespace.
+ */
+static bool
+filter_kind_matches_archive_desc(const char *kind, const char *archiveDesc)
+{
+	/* if no archive description or no kind, assume match (backwards compat) */
+	if (archiveDesc == NULL || IS_EMPTY_STRING_BUFFER(kind))
+	{
+		return true;
+	}
+
+	/* extension OIDs come from pg_extension, not pg_class */
+	if (strcmp(kind, "extension") == 0)
+	{
+		return strcmp(archiveDesc, "EXTENSION") == 0;
+	}
+
+	/* collation OIDs come from pg_collation, not pg_class */
+	if (strcmp(kind, "coll") == 0)
+	{
+		return strcmp(archiveDesc, "COLLATION") == 0;
+	}
+
+	return true;
+}
+
+
+/*
  * copydb_objectid_is_filtered_out returns true when the given oid belongs to a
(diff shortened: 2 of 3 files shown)


src/bin/pgcopydb/copydb_schema.c
```diff
@@ -1011,14 +1011,52 @@ copydb_matview_refresh_is_filtered_out(CopyDataSpec *specs, uint32_t oid)
...
+filter_kind_matches_archive_desc(const char *kind, const char *archiveDesc)
+{
+	/* if no archive description or no kind, assume match (backwards compat) */
+	if (archiveDesc == NULL || IS_EMPTY_STRING_BUFFER(kind))
+	{
+		return true;
+	}
+
+	/* extension OIDs come from pg_extension, not pg_class */
+	if (strcmp(kind, "extension") == 0)
```
dimitri, line 1036: Can we re-use `streq()` here?

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -1011,14 +1011,52 @@ copydb_matview_refresh_is_filtered_out(CopyDataSpec *specs, uint32_t oid)
...
+	{
+		return strcmp(archiveDesc, "EXTENSION") == 0;
+	}
+
+	/* collation OIDs come from pg_collation, not pg_class */
+	if (strcmp(kind, "coll") == 0)
```
dimitri, line 1042: `streq`

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -1011,14 +1011,52 @@ copydb_matview_refresh_is_filtered_out(CopyDataSpec *specs, uint32_t oid)
...
+	/* if no archive description or no kind, assume match (backwards compat) */
+	if (archiveDesc == NULL || IS_EMPTY_STRING_BUFFER(kind))
+	{
+		return true;
+	}
+
+	/* extension OIDs come from pg_extension, not pg_class */
+	if (strcmp(kind, "extension") == 0)
+	{
+		return strcmp(archiveDesc, "EXTENSION") == 0;
```
dimitri, line 1038: `streq`

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -1011,14 +1011,52 @@ copydb_matview_refresh_is_filtered_out(CopyDataSpec *specs, uint32_t oid)
...
+	}
+
+	/* collation OIDs come from pg_collation, not pg_class */
+	if (strcmp(kind, "coll") == 0)
+	{
+		return strcmp(archiveDesc, "COLLATION") == 0;
```
dimitri, line 1044: `streq`

src/bin/pgcopydb/copydb_schema.c
```diff
@@ -1031,7 +1069,8 @@ copydb_objectid_is_filtered_out(CopyDataSpec *specs,
 			return false;
 		}
 
-		if (result.oid != 0)
+		if (result.oid != 0 &&
+			filter_kind_matches_archive_desc(result.kind, archiveDesc))
```
dimitri, line 1073: Can we first check about the archive desc and only then maybe move on to the catalog lookup, which is more expensive?

src/bin/pgcopydb/copydb.h
```diff
@@ -405,7 +405,8 @@ bool copydb_prepare_sequence_specs(CopyDataSpec *specs, PGSQL *pgsql, bool reset
 bool copydb_fetch_schema_and_prepare_specs(CopyDataSpec *specs);
 bool copydb_objectid_is_filtered_out(CopyDataSpec *specs,
 									 uint32_t oid,
-									 char *restoreListName);
+									 char *restoreListName,
+									 const char *archiveDesc);
```
dimitri, line 409: At this point why not pass in the `ArchiveContentItem *item` itself?

dimitri (changes requested):
Thanks for your contribution!

In the review I'm pointing out to changes that I think should be made before merge. It's about re-using internal APIs that we have provided in the code base, and an opportunity for function prototype cleanup.

dimitri:
replaced by #954

PR #946, opened 2026-04-16
Fix test_decoding parser for REPLICA IDENTITY USING INDEX

scottjacobsen:
## Summary

- Teaches the `test_decoding` UPDATE parser to recognize replica identity columns, not just primary key columns
- Unblocks CDC on tables that use `REPLICA IDENTITY USING INDEX` on a non-PK unique index
- Adds a docker-based regression test

## Problem

For tables configured with `REPLICA IDENTITY USING INDEX` on a unique index that is **not** a primary key, pgcopydb's `test_decoding` parser fails every UPDATE message with:

```
ERROR  ld_test_decoding.c:1201  Failed to parse decoding message for UPDATE on table X: WHERE clause columns not found
ERROR  ld_transform.c:1362     Failed to parse test_decoding message, see above for details
```

Root cause (in `prepareUpdateTuppleArrays`): each column of the incoming UPDATE was sorted into "old key" (WHERE clause) or "new tuple" (SET clause) based on `SourceTableAttribute.attisprimary`. That flag is populated by a SQL query joining `pg_index` with `indisprimary`, so for a table with no primary key it is false for every column — `oldCount` stays zero and parsing fails.

This is how some list-partitioned tables end up configured: the partition key cannot always be part of a primary key, so the replica identity has to be a separate unique index.

## Fix

Add `attisreplident` alongside `attisprimary` on `SourceTableAttribute`, populated from `pg_index.indisreplident`. In the parser, a column counts as part of the WHERE-clause key when it is either a primary-key column **or** a replica-identity column.

Both fields are needed because, for `REPLICA IDENTITY DEFAULT` with a primary key, `pg_index.indisreplident` stays **false** on the PK index (Postgres uses the PK implicitly). So `attisreplident` alone would regress the common default case:

```
postgres=# create table t1 (id int primary key, name text);
postgres=# select indisprimary, indisreplident from pg_index where indrelid = 't1'::regclass;
 indisprimary | indisreplident
--------------+----------------
 t            | f
```

The OR composition preserves `attisprimary`'s existing semantics for any other callers and keeps the fields aligned with their Postgres counterparts (`indisprimary` / `indisreplident`).

## Test plan

- [x] `make -C tests cdc-replica-identity-index` — new regression test passes (asserts source and target rows match after INSERT + UPDATE (identity unchanged) + UPDATE (identity changed) + DELETE)
- [x] `make -C tests cdc-test-decoding` — existing CDC test still passes
- [x] `make -C src/bin/pgcopydb` — clean build

Verified the new test reproduces the bug: running it against the pre-fix parser produced the same `WHERE clause columns not found` error seen in production.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/cdc-replica-identity-index/ddl.sql
```diff
@@ -0,0 +1,30 @@
...
+
+begin;
+
+create table event_matches
+(
+    id            bigserial      not null,
+    created_at    timestamp(6)   not null default now(),
+    retain        boolean        not null default false,
+    name          text           not null,
+    payload       jsonb          not null default '{}'::jsonb
```
scottjacobsen: we can simplify this schema - I don' t believe this field is relevant to this test - we can just update the name field

dimitri (approved):
Thanks for your contribution,

I think I would prefer the test case to be included in an existing logical replication unit test, but that could be refactored later.

Issue #808, opened 2024-06-07
ERROR:  index is already associated with a constraint


apduvuri:
Logs from `pgcopydb index/constraint worker`

```bash
2024-06-04T15:46:50.589Z	test_db	83	indexes.c	NOTICE	            CREATE UNIQUE INDEX t_volume_adjust_pkey ON u_testdb_00.t_cal_volume_adjust USING btree (vat_nbid);
2024-06-04T15:46:50.589Z	test_db	83	pgsql.c	SQL	             [TARGET 28282] CREATE UNIQUE INDEX t_volume_adjust_pkey ON u_testdb_00.t_cal_volume_adjust USING btree (vat_nbid);;
2024-06-04T15:46:50.606Z	test_db	83	indexes.c	NOTICE	           ALTER TABLE u_testdb_00.t_cal_volume_adjust ADD CONSTRAINT t_volume_adjust_pkey PRIMARY KEY USING INDEX t_volume_adjust_pkey
2024-06-04T15:46:50.607Z	test_db	83	pgsql.c	SQL	             [TARGET 28282] ALTER TABLE u_testdb_00.t_cal_volume_adjust ADD CONSTRAINT t_volume_adjust_pkey PRIMARY KEY USING INDEX t_volume_adjust_pkey;
2024-06-04T15:46:50.663Z	test_db	83	indexes.c	NOTICE	           ALTER TABLE u_testdb_00.t_cal_volume_adjust ADD CONSTRAINT t_volume_adjust_pkey PRIMARY KEY USING INDEX t_volume_adjust_pkey
2024-06-04T15:46:50.663Z	test_db	83	pgsql.c	SQL	             [TARGET 28282] ALTER TABLE u_testdb_00.t_cal_volume_adjust ADD CONSTRAINT t_volume_adjust_pkey PRIMARY KEY USING INDEX t_volume_adjust_pkey;
2024-06-04T15:46:50.664Z	test_db	83	pgsql.c	ERROR	             [TARGET 28282] ERROR:  index "t_volume_adjust_pkey" is already associated with a constraint
2024-06-04T15:46:50.667Z	test_db	83	pgsql.c	ERROR	             [TARGET 28282] ERROR:  index "t_volume_adjust_pkey" is already associated with a constraint
[...]

Issue #700, opened 2024-03-11
Unable to copy target table data to source when target partitioned


jnels124:
I am attempting to use pgcopydb to migrate data from our current schema to our new schema which uses citus. When executing `pgcopydb copy table-data --table-jobs 8 --index-jobs 2` errors are logged indicating that `truncate only` was called on the parent table for a partitioned table. Truncate only is only valid to execute on specific partitions but not the parent table.

```
17:32:30.001 784336 ERROR  [TARGET 1674604687] ERROR:  cannot truncate only a partitioned table
17:32:30.001 784336 ERROR  [TARGET 1674604687] HINT:  Do not specify the ONLY keyword, or use TRUNCATE ONLY on the partitions directly.
17:32:30.003 784336 ERROR  [TARGET 1674604687] ERROR:  cannot truncate only a partitioned table
17:32:30.003 784336 ERROR  [TARGET 1674604687] SQL query: TRUNCATE ONLY public.transaction
17:32:30.003 784336 ERROR  [TARGET 1674604687] SQL params:
17:32:30.003 784336 ERROR  Failed to copy data for table with oid 16742 and part number 0, see above for details
```

dimitri:
Is my understanding correct that the table (here `public.transaction`) is a plain table on the source database, but a partitionned-table (and the parent table) on the target database?

arajkumar:
The same problem we have encountered while using pgcopydb for Postgres to Timescale migration as well. The source would have plain Postgres table, but the same table on target will be partitioned using Postgres inheritance.

kbarber:
I've seen this also, very rare but it happens. I'm having trouble reproducing it with my own smaller example however to share.

dimitri:
I think you need to split the schema and data parts of the processing, and then tweak the schema on the target database to introduce partitioning, in order to repro. See https://pgcopydb.readthedocs.io/en/latest/tutorial.html#how-to-edit-the-schema-when-copying-a-database for details.

Now if you can reproduce, the way to fix it might be to decide if we can use TRUNCATE or TRUNCATE ONLY by having a look at the TARGET schema to see if the table is partitioned there, whatever it might be on the source. At the moment we only have catalog entries for tables on the sourceDB, we need to grab them at the right point in time for the targetDB too now.

dimitri:
We have many cases to consider, but today only consider a few.
  - when the table is not partitioned on the source, not partitioned on the table, we're good (truncate)
[...]

PR #949, opened 2026-05-09
Fix TRUNCATE ONLY and COPY FREEZE on partitioned target tables

jmealo:
## Summary

When the source has a flat (non-partitioned) table but the target is a partitioned table with the same qualified name, pgcopydb's data-copy phase fails with:

```
ERROR: cannot truncate only a partitioned table
ERROR: cannot perform COPY FREEZE on a partitioned table
```

The same symptoms recur during CDC follow-mode replay of replicated `TRUNCATE` statements.

This is the asymmetric source-flat / target-partitioned case from #700. The symmetric case (both partitioned) was handled by #863 via a source-side filter. This PR closes the asymmetric gap by querying `pg_class.relkind` on the target at runtime, along the lines of Dimitri's [May 2024 analysis on #700](https://github.com/dimitri/pgcopydb/issues/700#issuecomment-2108023769).

## Approach

- `pgsql_truncate` emits `TRUNCATE` (without `ONLY`) when the target's `relkind` is `'p'`. Postgres' own partition routing handles the cascade.
- `pg_copy` skips `COPY FREEZE` when the target is partitioned, since Postgres rejects FREEZE on partitioned tables.
- `ld_apply.c` (CDC follow) rewrites replicated `TRUNCATE ONLY <qname>` to plain `TRUNCATE <qname>` at apply time when the target is partitioned. The lookup runs on the non-pipeline `controlPgConn` because `applyPgConn` is in pipeline mode and `pgsql_execute_with_params` rejects `parseFun` callbacks there.

A new `pgsql_get_table_relkind` helper performs the lookup, following the same `SingleValueResultContext` pattern used by `pgsql_get_search_path` and `pgsql_has_table_privilege`.

## Tests

Two integration tests, both passing against real Postgres in docker-compose:

- `tests/partitioned-target/` (non-CDC) — flat source, partitioned target with the same qualified name, `pgcopydb copy table-data`. Asserts row counts match and partition routing is correct.
- `tests/cdc-partitioned-target/` (CDC follow) — same fixture, drives INSERTs and a `TRUNCATE` through `stream prefetch` / `stream catchup` in two phases. Asserts:
  - initial copy distributes seed rows correctly (`events_a=2`, `events_b=1`)
  - CDC-replicated INSERTs land in the right partitions (`events_a=3`, `events_b=2`)
  - replicated `TRUNCATE` clears the target (count = 0)

I manually reproduced the bug against the test's target schema (`TRUNCATE ONLY partitioned_target.events` → `cannot truncate only a partitioned table`) to confirm the test exercises the actual bug path rather than something incidental.

## Out of scope

This PR fixes the **copy phase**. The **post-data phase** (CREATE INDEX / ADD CONSTRAINT) still fails when source and target partition shapes differ. Concrete failure modes:

1. `CREATE UNIQUE INDEX` from the source on a column that isn't in the target's partition key — Postgres rejects (unique index on a partitioned table must include the partitioning columns).
2. `ALTER TABLE ... ADD CONSTRAINT ... PRIMARY KEY USING INDEX` — if the partitioned target was pre-created with its own (composite) PK, the index name conflicts. The symptom looks like #808 but the cause is a real schema conflict, not concurrency.
3. `CREATE INDEX` collisions when the partitioned target was pre-created with indexes. `IF NOT EXISTS` is opt-in (resume path) and Postgres has no `IF NOT EXISTS` for constraints.

The current workaround is to pre-create the target schema and use `pgcopydb copy table-data` (skips post-data), or to split the operation as described in the [pgcopydb tutorial — How to edit the schema when copying a database](https://pgcopydb.readthedocs.io/en/latest/tutorial.html#how-to-edit-the-schema-when-copying-a-database). A proper post-data fix is a separate, larger piece of work; happy to scope it next, either under #700 or in a new issue, whichever Dimitri prefers.

## Notes

- The relkind lookup is per-call (no cache). `TRUNCATE` is infrequent in typical CDC traffic, so caching felt premature; happy to add a `(qname → relkind)` cache on `StreamApplyContext` if preferred.
- On lookup failure the code falls back to the pre-PR behavior (issue `TRUNCATE ONLY`, leave `freeze` set). That produces a real Postgres error rather than silent corruption, and `ld_apply.c` emits a `log_warn` so the fallback is observable in operator logs.
- Multi-relation `TRUNCATE ONLY a, b;` is a pre-existing pgcopydb limitation in the `test_decoding` parser (the relname gets garbled at parse time) — not introduced or made worse by this PR. The one-relation-per-statement invariant is now documented on both `stream_write_truncate` (writer) and the apply-side rewrite (consumer).

Closes (partial) #700.

### DIFF
--- a/src/bin/pgcopydb/ld_apply.c
+++ b/src/bin/pgcopydb/ld_apply.c
@@ -1164,7 +1164,51 @@ stream_apply_sql(StreamApplyContext *context,
 				*ptr = '\0';
 			}
 
-			if (!pgsql_execute(applyPgConn, sql))
+			/*
+			 * Postgres rejects TRUNCATE ONLY on partitioned tables. Mirror
+			 * the runtime fix in pgsql_truncate: when the target relation is
+			 * partitioned, drop the ONLY keyword. The lookup runs on the
+			 * non-pipeline controlPgConn because applyPgConn is in pipeline
+			 * mode (which forbids parseFun callbacks; see pgsql_execute_with_params).
+			 *
+			 * Invariant: stream_write_truncate emits exactly one relation per
+			 * statement (LogicalMessageTruncate.table is singular), so the
+			 * single-relation regclass lookup below is sufficient. If a future
+			 * change ever makes this multi-relation, the regclass cast will
+			 * fail, the rewrite will be skipped, and Postgres will surface a
+			 * loud error from the original TRUNCATE ONLY.
+			 */
+			char rewritten[BUFSIZE] = { 0 };
+			const char *truncateSQL = sql;
+			const char onlyPrefix[] = "TRUNCATE ONLY ";
+			size_t prefixLen = sizeof(onlyPrefix) - 1;
+
+			if (strncmp(sql, onlyPrefix, prefixLen) == 0)
+			{
+				char relkind = '\0';
+				const char *qname = sql + prefixLen;
+
+				if (pgsql_get_table_relkind(&(context->controlPgConn),
+											qname, &relkind))
+				{
+					if (relkind == 'p')
+					{
+						sformat(rewritten, sizeof(rewritten),
+								"TRUNCATE %s", qname);
+						truncateSQL = rewritten;
+					}
+				}
+				else
+				{
+					log_warn("Could not resolve relkind for replicated "
+							 "TRUNCATE on \"%s\"; replaying as TRUNCATE ONLY. "
+							 "If the target is partitioned, Postgres will "
+							 "reject this statement.",
+							 qname);
+				}
+			}
+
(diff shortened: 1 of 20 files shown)


jmealo:
CI failure was a transient network error. The tests are green on my fork.

jmealo:
Closing+reopening to retrigger CI; the previous run hit a transient `Connection reset by peer` from `deb.debian.org` during apt-get install in the `endpos-in-multi-wal-txn` test (build never reached test code).

jmealo:
Reopening to retrigger CI after the transient apt-get/network failure noted above.

dimitri:
Thanks @jmealo for your contribution(s)!

PR #950, opened 2026-05-22
perf(ld_test_decoding): cache table+attr catalog lookups in transform hot path

scottjacobsen:
## Summary

The test_decoding parser's UPDATE path (for tables with REPLICA IDENTITY DEFAULT, i.e. the common case where Postgres does not emit an explicit `old-key:` section) was performing per-row SQLite catalog queries:

- `catalog_lookup_s_table_by_name` — once per row
- `catalog_s_table_fetch_attrs` — once per row
- `catalog_lookup_s_attr_by_name` — **once per column per row**

On high-throughput workloads (~10k rows/sec on the source) this dominated transform-process CPU and caused the follow pipeline to fall behind WAL write rate on the primary, with replication slot lag growing unbounded.

## Evidence

\`perf top\` against a transform process pegged at 100% CPU showed ~30% of cycles in SQLite internals:

\`\`\`
13.08%  sqlite3VdbeExec
 4.70%  sqlite3RunParser
 3.63%  yy_reduce.isra.0
 3.02%  resolveExprStep
 2.39%  btreeParseCellPtr
 1.41%  sqlite3WhereBegin
 1.32%  whereLoopAddBtreeIndex
 1.12%  pcache1Fetch
 ...
\`\`\`

The repeated SQL statement parsing (\`sqlite3RunParser\` + \`yy_reduce\` + \`sqlite3GetToken\` + \`keywordCode\`) is the smoking gun: the same prepared statements were being re-parsed and re-planned per row instead of reused.

## Change

Add a per-process \`TestDecodingTableCache\` (uthash keyed by \`(nspname, relname)\`) that holds the table OID plus a nested uthash of per-column \`(attnum, attisprimary, attisreplident)\` keyed by \`attname\`. On first hit for a table, do the catalog work once and populate the cache; subsequent UPDATEs against the same table do a single \`HASH_FIND\` plus \`HASH_FIND_STR\` per column with no SQLite work.

The cache lives for the lifetime of the transform process and is never invalidated — pgcopydb does not support schema changes during follow, which matches the behaviour of the existing \`generatedColumnsCache\` immediately adjacent in \`StreamContext\`.

Replica-identity-index semantics are preserved: the cache stores both \`attisprimary\` and \`attisreplident\`, and the \`pkeyArray[c] = attribute->attisprimary || attribute->attisreplident\` check is unchanged.

## Test plan

- [x] Builds cleanly (\`make\` in \`src/bin/pgcopydb\`)
- [x] \`make -C tests cdc-test-decoding\` — directly exercises the patched path
- [x] \`make -C tests cdc-replica-identity-index\` — exercises the \`attisreplident\` branch
- [x] \`make -C tests cdc-endpos-between-transaction\`
- [x] \`make -C tests cdc-low-level\`
- [x] \`make -C tests cdc-wal2json\` (regression check)
- [x] \`make -C tests follow-wal2json\` (regression check)
- [x] Re-profile transform process on the original workload to confirm SQLite frames drop

Draft because I'd like to gather profile-after numbers before marking ready.

### DIFF
--- a/src/bin/pgcopydb/ld_stream.h
+++ b/src/bin/pgcopydb/ld_stream.h
@@ -332,6 +332,51 @@ typedef struct GeneratedColumnsCache
 
 
 /*
+ * TestDecodingAttrCache caches per-column attributes needed by the
+ * test_decoding parser hot path (replica-identity classification of each
+ * column on UPDATE messages).
+ */
+typedef struct TestDecodingAttrCache
+{
+	char attname[PG_NAMEDATALEN];
+
+	int attnum;
+	bool attisprimary;
+	bool attisreplident;
+
+	UT_hash_handle hh;           /* hashable by attname */
+} TestDecodingAttrCache;
+
+
+/*
+ * TestDecodingTableCache caches the SourceTable + per-attribute lookups for
+ * each table seen by the test_decoding parser. Without it, every UPDATE
+ * message (for tables that lack an old-key section, i.e. REPLICA IDENTITY
+ * DEFAULT) costs one SQL query per row to look up the table, one to fetch
+ * its attributes, plus one SQL query per column to fetch each attribute --
+ * which dominates transform CPU on high-throughput workloads.
+ *
+ * The (nspname, relname) key is stored as-is without the normalization
+ * applied by GeneratedColumnsCache: test_decoding emits identifiers
+ * already escaped by Postgres' deterministic quote_identifier (see the
+ * comment in parseTestDecodingMessageHeader), so the same real table
+ * always produces the same byte sequence -- no normalization needed.
+ */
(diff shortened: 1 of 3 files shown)


dimitri (approved):

scottjacobsen:
Some good new here - pgcopydb has been able to ingest the WAL log on a very high throughput database with essentially zero lag for 20+ hours so far. :tada:

dimitri:
> Some good new here - pgcopydb has been able to ingest the WAL log on a very high throughput database with essentially zero lag for 20+ hours so far. 🎉

Would you then say the PR is ready for review? it's been marked “draft” at the moment...

scottjacobsen:
Yes, it's ready to to review. It's burned through about 5 TB of log files without error so far.

PR #501, opened 2023-10-17
pgcopydb list matviews

dimitri:
This is the first step towards support for REFRESHing materialized views within pgcopydb instead of in pg_restore.

PR #1009, opened 2026-06-25
drive REFRESH MATERIALIZED VIEW within pgcopydb  (#484, #501)

### LINKED ISSUE #484: Refresh materialized views fails during post data pg_restore phase
REFRESH MATERIALIZED VIEWS fail when there is a reference to a object which is not fully qualified and the query is enclosed in a string.

**Query to reproduce** 
```
CREATE SCHEMA pgcopydb;

CREATE TABLE pgcopydb.test ("Name" text,  "LocalTag" text);

CREATE MATERIALIZED VIEW public.test_view AS
SELECT row_number() OVER (PARTITION BY true::boolean) AS row_number,
    ts_stat.word
   FROM ts_stat('SELECT to_tsvector(''simple'', "Name") ||
                 to_tsvector(''simple'', coalesce("LocalTag",'' ''))
            FROM test'::text) ts_stat(word, ndoc, nentry)
  WITH NO DATA;

SET search_path = 'pgcopydb';

REFRESH MATERIALIZED VIEW public.test_view;
```

**Logs**
```
19:26:42 43355 INFO    /usr/bin/pg_restore --dbname 'postgres://azureuser@100db-testcase10-fspg.postgres.database.azure.com/postgres?keepalives=1&keepalives_idle=10&keepalives_interval=10&keepalives_count=60' --single-transaction --clean --if-exists --use-list /tmp/pgcopydb/schema/post-filtered.list /tmp/pgcopydb/schema/post.dump
19:26:48 43355 ERROR  pg_restore: while PROCESSING TOC:
19:26:48 43355 ERROR  pg_restore: from TOC entry 4279; 0 139077 MATERIALIZED VIEW DATA test_view azureuser
19:26:48 43355 ERROR  pg_restore: error: could not execute query: ERROR:  relation "test" does not exist
19:26:48 43355 ERROR  LINE 3:             FROM test
19:26:48 43355 ERROR                           ^
19:26:48 43355 ERROR  QUERY:  SELECT to_tsvector('simple', "Name") ||
19:26:48 43355 ERROR                   to_tsvector('simple', coalesce("LocalTag",' '))
19:26:48 43355 ERROR              FROM test
19:26:48 43355 ERROR  Command was: REFRESH MATERIALIZED VIEW public.test_view;
19:26:48 43355 ERROR  Failed to run pg_restore: exit code 1
19:26:48 43355 ERROR  Failed to clone source database, see above for details
19:26:48 43353 ERROR  clone process 43355 has terminated [6]
```

dimitri:
## Problem

`pgcopydb clone` fails to populate materialized views whose definition
embeds an unqualified table name inside a text argument. The canonical
example is `ts_stat()`:

```sql
CREATE MATERIALIZED VIEW public.mv_word_stats AS
SELECT word, ndoc, nentry
FROM ts_stat(
    'SELECT to_tsvector(''simple'', title || '' '' || body)
     FROM documents'          -- ← unqualified
)
WITH NO DATA;
```

PostgreSQL cannot schema-qualify a name that appears inside opaque text
at `CREATE MATERIALIZED VIEW` time. At `REFRESH` time the embedded query
is evaluated with whatever `search_path` is in effect. `pg_restore`
always issues `SET search_path = ''` before running `REFRESH MATERIALIZED
VIEW` (standard pg_dump/pg_restore convention). With an empty
`search_path`, `documents` is not found and the clone aborts:

```
ERROR:  relation "documents" does not exist
```

The same failure affects any matview whose definition relies on
`search_path` to resolve names, not only `ts_stat()`.

## Root Cause

pg_restore uses a single libpq session for its entire post-data phase.
It opens that session, strips the `search_path` (`SET search_path = ''`),
and then replays every post-data TOC entry in that session — including
`REFRESH MATERIALIZED VIEW`. The `ALTER DATABASE SET search_path` that
the source database had configured is restored during pre-data, but it
only takes effect for **new** connections; pg_restore's already-open
session never sees it.

The fix cannot be applied inside pg_restore. pgcopydb must run the
`REFRESH` statements itself, in its own fresh connections.

## Fix

pgcopydb now owns REFRESH entirely. `REFRESH MATERIALIZED VIEW` entries
are removed from the pg_restore post-data list; pgcopydb drives them
directly through the vacuum worker pool that is already running during
the data phase.

### TOC dependency order

pg_dump sorts all dumpable objects — including `DO_REFRESH_MATVIEW`
entries — into a topological order that satisfies their dependency
constraints. This happens in two steps in the pg_dump source:

1. `buildMatViewRefreshDependencies` (`pg_dump.c`, called at line 1096)
   queries `pg_depend` (through `pg_rewrite`) for all matview→matview
   dependencies and registers them with `addObjectDependency`, wiring
   `DO_REFRESH_MATVIEW` objects together in the dump graph.

2. `sortDumpableObjects` (`pg_dump_sort.c:559`) runs a topological sort
   (`TopoSort`) over all objects, including those `DO_REFRESH_MATVIEW`
   nodes, and writes the result into the archive TOC in safe order.

The archive TOC is therefore a topologically sorted sequence of REFRESH
entries: if matview B depends on matview A, A's REFRESH entry appears
before B's in the TOC. pgcopydb reads this order out of the TOC and
uses it directly — no independent dependency resolution is needed.

### Catalog additions

Three new pieces of catalog state track REFRESH progress:

| What | Where | Purpose |
|---|---|---|
| `toc_seq INTEGER` | column on `s_matview` | position of this REFRESH in the pg_dump TOC |
| `s_matview_dep(matview_oid, dep_oid)` | new table | matview→matview edges from `pg_depend` |
| `s_matview_refresh_done(oid)` | new table | which matviews have been successfully refreshed |

### Preparation (catalog phase, before COPY)

During `copydb_target_prepare_schema`:

1. **`schema_list_matview_deps`** (`schema.c`) queries `pg_depend` on the
   source for all matview→matview dependency edges (`deptype = 'n'`,
   both ends `relkind = 'm'`) and stores them in `s_matview_dep`.

2. **`copydb_collect_matview_toc_order`** (`dump_restore.c`) re-reads
   the pg_dump archive TOC (the same file written for `--section=pre-data`)
   and assigns a monotone `toc_seq` to each `REFRESH MATERIALIZED VIEW`
   entry, storing it in `s_matview`. Because the TOC is already in
   topological order (see above), `toc_seq` encodes a safe refresh
   sequence.

All `REFRESH MATERIALIZED VIEW` entries are also removed from the
pg_restore lists for both pre-data and post-data, so pg_restore never
sees them.

### Dispatch (after CREATE INDEX)

A new SysV queue message type, `QMSG_TYPE_MATVIEW_OID`, is added to the
vacuum queue. After all `CREATE INDEX` work completes, the INDEX
supervisor calls **`vacuum_send_matviews`** (`vacuum.c`), which sends
one `QMSG_TYPE_MATVIEW_OID` per matview in ascending `toc_seq` order.

Vacuum workers are always started — even when `--skip-vacuum` is set —
because REFRESH messages must be serviced regardless of VACUUM ANALYZE.

### Refresh (vacuum workers)

Each vacuum worker handles `QMSG_TYPE_MATVIEW_OID` as follows:

1. **Dependency check** — query `s_matview_dep` to find all matviews
   this one depends on, then check `s_matview_refresh_done` to see if
   they are all done. Sleep 10 ms and retry until they are.

2. **Fresh connection** — open a new libpq connection to the target
   database. This connection is **not** an existing pg_restore session;
   it inherits the database-level `search_path` set by `ALTER DATABASE
   SET search_path`, which pg_restore already restored during pre-data.
   No explicit `SET search_path` is needed.

3. **REFRESH** — execute `REFRESH MATERIALIZED VIEW "schema"."name"`.

4. **Mark done** — `INSERT OR IGNORE INTO s_matview_refresh_done(oid)`.

Multiple workers run in parallel; the dep-check + sleep loop ensures
each matview waits only for its direct prerequisites and no longer.

### Tests

A new fixture (`tests/unit/setup/22-matview-refresh.sql`) builds a
five-matview dependency tree that exercises both the `ts_stat` bug and
dependency ordering:

```
public.mv_word_stats      (ts_stat with unqualified table name)
mvtest.mv_author_summary  (independent)
mvtest.mv_tag_stats       (independent)
        ↓                      ↓
   mvtest.mv_combined  (depends on both roots above)
        ↓
   mvtest.mv_final     (depends on mv_combined)
```

The source database has `ALTER DATABASE SET search_path TO mvtest, public`
so that new connections on the target (opened by vacuum workers) resolve
the unqualified `documents` name correctly.

A standalone integration test (`tests/matview-refresh/`) runs
`pgcopydb clone` end-to-end and verifies that all five matviews contain
the expected row counts on the target.

Closes #484
Closes #501

### DIFF
--- a/src/bin/pgcopydb/catalog.c
+++ b/src/bin/pgcopydb/catalog.c
@@ -113,7 +113,18 @@ static char *sourceDBcreateTableDDLs[] = {
 	"create table s_matview("
 	"  oid integer primary key, "
 	"  qname text, nspname text, relname text, restore_list_name text, "
-	"  exclude_data boolean"
+	"  exclude_data boolean, "
+	"  toc_seq integer default 0"
+	")",
+
+	"create table s_matview_dep("
+	"  matview_oid integer references s_matview(oid), "
+	"  dep_oid integer references s_matview(oid), "
+	"  primary key(matview_oid, dep_oid)"
+	")",
+
+	"create table s_matview_refresh_done("
+	"  oid integer primary key references s_matview(oid)"
 	")",
 
 	"create table s_table_size("
@@ -661,6 +672,8 @@ static char *sourceDBdropDDLs[] = {
 
 	"drop table if exists s_database",
 	"drop table if exists s_database_property",
+	"drop table if exists s_matview_refresh_done",
+	"drop table if exists s_matview_dep",
 	"drop table if exists s_table",
 	"drop table if exists s_matview",
 	"drop table if exists s_attr",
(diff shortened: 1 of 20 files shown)


dimitri:
## Why this is closed as wontfix

### The bug is PG ≤ 16 only

PostgreSQL 17 (commit `2af07e2f74`, Jeff Davis) added `RestrictSearchPath()` inside `RefreshMatViewByOid()`, which forces `search_path = 'pg_catalog, pg_temp'` for **every** `REFRESH MATERIALIZED VIEW` and every `CREATE MATERIALIZED VIEW ... WITH DATA`.  A matview that references an unqualified table name inside a `ts_stat()` argument string — the exact scenario behind #484/#501 — cannot even be created on PG17+, let alone refreshed.  The problem simply does not exist on modern PostgreSQL.

### pg_restore --jobs already parallizes REFRESH correctly

`identify_locking_dependencies()` in `pg_backup_archiver.c` skips exclusive-lock dependencies for REFRESH entries: they carry a dependency on the matview schema entry, which has `desc = 'MATERIALIZED VIEW'`, not `'TABLE'` or `'TABLE DATA'`, so `nlockids` is always 0.  Independent matviews therefore dispatch freely across pg_restore worker connections; dependent ones are serialized by the TOC dependency graph built by `buildMatViewRefreshDependencies()` + `sortDumpableObjects()`.  The parallelism argument for pgcopydb owning REFRESH evaporates.

### What this PR did

The full implementation — new SQLite tables (`s_matview_dep`, `s_matview_refresh_done`, `toc_seq` column), a dep-check loop in vacuum workers, `catalog_iter_s_matview_toc_order`, `vacuum_refresh_matview_by_oid`, `vacuum_send_matviews`, and the filtering hook — was ~570 lines of new code.  The sole real benefit was fixing the search_path bug on PG ≤ 16 targets.  Given that PG16 reaches end-of-life in November 2025, shipping that mass of complexity for a bug that can only affect one already-aging release is not justified.

The final commit on this branch adds a `targetPgVersionNum` guard so the implementation is completely bypassed on PG17+ targets, confirming that the pg_restore path works correctly.  On PG ≤ 16 the implementation runs and is correct, but the question is whether those users are better served by qualifying their matview table references (the straightforward fix) than by adding 570 lines of infrastructure to pgcopydb.

PR #1026, opened 2026-06-30
Restore role ownership when copying LOBs for PG=>17

### LINKED ISSUE #1027: Large object ownership not preserved on PostgreSQL 17+
On PostgreSQL 17+, pgcopydb doesn't preserve large object ownership (`pg_largeobject_metadata.lomowner`) - every blob ends up owned by the connecting role instead of its original owner. PG16 and earlier are fine. Root cause: PG17's `pg_dump` moved large object creation out of `SECTION_PRE_DATA` into the data section, but pgcopydb only restores pre-data, so `pg_copy_large_object()` creates the blob itself via `lo_create()` (owned by the connecting role) and never restores `lomowner`. On PG16 the pre-data restore still carried `ALTER LARGE OBJECT ... OWNER TO`, which is why it worked there.

To reproduce: reassign a blob to a non-superuser role on the source (`ALTER LARGE OBJECT <oid> OWNER TO somerole`), migrate to a PG17+ target, and the target blob comes back owned by the connecting role.

kbarber:
Seems we lost ownership copying with the changes to 17 and up, this will make sure when ownership is expected, ownership is transferred with the LOB.

Fixes: https://github.com/dimitri/pgcopydb/issues/1027

### DIFF
--- a/src/bin/pgcopydb/blobs.c
+++ b/src/bin/pgcopydb/blobs.c
@@ -21,6 +21,7 @@ typedef struct BlobMetadataArray
 {
 	int count;
 	Oid oids[MAX_BLOB_PER_FETCH];
+	char rolnames[MAX_BLOB_PER_FETCH][PG_NAMEDATALEN];
 } BlobMetadataArray;
 
 typedef struct BlobMetadataArrayContext
@@ -424,12 +425,14 @@ copydb_blob_worker(CopyDataSpec *specs)
 
 				if (!pg_copy_large_object(src, &dst,
 										  dropIfExists,
-										  mesg.data.oid,
+										  !specs->restoreOptions.noOwner,
+										  mesg.data.lo.oid,
+										  mesg.data.lo.rolname,
 										  &bytesTransmitted))
 				{
 					log_error("Failed to copy Large Object with oid %u, "
 							  "see above for details",
-							  mesg.data.oid);
+							  mesg.data.lo.oid);
 					return false;
 				}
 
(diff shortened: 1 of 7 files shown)


dimitri:
Thanks for digging into this and providing a well-scoped fix with a clear test — the root cause analysis in the issue is spot on.

One small thing to fix before merge: the ALTER error path calls `pgcopy_log_error(dst, NULL, context)` after `pgsql_execute` already logged the PQ error internally, so the error ends up logged twice. `pgcopy_log_error` also calls `pgsql_finish(dst)` internally, making the subsequent explicit `pgsql_finish(dst)` redundant (safe because `pgsql_finish` guards on a NULL connection, but confusing). The pattern used elsewhere in `pg_copy_large_object` is simpler:

```c
if (!pgsql_execute(dst, sql))
{
    log_error("Failed to set owner of large object %u to %s", blobOid, rolname);
    pgsql_finish(src);
    pgsql_finish(dst);
    return false;
}
```

Everything else looks good:

- The `dstIsPG17orLater` guard is correct — it matches the existing architectural split in the function where blobs are either pre-created by the pre-data restore (PG16 path, owner carried by pg_dump pre-data) or created by `lo_create()` here (PG17+ path, owner must be set explicitly).
- Buffer sizing is fine: `PG_NAMEDATALEN` in this codebase is `NAMEDATALEN * 2 + 2 = 130` bytes, sized specifically for quoted identifiers.
- `format('%I', pg_catalog.pg_get_userbyid(lomowner))` on the source is the right way to get a safely-quoted role name for the `ALTER`.
- `--no-owner` is correctly respected via `!specs->restoreOptions.noOwner`.
- The test is solid: mixed ownership on the source (one blob reassigned to `blobowner`, the rest owned by the connecting role), then a `diff` on `(oid, owner, parts, size)` to prove per-blob ownership is preserved end-to-end.

kbarber:
> Thanks for digging into this and providing a well-scoped fix with a clear test — the root cause analysis in the issue is spot on.

Aah good to hear.

> One small thing to fix before merge: the ALTER error path calls `pgcopy_log_error(dst, NULL, context)` after `pgsql_execute` already logged the PQ error internally, so the error ends up logged twice. `pgcopy_log_error` also calls `pgsql_finish(dst)` internally, making the subsequent explicit `pgsql_finish(dst)` redundant (safe because `pgsql_finish` guards on a NULL connection, but confusing). The pattern used elsewhere in `pg_copy_large_object` is simpler:

Done.

> Everything else looks good:

Happy to help @dimitri hope you're doing well 👋 . Let me know if there is anything else.

